PR #59, opened 2018-08-27
Ternary update

DanRStevens:
Saw some odd looking ternary expressions that reminded me of stuff I used to do. Updated them to use the result of the ternary expression, rather than trying to use the ternary operator as some kind of statement with side effects.


DanRStevens:
Ahh, and you may want to merge in the Linux fixes and CI updates from the other branch first if you want the CI builds to pass.

Issue #11, opened 2018-04-27
Connectedness Search Fails


ldicker83:
There are some cases when the depth first connectedness search fails and marks tiles as connected when they shouldn't be:

![image](https://user-images.githubusercontent.com/13858902/39387321-f5cfc7ac-4a46-11e8-9996-ed0f40799b96.png)

Placement behavior is correct.

[Broken Depth Search.zip](https://github.com/ldicker83/OPHD/files/1956831/Broken.Depth.Search.zip)

PR #91, opened 2019-02-22
Fix for issue #11

Sirbomber:
Fix for issue #11


ldicker83:
Okay, so it's looking weird online, for whatever reason it's being treated as an entirely new file. Using a manual DIFF check, I can see the changes which generally look good. There's some stylistic choices that aren't consistent with the current code base but it's older code so I wouldn't expect it to match.

Going to merge these changes and touch it up to match the current code styling I've been using in the rest of the code base.

PR #106, opened 2019-04-28
Remove unused variables

DanRStevens:
Ok, this PR might go slightly too far. It removes unused variables. Some of them I assume were planning to be used, but their use hasn't been implemented yet. Others have maybe been forgotten about and will never be used.

### DIFF
--- a/src/Map/TileMap.cpp
+++ b/src/Map/TileMap.cpp
@@ -328,8 +328,6 @@ void TileMap::draw()
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


ldicker83:
Nope, these changes are fine. I was just focusing on higher priority stuff.

Some of the unused variables in TextField I hadn't gotten around to using after updating to the latest versions of NAS2D that didn't yet have clippers available. At some point I'll reimplement text fields that allow for scrolling back and forth.

PR #121, opened 2019-12-24
Add enum scope

### LINKED ISSUE #120: Enum values used without enum scope
A number of enum values are used without the enum scope as a prefix. This will break if the `enum` is updated to be an `enum class`.

In NAS2D, a change has already been introduced to update `enum` definitions to `enum class` definitions. We should update references in OPHD now to make upgrade of NAS2D easier later. In particular, the `EventHandler.h` related `enum` references need some updating.

Example:
`EventHandler::BUTTON_LEFT` => `EventHandler::MouseButton::BUTTON_LEFT`
`EventHandler::KEY_ENTER` => `EventHandler::KeyCode::KEY_ENTER`

DanRStevens:
Closes #120

Adds consistent scope prefix to `enum` values:
- `EventHandler::MouseButton`
- `EventHandler::KeyCode`

This is necessary for when those `enum` defintions are changed to `enum class` (already merged to NAS2D).

### DIFF
--- a/src/States/MainReportsUiState.cpp
+++ b/src/States/MainReportsUiState.cpp
@@ -258,7 +258,7 @@ void MainReportsUiState::_deactivate()
 void MainReportsUiState::onKeyDown(EventHandler::KeyCode key, EventHandler::KeyModifier mod, bool repeat)
 {
 	if (!active()) { return; }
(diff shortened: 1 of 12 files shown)


ldicker83:
This has been addressed and resolved in a recent commit and PR that I did that updates OPHD to the latest builds of NAS2D.

PR #145, opened 2020-01-19
VS project file cleanup

DanRStevens:
This removes some old cruft that seems to have accumulated over time. It looks safe to remove, and I don't expect Visual Studio will automatically add it back.

### DIFF
--- a/proj/vs2019/ophd.vcxproj
+++ b/proj/vs2019/ophd.vcxproj
@@ -65,7 +65,6 @@
   <ImportGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
-  <PropertyGroup Label="UserMacros" />
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


DanRStevens:
We can probably do more to simplify the VS project file. The remaining changes look a bit higher risk though, so I would want to handle them in a separate PR. In particular, a few conditional groups have identical settings and could have the conditional section simplified. The Code Analysis settings look to be slightly inconsistent between configurations, which we might want to adjust a bit. There is a `_ProjectFileVersion` setting which I think is outdated. Might be helpful to experiment with them from an actual copy of Visual Studio.

ldicker83:
Looks okay for me but this seems like wasted effort overall? The solution file is generated by Visual Studio and may just be added back over time.

Anyway, in the mean time I'll merge these changes.

DanRStevens:
Valid points. Not really high priority changes, and I don't want to make a habit of editing auto generated output. Visual Studio likely did add that cruft in the first place, and could add it back again. In particular, if those specific settings are changed, it may add stuff back.

I suppose this comes from having to manually edit project files on Linux whenever new source code files are added, hence I often have to look at those files. I've also had to occasionally modify configuration settings this way. Having less in the project files makes it easier to find the place I want to edit. Having them consistent also makes it easier to compare settings across configurations to make sure nothing was missed. In particular, there was the warning suppression for just the `Debug` configurations that was removed.

I suppose each time I look at the files I've been tempted to experiment with more advanced project file features. That would be easier to do from a cleaner project file, though I suspect I'll resist the temptation to make such changes.

PR #150, opened 2020-01-27
Place multiple tubes sections while holding Shift key

ldicker83:


src/States/MapViewState.h
```diff
@@ -115,6 +115,10 @@ class MapViewState : public Wrapper
 	void placeRobot();
 	void placeStructure();
 	void placeTubes();
+	
+	Rectangle_2d tubeStart;
```
Goof-fr: Is there a better object to store a point (X/Y) and a Boolean.
I use x and y in a natural way and the height as a Boolean to avoid some unwanted behaviors

DanRStevens: There are now `Point` and `Vector` template classes to store `x` and `y` pairs.

One small caveat, `Vector` is a struct with public `x` and `y` fields, while `Point` has private `x` and `y` with accessor methods `x()` and `y()`. We're planning to update the `Point` class to just have plain public member fields.

PR #152, opened 2020-01-29
Adds a UI Label control

cugone:
Useful for drawing text in UI Windows to label other UI elements.

### DIFF
--- a/src/States/MainMenuState.cpp
+++ b/src/States/MainMenuState.cpp
@@ -74,7 +74,7 @@ void MainMenuState::initialize()
 	mFileIoDialog.anchored(false);
 	mFileIoDialog.hide();
 
-    Font* tiny_font = Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
+	Font* tiny_font = Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
 	auto& r = NAS2D::Utility<NAS2D::Renderer>::get();
 	lblVersion.font(tiny_font);
 	lblVersion.text(constants::VERSION);
--- a/src/States/MainMenuState.cpp
+++ b/src/States/MainMenuState.cpp
@@ -74,6 +74,13 @@ void MainMenuState::initialize()
 	mFileIoDialog.anchored(false);
 	mFileIoDialog.hide();
 
+    Font* tiny_font = Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
+	auto& r = NAS2D::Utility<NAS2D::Renderer>::get();
+	lblVersion.font(tiny_font);
+	lblVersion.text(constants::VERSION);
+	lblVersion.position(r.width() - tiny_font->width(constants::VERSION) - 5, r.height() - tiny_font->height() - 5);
+	lblVersion.color(NAS2D::Color::White);
+
 	positionButtons();
 	disableButtons();
 
@@ -291,9 +298,7 @@ NAS2D::State* MainMenuState::update()
 
 	mFileIoDialog.update();
 
-	Font* tiny_font = Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
-
-	r.drawText(*tiny_font, constants::VERSION, r.width() - tiny_font->width(constants::VERSION) - 5, r.height() - tiny_font->height() - 5, 255, 255, 255);
+	lblVersion.update();
 
 	if (r.isFading())
 	{
--- a/src/UI/Core/Label.cpp
+++ b/src/UI/Core/Label.cpp
@@ -6,10 +6,7 @@
 #include "NAS2D/Renderer/Renderer.h"
 #include "NAS2D/Resources/Font.h"
 
-#include "NAS2D/MathUtils.h"
 #include "NAS2D/Utility.h"
-
-#include <locale>
 
 static const int FIELD_PADDING = 4;
 static NAS2D::Font* TXT_FONT = nullptr;
--- a/src/UI/Core/Label.cpp
+++ b/src/UI/Core/Label.cpp
@@ -19,10 +19,6 @@ Label::Label() {
     height(static_cast<float>(TXT_FONT->height() + FIELD_PADDING * 2));
 }
 
-Label::~Label() {
-
-}
-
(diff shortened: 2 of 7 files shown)


src/UI/Core/Label.cpp
```diff
@@ -0,0 +1,41 @@
...
+static NAS2D::Font* TXT_FONT = nullptr;
+
+Label::Label() {
+    TXT_FONT = NAS2D::Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
+    height(static_cast<float>(TXT_FONT->height() + FIELD_PADDING * 2));
+}
+
+Label::~Label() {
+
+}
```
DanRStevens: Can this be set as `default` in the header for a `virtual` method? I suspect that would be possible.

src/UI/Core/Label.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "NAS2D/Renderer/Renderer.h"
+#include "NAS2D/Resources/Font.h"
+
+#include "NAS2D/MathUtils.h"
+#include "NAS2D/Utility.h"
+
+#include <locale>
+
+static const int FIELD_PADDING = 4;
+static NAS2D::Font* TXT_FONT = nullptr;
```
DanRStevens: This looks like it would be more suitable as a member variable rather than a `static` global.

src/UI/Core/Label.h
```diff
@@ -0,0 +1,41 @@
...
+#include "Control.h"
+
+#include "NAS2D/Renderer/Color.h"
+#include "NAS2D/Resources/Font.h"
+
+/**
+ * \class TextField
+ * \brief A Basic Text Field.
+ *
+ * The Field Control is just a set of rendered text a user can enter.
```
DanRStevens: This comment appears to have been copy/pasted and was not updated.

I was wondering why the style was so consistent with existing code until I saw this. :wink:

src/UI/Core/Label.h
```diff
@@ -0,0 +1,41 @@
...
+// = Copyright � 2008 - 2012, New Age Software
+// ==================================================================================
+// = This file is part of the NAS2D+GUI library.
+// ==================================================================================
+
+#pragma once
+
+#include "NAS2D/EventHandler.h"
+#include "NAS2D/Timer.h"
+#include "NAS2D/Resources/Image.h"
```
DanRStevens: These 3 headers don't seem to be used here.

src/UI/Core/Label.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "../../Constants.h"
+#include "../../FontManager.h"
+
+#include "NAS2D/Renderer/Renderer.h"
+#include "NAS2D/Resources/Font.h"
+
+#include "NAS2D/MathUtils.h"
+#include "NAS2D/Utility.h"
+
+#include <locale>
```
DanRStevens: I don't see any obvious use of `<locale>`.

src/States/MainMenuState.cpp
```diff
@@ -74,6 +74,13 @@ void MainMenuState::initialize()
 	mFileIoDialog.anchored(false);
 	mFileIoDialog.hide();
 
+    Font* tiny_font = Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
```
DanRStevens: First off, I realize this name was pre-existing.

Would prefer a more descriptive name of how the font is being used, rather than a presumed property of the style of the font (`tiny_font`). Actual characteristics of the font may change due to style updates, though how it is being used, or rather what it represents is not based on style. The name makes me think of text used in a footer, but is hard to know looking at just the code.

src/States/MainMenuState.cpp
```diff
@@ -74,6 +74,13 @@ void MainMenuState::initialize()
...
+	auto& r = NAS2D::Utility<NAS2D::Renderer>::get();
+	lblVersion.font(tiny_font);
+	lblVersion.text(constants::VERSION);
+	lblVersion.position(r.width() - tiny_font->width(constants::VERSION) - 5, r.height() - tiny_font->height() - 5);
+	lblVersion.color(NAS2D::Color::White);
```
DanRStevens: Would be nice if a constructor could take some or all of these properties.

src/UI/Core/Label.cpp
ldicker83: It should be as long as it's declared virtual (though I suspect that only really matters in the parent class)

src/UI/Core/Label.cpp
ldicker83: This is matching the code from many of the other derived controls. I did it mostly as a way to avoid getting pointers to a Font object every time it's needed. In the other controls they're all static local variables and are there literally to be able to grab a pointer once and then use the pointer where needed. The pointer can be updated as needed.

src/UI/Core/Label.cpp
ldicker83: This is probably from the TextField and/or TextArea controls and can be safely removed unless there are plans to make use of it in the future.

src/States/MainMenuState.cpp
ldicker83: This is left over from the very beginning of the development process. I'm not opposed to changing it though it's probably better noted and left for a different change set.

src/States/MainMenuState.cpp
ldicker83: I opted for cleaner c'tors. Again, not opposed though to be fair the C# notion of properties and setting them during object instantiation is awesome:

```csharp
var SomeObject = new SomeObjectType
{
    Property1 = Whatever;
    Property2 = Whatever;
    // ... etc.
};
```

If modern C++ has a similar notion of initializing an object in this way (I recall seeing similar syntax somewhere in the NAS2D code) I'd be completely okay doing it that way ... would make this a little cleaner at least versus calling a bunch of member methods after object instantiation.

To be fair though, it's a different change set and could be figured out later on.

src/UI/Core/Label.cpp
DanRStevens: Oh yeah, I suppose the destructor could just be removed if it's not actually being overridden. Ehh, minor point.

src/UI/Core/Label.cpp
DanRStevens: It also has the side effect of changing the font for all controls that share the pointer at the same. It's not possible to have two different labels with two different fonts shown at the same time.

I agree with your assessment about the change being out of scope though.

PR #155, opened 2020-01-30
Fix warnings pass

cugone:
This fixes over 400 warnings in both Debug and Release builds.

It deliberately skips warnings emitted from third-party code such as Micropather.

Keeping on top of warnings as they occur will reduce massive chagesets like this in teh future.

### DIFF
--- a/src/GraphWalker.cpp
+++ b/src/GraphWalker.cpp
@@ -39,7 +39,7 @@ static bool validConnection(Structure* src, Structure* dst, Direction _d)
 	if (src == nullptr || dst == nullptr)
 	{
 		#ifdef DEBUG
-		throw std::runtime_error("GraphWalker::validConnection() was passed a NULL Pointer.");
+		std::cout << "GraphWalker::validConnection() was passed a NULL Pointer." << std::endl;
 		#endif
 		return false;
 	}
--- a/src/Map/TileMap.h
+++ b/src/Map/TileMap.h
@@ -79,7 +79,7 @@ public:
 	/** MicroPather public interface implementation. */
 	virtual float LeastCostEstimate(void* stateStart, void* stateEnd);
 	virtual void AdjacentCost(void* state, std::vector<micropather::StateCost>* adjacent);
-	virtual void PrintStateInfo(void* state) {};
+	virtual void PrintStateInfo(void* /*state*/) { /* DO NOTHING */ };
 
 protected:
 	/**
--- a/src/Map/TileMap.h
+++ b/src/Map/TileMap.h
@@ -79,7 +79,7 @@ public:
 	/** MicroPather public interface implementation. */
 	virtual float LeastCostEstimate(void* stateStart, void* stateEnd);
 	virtual void AdjacentCost(void* state, std::vector<micropather::StateCost>* adjacent);
-	virtual void PrintStateInfo(void* /*state*/) { /* DO NOTHING */ };
+	virtual void PrintStateInfo(void* /*state*/) {};
 
 protected:
 	/**
--- a/src/RobotPool.cpp
+++ b/src/RobotPool.cpp
@@ -147,17 +147,17 @@ bool RobotPool::robotAvailable(RobotType _type)
 		return getDigger() != nullptr;
 	}
 	case ROBOT_DOZER:
-    {
-        return getDozer() != nullptr;
-    }
+	{
+		return getDozer() != nullptr;
+	}
 	case ROBOT_MINER:
-    {
-        return getMiner() != nullptr;
-    }
+	{
+		return getMiner() != nullptr;
+	}
 	default:
-    {
-        return false;
-    }
+	{
+		return false;
+	}
 	}
 }
 
--- a/src/RobotPool.cpp
+++ b/src/RobotPool.cpp
@@ -143,23 +143,22 @@ bool RobotPool::robotAvailable(RobotType _type)
 	switch (_type)
 	{
 	case ROBOT_DIGGER:
+	{
 		return getDigger() != nullptr;
-		break;
-
-	case ROBOT_DOZER:
-		return getDozer() != nullptr;
-		break;
-
-	case ROBOT_MINER:
-		return getMiner() != nullptr;
-		break;
-
-	default:
-		return false;
-		break;
 	}
-
-	return false;
+	case ROBOT_DOZER:
+    {
+        return getDozer() != nullptr;
+    }
+	case ROBOT_MINER:
+    {
+        return getMiner() != nullptr;
+    }
+	default:
+    {
+        return false;
+    }
+	}
 }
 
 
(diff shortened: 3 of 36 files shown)


src/GraphWalker.cpp
```diff
@@ -39,7 +39,7 @@ static bool validConnection(Structure* src, Structure* dst, Direction _d)
 	if (src == nullptr || dst == nullptr)
 	{
 		#ifdef DEBUG
-		throw std::runtime_error("GraphWalker::validConnection() was passed a NULL Pointer.");
+		std::cout << "GraphWalker::validConnection() was passed a NULL Pointer." << std::endl;
```
DanRStevens: I don't see why this change is included in this change set.

I would be tempted to argue that a `nullptr` value there is a programming error that should crash hard. Though perhaps a better design would use references instead, which are not nullable.

src/RobotPool.cpp
```diff
@@ -143,23 +143,22 @@ bool RobotPool::robotAvailable(RobotType _type)
 	switch (_type)
 	{
 	case ROBOT_DIGGER:
+	{
 		return getDigger() != nullptr;
-		break;
-
+	}
```
DanRStevens: I believe the `break` is still needed here.

src/Map/TileMap.h
```diff
@@ -79,7 +79,7 @@ class TileMap: public micropather::Graph
 	/** MicroPather public interface implementation. */
 	virtual float LeastCostEstimate(void* stateStart, void* stateEnd);
 	virtual void AdjacentCost(void* state, std::vector<micropather::StateCost>* adjacent);
-	virtual void PrintStateInfo(void* state) {};
+	virtual void PrintStateInfo(void* /*state*/) { /* DO NOTHING */ };
```
DanRStevens: I would prefer to not include `/* DO NOTHING */` comments. It doesn't really tell us anything that the empty `{}` doesn't. The best comment is no comment.

Same for all the other similar comments.

src/UI/Core/ComboBox.cpp
```diff
@@ -115,7 +115,8 @@ void ComboBox::onMouseDown(EventHandler::MouseButton button, int x, int y)
 }
 
 
-void ComboBox::onMouseWheel(int x, int y)
+//TODO: Implement in terms of internal ListBox.
```
DanRStevens: This is better as an issue on GitHub than a source code comment.

src/UI/Core/ListBox.cpp
```diff
@@ -198,8 +198,8 @@ void ListBox::dropAllItems()
 	_updateItemDisplay();
 }
 
-
-void ListBox::onMouseDown(EventHandler::MouseButton button, int x, int y)
+//TODO: Intended to only react if LMB?
```
DanRStevens: That's a very good point. Again, this should be an issue on GitHub, rather than a source code comment.

Related, there may be hotkeys, such as `<Enter>` which might function similarly.

src/main.cpp
```diff
@@ -51,6 +51,7 @@ void validateVideoResolution()
 
 int main(int argc, char *argv[])
 {
+	(void)argc; //UNUSED
```
DanRStevens: I'm not fond of this idiom. I would prefer the name simply be commented out in the parameter list.

src/RobotPool.cpp
cugone: The return statement makes the `break;` unreachable.

DanRStevens: Oh geeze, right.

src/RobotPool.cpp
```diff
@@ -143,23 +143,22 @@ bool RobotPool::robotAvailable(RobotType _type)
...
 	case ROBOT_DOZER:
-		return getDozer() != nullptr;
-		break;
-
+    {
+        return getDozer() != nullptr;
+    }
 	case ROBOT_MINER:
```
DanRStevens: Hmm, indentation looks a bit off in this section.

cugone: (╯°□°)╯︵ ┻━┻ clang-format!

cugone: "Fixed"

DanRStevens: Check your diffs when you commit :wink:

And maybe disable clang-format.

cugone: I do, I can't see the difference between tabs and spaces in the diffs. Since I have tabs stops set to four, spaces and tabs look the same on my machine vs github and the diff coloring is just a solid bar of color, no whitespace markers.

See: https://stackoverflow.com/a/52764248/421178 for an example of what GitKraken shows me.

src/GraphWalker.cpp
cugone: Debug builds complain of unreachable code, otherwise. To clarify: if you're checking for `nullptr`, by definition a `nullptr` value isn't exceptional and shouldn't throw.

src/Map/TileMap.h
cugone: Sorry about that. It's a (bad?) habit to let people and future me know that the implementation isn't supposed to do anything on purpose. Fixed now.

DanRStevens:
Too tired for further review tonight. Would still appreciate if this was broken up a bit. In particular my eyes glaze over on the `static_cast` commit. If you do break things up a bit, you could potentially squash some of the fixup commits into the original changes.

cugone:
I was under the impression that grouping related changes to multiple files was better than a bunch of tiny whole PRs so they'd all get pushed at once. In this case, the grouping is "I'm fixing all the warnings."

DanRStevens:
Merging another branch has caused a merge conflict which will need to be resolved.

----

Given the vast number of possible warnings, I don't think "fixing all the warnings" is necessarily a great grouping for changes. It can be fine, when the changes are small in number, and simple for the fix.

In this case, there are quite a few changes. Being able to break them up can make it easier to review them. A smaller change set also means a smaller chance for a merge commit for any given section of changes. It also means a smaller chance that fixes in one area will hold up other areas.

There is nothing wrong with having lots of small PRs. I would actually prefer that. It's not wrong to create a PR for a 1 character change for a typo in a comment. In fact, those are really nice and easy PRs to review, so I rather enjoy them.


In this case, there are a few different types of warnings, so it may make sense to break things up by type. There are also different solutions for a given warning, based on the characteristics of the code, which would be another good way to break things up. In particular, the "narrowing conversions" have different solutions. Some add `static_cast`, while others change the data type. The different types of solutions can be split from each other. In this case, there are many, so I think they should be split up. Particularly since the diff interleaves the types of changes, which makes review slower and harder (some `static_cast` changes, some type changes, some `static_cast` changes, some type suffixes, some type changes, some more `static_cast` changes, etc.).

I want to hold off reviewing until this one can be split up a bit, particularly for the narrowing conversion changes.

cugone:
Closing due to split into separate PRs listed above.

PR #153, opened 2020-01-29
Implements Main Menu Options button

cugone:
Implements a basic options window on the main menu.

PR #157, opened 2020-01-30
Makes WINDOW_TITLE_HEIGHT visible to derived classes

cugone:
This is a branch off of #153 only the last two commits are new.

Previously WINDOW_TITLE_HEIGHT was internal to `Window.cpp` and unavailable to derived classes for GUI element positioning. This PR makes the variable a compile-time constant and visible to derived classes through a getter.

### DIFF
--- a/src/UI/Core/Window.cpp
+++ b/src/UI/Core/Window.cpp
@@ -57,7 +57,7 @@ void Window::onMouseDown(EventHandler::MouseButton button, int x, int y)
 
 	UIContainer::onMouseDown(button, x, y);
 
-	mMouseDrag = (button == EventHandler::MouseButton::BUTTON_LEFT && pointInRect_f(x, y, rect().x(), rect().y(), rect().width(), Window::getWindowTitleBarHeight()));
+	mMouseDrag = (button == EventHandler::MouseButton::BUTTON_LEFT && pointInRect_f(x, y, rect().x(), rect().y(), rect().width(), sWindowTitleBarHeight));
 }
 
 
@@ -97,7 +97,7 @@ void Window::update()
 	Renderer& r = Utility<Renderer>::get();
 
 	r.drawImage(mTitle[0], rect().x(), rect().y());
-	r.drawImageRepeated(mTitle[1], rect().x() + 4, rect().y(), rect().width() - 8, Window::getWindowTitleBarHeight());
+	r.drawImageRepeated(mTitle[1], rect().x() + 4, rect().y(), rect().width() - 8, sWindowTitleBarHeight);
 	r.drawImage(mTitle[2], rect().x() + rect().width() - 4, rect().y());
 
 	r.drawImageRect(rect().x(), rect().y() + 20, rect().width(), rect().height() - 20, mBody);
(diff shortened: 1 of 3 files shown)


src/UI/Core/Window.h
```diff
@@ -19,7 +19,7 @@ class Window : public UIContainer
 	void onMouseDown(NAS2D::EventHandler::MouseButton button, int x, int y);
 	void onMouseUp(NAS2D::EventHandler::MouseButton button, int x, int y);
 	void onMouseMotion(int x, int y, int dX, int dY);
-
+	static inline constexpr float getWindowTitleBarHeight() { return sWindowTitleBarHeight; };
```
DanRStevens: I would probably have just moved the constant to the header file as a `protected` `static const` member. That would have been easier than introducing a new `get` method.

As it stands now, the constant was moved to the header file anyway, it just got put in the `private` section.

DanRStevens:
Was also thinking, this could have been a branch off of `master`, since it doesn't depend on PR #153.

This could be merged into PR #153 to complete the implementation there, before PR #153 gets merged to `master`. That's a bit of a different thing though.

src/UI/Core/Window.h
cugone: I was unsure whether or not the derived classes were allowed to *change* the value since it was originally a local variable, so I went the safe route.

DanRStevens: I'm not sure I follow.

The original seems to be a `const` declared in the .cpp file. By moving it into the header as a protected member, it becomes accessible to subclasses. It's still just a `static const` though. Subclasses can choose if they want to use it or not, or define their own constant.

cugone: It's safer to code with only as much access as needed. Only bumping it up on a case-by-case basis as needed.

In the future if a class `WindowBorderless` gets created or a derived window needs the option to create a window without a title bar, then `sWindowTitleBarHeight` can no longer be `const` nor `static` and the variable can move.

Right now, it acts as if it were a compile-time literal value, with all the compiler optimization that entails.

DanRStevens:
The base branch has been merged. Please rebase on `master`. (This may be important as the base branch has had some history edited).

----

My guess is, the second last commit could have been an independent change, which didn't depend on #153. That one commit could have been it's own PR, and perhaps even merged before #153 was merged. Looks like the last commit could have become part of #153 in that case, if the second last commit had been merged there as well. Just some thoughts about how to make changes more independent.

cugone:
To clarify: Rebase on second-to-last or last commit on master?

src/UI/Core/Window.cpp
```diff
@@ -12,7 +12,6 @@
 
 using namespace NAS2D;
 
-const float WINDOW_TITLE_BAR_HEIGHT = 20.0f;
```
DanRStevens: I see this constant has been moved to the header file. I would recommend placing it in the `protected` section. That would be a simpler change, and would not require adding a new `protected` `static` method for `getWindowTitleBarHeight`. Either which way, a new `protected` member must be added, either as a constant, or as a static function. Given the definition of the new static function, it's unlikely it would ever return anything other than a constant anyway.

PR #158, opened 2020-01-30
Adds combobox item/text setters

### LINKED ISSUE #154: Combobox needs a way to set its current item and the text field.
While writing the code for the Main Menu Options dialog I needed a way to set the current selection so it appears in the text field of the box during construction but before a user selected anything. There is no way to either set the text directly or a way to set an item as selected so it sets the text when selected.

A simple override of `Control::text` will work for the text field:
```
virtual void Combobox::text(const std::string& text) override;
//...
void Combobox::text(const std::string& text)
{
    txtField.text(text);
}
```

Setting the current selection via an override of `currentSelection` that exposes the underlying Listbox should work.
```
void currentSelection(int index)
{
    lstItems.currentSelection(index);
    text(lstItems.selectionText());
    selectionChanged();
}
```

Ref: https://www.tutorialspoint.com/vb.net/vb.net_combobox.htm

cugone:
Closes #154

Branches off #153. Only the last commit is new.

### DIFF
--- a/src/UI/Core/ComboBox.cpp
+++ b/src/UI/Core/ComboBox.cpp
@@ -186,6 +186,12 @@ int ComboBox::selectionTag() const
 }
 
 
+void ComboBox::currentSelection(unsigned int index) {
+	lstItems.currentSelection(index);
+	text(lstItems.selectionText());
+	mSelectionChanged();
+}
+
 /**
  * 
  */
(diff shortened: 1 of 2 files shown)


DanRStevens:
The last commit rebases cleanly on `master`. It does not need to be based off of #153.

I recommend rebasing the last commit on `master` (force push this branch). That way just the last commit is reviewed and merged.

cugone:
Rebased on last commit.

src/UI/Core/ComboBox.cpp
```diff
@@ -186,6 +186,12 @@ int ComboBox::selectionTag() const
 }
 
 
+void ComboBox::currentSelection(unsigned int index) {
+    lstItems.currentSelection(index);
+	text(lstItems.selectionText());
+	mSelectionChanged();
```
DanRStevens: Indentation here looks off.

cugone:
Fixed

PR #159, opened 2020-01-30
Fixes various bugs in MainMenuOptions

cugone:
Fixes various bugs in the options menu:
- Disables start in maximized windowed mode.
- Renables the options button due to master merge disabling it.
- Resolution combobox does not have a default value
- Fullscreen checkbox is tied to vsync checkbox
- isInit is removed
- Uses `std::to_string` instead of `std::ostringstream`
- Resolution popup appears when it shouldn't.
- Removes resolution popup
- Resolution combobox is changed when re-entering the menu even if Cancel is clicked.
- Apply button does not properly disable itself when no changes have been made.
- Occasionally fullscreen is applied even though the setting is unchecked.

### DIFF
--- a/src/UI/MainMenuOptions.cpp
+++ b/src/UI/MainMenuOptions.cpp
@@ -101,7 +101,7 @@ void MainMenuOptions::init()
 			desired_mode.h = resolution.height;
 			if(SDL_GetClosestDisplayMode(display_index, &desired_mode, &closest_mode))
 			{
-                NAS2D::DisplayDesc closest{closest_mode.w, closest_mode.h, closest_mode.refresh_rate};
+				NAS2D::DisplayDesc closest{closest_mode.w, closest_mode.h, closest_mode.refresh_rate};
 				if(current_dimensions == closest)
 				{
 					//Set combobox to current dimensions of window, not desktop
@@ -110,7 +110,7 @@ void MainMenuOptions::init()
 			}
 			std::string resolution_str = std::to_string(resolution.width) + "x" + std::to_string(resolution.height) + "x" + std::to_string(resolution.refreshHz);
 			cmbResolution.addItem(resolution_str, resolution_index);
-            ++resolution_index;
+			++resolution_index;
 		}
 		cmbResolution.currentSelection(currentResolutionSelection);
 	}
--- a/src/UI/MainMenuOptions.cpp
+++ b/src/UI/MainMenuOptions.cpp
@@ -73,16 +73,17 @@ void MainMenuOptions::init()
 
     auto& r = NAS2D::Utility<NAS2D::Renderer>::get();
     auto resolutions = r.getDisplayModes();
-    resolutions.erase(std::remove_if(std::begin(resolutions), std::end(resolutions),
-        [](const NAS2D::DisplayDesc& desc) {
-            const auto w = desc.width;
-            const auto h = desc.height;
-            const auto min_w = constants::MINIMUM_WINDOW_WIDTH;
-            const auto min_h = constants::MINIMUM_WINDOW_HEIGHT;
-            const auto is_invalid = w < min_w || h < min_h;
-            return is_invalid;
-        }),
-    std::end(resolutions));
+    resolutions.erase(
+        std::remove_if(
+            std::begin(resolutions),
+            std::end(resolutions),
+            [](const NAS2D::DisplayDesc& desc) {
+                const auto is_invalid = desc.width < constants::MINIMUM_WINDOW_WIDTH || desc.height < constants::MINIMUM_WINDOW_HEIGHT;
+                return is_invalid;
+            }
+        ),
+        std::end(resolutions)
+    );
 
 
 	{
--- a/src/UI/MainMenuOptions.cpp
+++ b/src/UI/MainMenuOptions.cpp
@@ -120,12 +120,6 @@ void MainMenuOptions::init()
     cbxSkipSplash.size(200, 0);
     cbxSkipSplash.click().connect(this, &MainMenuOptions::onOptionsChanged);
 
-    lblStartMaximized.text("Start Maximized");
-    lblStartMaximized.size(0, 0);
-
-    cbxStartMaximized.size(200, 0);
-    cbxStartMaximized.click().connect(this, &MainMenuOptions::onOptionsChanged);
-
     lblVSync.text("VSync");
     lblVSync.size(0, 0);
 
@@ -179,8 +173,6 @@ void MainMenuOptions::init()
     pnlControls.add(&sldrMusicVolume, right_edge - sldrMusicVolume.width() - border_right_width, lblMusicVolume.rect().startPoint().y() - sldrMusicVolume.height());
     pnlControls.add(&lblSkipSplash, left_edge, lblMusicVolume.rect().startPoint().y() + element_spacing);
     pnlControls.add(&cbxSkipSplash, right_edge - cbxSkipSplash.width(), sldrMusicVolume.rect().startPoint().y() + element_spacing);
-    pnlControls.add(&lblStartMaximized, left_edge, lblSkipSplash.rect().startPoint().y() + element_spacing);
-    pnlControls.add(&cbxStartMaximized, right_edge - cbxStartMaximized.width(), cbxSkipSplash.rect().startPoint().y() + element_spacing);
     add(&pnlControls, 0, 0);
 
     pnlButtons.add(&btnOk, 0, 0);
@@ -344,7 +336,6 @@ void MainMenuOptions::setConfigFromControls()
     cf.vsync(cbxVSync.checked());
     cf.fullscreen(cbxFullscreen.checked());
     cf.option("skip-splash", (cbxSkipSplash.checked() ? "true" : "false"));
-    cf.option("maximized", (cbxStartMaximized.checked() ? "true" : "false"));
 
     auto& mixer = NAS2D::Utility<NAS2D::Mixer>::get();
     mixer.musicVolume(cf.audioMusicVolume());
@@ -366,7 +357,5 @@ void MainMenuOptions::setControlsFromConfig()
     cbxFullscreen.checked(cf.vsync());
 
     const auto skipsplash_str = NAS2D::toLowercase(cf.option("skip-splash"));
-    const auto maximized_str = NAS2D::toLowercase(cf.option("maximized"));
     cbxSkipSplash.checked(skipsplash_str == "true" ? true : (skipsplash_str == "false" ? false : false));
-    cbxStartMaximized.checked(maximized_str == "true" ? true : (maximized_str == "false" ? false : false));
 }
--- a/src/UI/MainMenuOptions.cpp
+++ b/src/UI/MainMenuOptions.cpp
@@ -92,16 +92,19 @@ void MainMenuOptions::init()
 		const auto gfx_width = cf.graphicsWidth();
 		const auto gfx_height = cf.graphicsHeight();
         int resolution_index = 0;
+        SDL_DisplayMode cur_mode{};
+        SDL_GetWindowDisplayMode(underlyingWindow, &cur_mode);
+        NAS2D::DisplayDesc current_dimensions{cur_mode.w, cur_mode.h, cur_mode.refresh_rate};
         for(const auto& resolution : resolutions)
 		{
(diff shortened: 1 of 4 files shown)


DanRStevens:
Some merge conflicts will need to be resolved here. Let's try to merge #158 first, and then rebase this one and fix the merge conflicts.

cugone:
Rebased and resolved.

src/UI/MainMenuOptions.cpp
```diff
@@ -244,7 +243,8 @@ void MainMenuOptions::loadOptions() {
 }
 
 void MainMenuOptions::applyVideoChanges() {
-    if (videoOptionsChanged) {
+    if(videoOptionsChanged) {
```
DanRStevens: Spacing. Multiple occurrences.

src/UI/MainMenuOptions.cpp
```diff
@@ -70,27 +70,37 @@ void MainMenuOptions::init()
     cmbResolution.maxDisplayItems(1);
 
     cmbResolution.clearSelection();
+    { //Default resolution
```
DanRStevens: Odd formatting

src/UI/MainMenuOptions.cpp
```diff
@@ -70,27 +70,37 @@ void MainMenuOptions::init()
...
+        std::ostringstream ss;
+        ss << constants::MINIMUM_WINDOW_WIDTH << 'x' << constants::MINIMUM_WINDOW_HEIGHT << 'x' << 60;
+        cmbResolution.text(ss.str());
```
DanRStevens: Consider using [`std::to_string`](https://en.cppreference.com/w/cpp/string/basic_string/to_string) rather than `std::ostringstream`.

src/UI/MainMenuOptions.h
```diff
@@ -68,6 +68,5 @@ class MainMenuOptions : public Window
 
     bool optionsChanged = false;
     bool videoOptionsChanged = false;
-    bool inInit = true;
```
DanRStevens: Kind of a minor point for a PR, though it looks like this was removed in a commit before all uses of it were removed. That means the commit with a partial removal can not build.

src/UI/MainMenuOptions.cpp
cugone: Is the a difference?

DanRStevens: The `std::to_string` calls can be inlined, allowing this to be done in 1 line of code. The `ostringstream` needs to be declared first, then used, then have a `str()` extracted from it.

DanRStevens:
I was looking at this one and noticed random displays of the popup "You must restart the game for the video options to take effect". This problem seems to exist on master, so it not unique to this PR. I'm not sure if this PR was intended to fix that or not. It only says "fixes various bugs in the options menu". That's way to vague of a PR description for me to know what this PR is about. Could you please update the description of what this actually does?

Additionally, this PR contains large whitespace changes, which makes it very difficult to see at a glance what this PR is actually changing. We should avoid doing large whitespace changes in the same PR as a large functional change. Could you please rework this one to separate the whitespace changes from the functional changes? Maybe we could apply whitespace changes first, in their own PR, to fix the inconsistency. Postponing whitespace changes until later is also an option if that is easier.

Additionally, I created and merged #194, in case you want to build on top of that.

cugone:
I've edited the description. The merge from master screwed all kinds of things up. I'll split out the whitespace changes. It was mostly tabifying the document.

DanRStevens:
The description is much clearer now. Now I know what to expect when I look at the code.

ldicker83:
Haven't looked at the code yet on this one but skimming through the comments and I noticed this:

> "You must restart the game for the video options to take effect"

There should be no need for the game to be restarted for any video changes to take effect, there is nothing in the renderer that can't be done on the fly.

Anyway, I'll take a look at this at some point and try to make some sense of it (see my comments in another PR).

cugone:
@ldicker83 
How? There are no functions on `Renderer` to change the window size except on construction. Hence the dialog.

cugone:
@ldicker83 
@DanRStevens 

Determining what these functions do is not intuitive *at all*

```
    virtual void size(int w, int h) = 0;
    const Point_2df& size();
```

DanRStevens:
You're right. They are a bit inconsistent about packed/unpacked parameters. They should really be:
```cpp
    virtual void size(Vector<float> newSize) = 0;
    Vector<float> size();
```

:wink:

cugone:
Oh ha ha. :/

That said. I've implemented resolution changes without restarting.

cugone:
@DanRStevens 
@ldicker83 

Validating the display resolution to be no smaller than `1000x700` breaks any resolution selection below that value. The game works at 640x480. Recommend removing the validation function.

DanRStevens:
I would be fine if we dropped the minimum supported resolution. That may be an easier and more sensible change than removing any kind of minimum.

DanRStevens:
Just a reminder, I was asking for a rebase to make these changes easier to review. The large whitespace changes should either be squashed or moved to a new PR, depending on whether the changes were introduced here or pre-existing.

cugone:
Since the merge with master I'm having trouble picking out the `Whitespace` commit and blowing out the merge to get things back to  a readable state.

cugone:
Poking for re-review

cugone:
I think the history is fixed. There was one section that could not be split up into any smaller hunks and the changelist was so large I was getting lost on how to use the "edit hunk" feature, especially while staring at a CLI.

DanRStevens:
FYI:
```
src/UI/MainMenuOptions.cpp:92:14: warning: unused variable 'gfx_width' [-Wunused-variable]
                const auto gfx_width = cf.graphicsWidth();
                           ^
src/UI/MainMenuOptions.cpp:93:14: warning: unused variable 'gfx_height' [-Wunused-variable]
                const auto gfx_height = cf.graphicsHeight();
                           ^
```

src/UI/MainMenuOptions.cpp
```diff
@@ -74,35 +74,45 @@ void MainMenuOptions::init()
...
+    resolutions.erase(std::remove_if(std::begin(resolutions), std::end(resolutions),
+        [](const NAS2D::DisplayDesc& desc) {
+            const auto w = desc.width;
+            const auto h = desc.height;
+            const auto min_w = constants::MINIMUM_WINDOW_WIDTH;
+            const auto min_h = constants::MINIMUM_WINDOW_HEIGHT;
+            const auto is_invalid = w < min_w || h < min_h;
+            return is_invalid;
+        }),
+    std::end(resolutions));
```
DanRStevens: The formatting looks a bit inconsistent with the indentation for this block.

src/UI/MainMenuOptions.cpp
```diff
@@ -74,35 +74,45 @@ void MainMenuOptions::init()
...
-			SDL_DisplayMode cur_mode{};
-			SDL_GetDisplayMode(display_index, i, &cur_mode);
 			SDL_DisplayMode closest_mode{};
 			SDL_DisplayMode desired_mode{};
 			desired_mode.w = gfx_width;
 			desired_mode.h = gfx_height;
 			if(SDL_GetClosestDisplayMode(display_index, &desired_mode, &closest_mode))
 			{
-				if(cur_mode.w == closest_mode.w && cur_mode.h == closest_mode.h && cur_mode.refresh_rate == closest_mode.refresh_rate)
+                NAS2D::DisplayDesc closest{closest_mode.w, closest_mode.h, closest_mode.refresh_rate};
```
DanRStevens: Indentation: spaces versus tabs

src/UI/MainMenuOptions.cpp
```diff
@@ -92,23 +92,27 @@ void MainMenuOptions::init()
...
-				if(resolution == closest)
+				if(current_dimensions == closest)
 				{
 					//Set combobox to current dimensions of window, not desktop
 					currentResolutionSelection = resolution_index;
 				}
 			}
 			std::string resolution_str = std::to_string(resolution.width) + "x" + std::to_string(resolution.height) + "x" + std::to_string(resolution.refreshHz);
 			cmbResolution.addItem(resolution_str, resolution_index);
+            ++resolution_index;
```
DanRStevens: Indentation: Spaces versus tabs

cugone:
Weird. Doesn't warn in VS. Fixed now.

Should probably bump the warning level to 4.

src/UI/MainMenuOptions.cpp
```diff
@@ -69,30 +69,44 @@ void MainMenuOptions::init()
...
-            cmbResolution.addItem(ss.str(), i);
-            ss.str("");
-        }
-    }
-    cmbResolution.selectionChanged().connect(this, &MainMenuOptions::onVideoOptionsChanged);
+	cmbResolution.clearSelection();
+	{
+		const auto default_resolution = std::to_string(constants::MINIMUM_WINDOW_WIDTH) + 'x' + std::to_string(constants::MINIMUM_WINDOW_HEIGHT) + 'x' + std::to_string(60);
+		cmbResolution.text(default_resolution);
+	}
```
DanRStevens: If this is removed later, the removal could have been squashed with the addition.

src/UI/MainMenuOptions.cpp
cugone: I was trying to convey that the lambda went with and was part of the `eraase-remove` idiom calls.

What would you prefer?

src/UI/MainMenuOptions.cpp
cugone: I don't understand. This was removed in a later commit. Using the `Start Review` button instead of commit-by-commit only shows you the most-up-to-date versions of each file, thus, you would have never seen it's addition to comment on, just the deletion.

Also, I'm pretty sure it's addition was part of that big commit that I couldn't separate.

DanRStevens: I added the comment after viewing the removal.

My point was more that it could have been cleaned up before review.

src/UI/MainMenuOptions.cpp
DanRStevens: Well, to fully spill things, it might look like:
```cpp
	resolutions.erase(
		std::remove_if(
			std::begin(resolutions),
			std::end(resolutions),
			[](const NAS2D::DisplayDesc& desc) {
				const auto w = desc.width;
				const auto h = desc.height;
				const auto min_w = constants::MINIMUM_WINDOW_WIDTH;
				const auto min_h = constants::MINIMUM_WINDOW_HEIGHT;
				const auto is_invalid = w < min_w || h < min_h;
				return is_invalid;
			}
		),
		std::end(resolutions)
	);
```

----

On another matter, I don't think I would use all those `const` just to create a one time use short alias. It would be more straightforward to just use the original values in the comparison. No tracing of assignments to find where things came from. It makes the comparison line look a little longer, but even if that got spilled, it would still be shorter than creating the `const` aliases.

DanRStevens: Of course, if we merge that other PR for the comparisons, that would also really simplify the lambda.

src/UI/MainMenuOptions.cpp
cugone: Ah

src/UI/MainMenuOptions.cpp
cugone: Oh god no. That's arrow code right there.

I have to split them up because I don't care about the refresh rate and I don't want any false positives from `DisplayDesc::operator<` removing descriptions that were okay to keep because the refresh rates were different.

DanRStevens: Technically, I did nothing to the actual structure of the code. The nesting is exactly the same as before. The cyclomatic complexity is exactly the same as before. If the newly formatted code is "arrow code", or deeply nested code, then so was the original. The indentation merely makes it more obvious.

----

More to the point, the original code had indentation that was contrary to the structure. The last line should not have been indented the same as the first line. The second last line mixes two levels of nesting in an inconsistent way.

cugone: Fixed. It's only one extra line anyway.

cugone:
Quick question(s) before I get started on this: If I commit the white space changes to this PR, wouldn't cherry-picking *earlier* commits not contain them?

On that note, committing the white space changes to a *new* PR would have that new PR be a base to the cherry-picked commits in order for them to contain the new white space, right?

PR #160, opened 2020-01-30
Changes name of MINIMUM_WINDOW_WIDTH/HEIGHT to DEFAULT_WINDOW_WIDTH/HEIGHT

cugone:
Branched off #159. Only the last commit is new.

See title.

### DIFF
--- a/src/Constants/Numbers.h
+++ b/src/Constants/Numbers.h
@@ -36,8 +36,8 @@ namespace constants
 
 	const int DEFAULT_STARTING_MORALE = 600;
 
-	const int MINIMUM_WINDOW_WIDTH = 1000;
-	const int MINIMUM_WINDOW_HEIGHT = 700;
+	const int DEFAULT_WINDOW_WIDTH = 1000;
+	const int DEFAULT_WINDOW_HEIGHT = 700;
 
 	const float ROBOT_COM_RANGE = 15.0f;
 	const float COMM_TOWER_BASE_RANGE = 10.0f;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -80,8 +80,8 @@ int main(int argc, char *argv[])
 		// If no config file, set defaults.
 		if (!f.exists("config.xml"))
 		{
-			cf.graphicsWidth(constants::MINIMUM_WINDOW_WIDTH);
-			cf.graphicsHeight(constants::MINIMUM_WINDOW_HEIGHT);
+			cf.graphicsWidth(constants::DEFAULT_WINDOW_WIDTH);
+			cf.graphicsHeight(constants::DEFAULT_WINDOW_HEIGHT);
 			cf.fullscreen(false);
 		}
 


DanRStevens:
Why was this based off of #159? This looks like an independent change.

cugone:
One of the files that is changed in the last commit here is `MainMenuOptions.cpp`. So at least it requires #153 to be pushed first. That's probably what I should have based it off of instead.

Additionally, I forgot to change the referenced variables in `main.cpp` as well.

cugone:
Rebased

src/UI/MainMenuOptions.cpp
```diff
@@ -70,6 +70,11 @@ void MainMenuOptions::init()
     cmbResolution.maxDisplayItems(1);
 
     cmbResolution.clearSelection();
+    { //Default resolution
+        std::ostringstream ss;
+        ss << constants::DEFAULT_WINDOW_WIDTH << 'x' << constants::DEFAULT_WINDOW_HEIGHT << 'x' << 60;
+        cmbResolution.text(ss.str());
+    }
```
DanRStevens: This part looks like it comes from another branch (#159). I think it may have been included here by accident.

cugone: Notice #159 has the old name. This one is the new name. Because #159 is a prerequisite and they were created independently of each other, a merge conflict will need to be resolved.

cugone: Notice #159 has the old name. This one is the new name. Because #159 is a prerequisite and they were created independently of each other, a merge conflict will need to be resolved.

DanRStevens: Yes, but this change doesn't need to be on this branch at all. These 5 lines are showing up as a code addition. They have nothing to do with renaming a variable when that variable wasn't originally used here.

src/UI/MainMenuOptions.cpp
```diff
@@ -69,7 +69,6 @@ void MainMenuOptions::init()
     cmbResolution.size(200, 0);
     cmbResolution.maxDisplayItems(1);
 
-    cmbResolution.clearSelection();
```
DanRStevens: Err, why is this line being removed? This looks out of place.

cugone: That's one of those auto-commits where if there's no conflict, the change just gets merged without my ability to check it first.

cugone: But yeah, that line shouldn't have been removed.

cugone: It exists in #159 so merging this first should ensure it gets put back.

At the same time, now that I think about it, this code is run in the constructor. the resolution list is already going to be empty so clearing it before filling it won't make a difference.

I just tested the code with and without clearing the list first. No difference.

DanRStevens: This is no longer branched off of #159. It was rebased. It is now an independent branch with parallel development. If both branches modify the same line, it will result in a merge conflict.

More to the point, the change has nothing to do with this commit, nor with this branch even. This branch was about renaming a constant. It's out of place and confusing to have that modification here.

DanRStevens:
Discovered post merge that this PR broke the build. Reverting changes in PR #181.

The changes here can be reworked and presented in a new PR.

PR #161, opened 2020-01-30
Changes the default window dimensions to 1024x768

cugone:
Branches off #160. Only the last commit is new.

The parent branches fix the options menu such that any window size will work even, 640x480, but that's old school 90s support. This commit assumes you at least have a monitor from 2000.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,9 @@ Installation is simple: Extract the contents of the ZIP Archive into any folder.
 As this is not an installer package it is not distributed with any necessary dependencies. Please see the following sections for the requirements needed for your particular OS of choice.
 
 ### Windows Dependencies
-OutpostHD is built using Microsoft's Visual C++ 2017. If you haven't already, you may need to download and install the redistributable package from Microsoft. Follow this link and download the 32-Bit version of the Visual C++ runtime:
+OutpostHD is built in C++ using Microsoft's Visual Studio 2019. If you haven't already, you may need to download and install the redistributable package from Microsoft. Follow this link and download the 32-Bit version of the Visual C++ runtime:
 
-https://outpost2.net/files/ophd/VC_redist.x86.exe
+https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads
 
 ### MacOS X Dependencies
 The only current build of OutpostHD is Windows but as I understand MacOS X development and program distribution, MacOS distributions come as a 'bundle' with all necessary dependencies packaged in. That stated, there aren't any official osX builds (yet) but we're working on it!
--- a/src/Constants/Numbers.h
+++ b/src/Constants/Numbers.h
@@ -36,8 +36,8 @@ namespace constants
 
 	const int DEFAULT_STARTING_MORALE = 600;
 
-	const int DEFAULT_WINDOW_WIDTH = 1000;
-	const int DEFAULT_WINDOW_HEIGHT = 700;
+	const int DEFAULT_WINDOW_WIDTH = 1024;
+	const int DEFAULT_WINDOW_HEIGHT = 768;
 
 	const float ROBOT_COM_RANGE = 15.0f;
 	const float COMM_TOWER_BASE_RANGE = 10.0f;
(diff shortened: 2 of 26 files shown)


DanRStevens:
The base branch was merged, and then undone due to build problems. As such, this branch no longer has a stable base to build on top of. It may either need rework, or perhaps be dropped.

----

I've noticed the odd resolution parameters. I can understand the desire to change them. I wonder if they may have been chosen slightly smaller due to window border or task bar allowances. If the game is run in windowed mode, less than the full screen resolution will be available. @ldicker83, can you comment on this?

----

If the game defaults to windowed mode, maybe we shouldn't have a hardcoded constant for the default size. Maybe we should just try to run maximized, and work with whatever that is, assuming it's not less than some minimum. Though, considering you've successfully tested the game at 640x480, we probably won't be running into problems with a true minimum size restriction.

I assume there should be an easy enough way to determine a maximized size, for the current desktop resolution. I suggest dynamically determining the default initial size, for when the configuration settings don't specify one, or when they specify an invalid size.

cugone:
> 
> If the game defaults to windowed mode, maybe we shouldn't have a hardcoded constant for the default size. Maybe we should just try to run maximized, and work with whatever that is, assuming it's not less than some minimum. Though, considering you've successfully tested the game at 640x480, we probably won't be running into problems with a true minimum size restriction.

This is an easy fix, though running maximized and NOT borderless sometimes causes graphical issues with the backbuffer sizes. Maybe the reason an odd resolution was chosen was because at the time figuring out how to create a borderless window was unknown.

> 
> I assume there should be an easy enough way to determine a maximized size, for the current desktop resolution. I suggest dynamically determining the default initial size, for when the configuration settings don't specify one, or when they specify an invalid size.

SDL and Windows both have functions for getting the current desktop size; forcing the default window to that size isn't hard. Again, should default to make it a borderless window.

DanRStevens:
I would like to be able to play fullscreen borderless.

There would need to be a way to exit full screen. Currently window sizing is controlling by dragging the edges of the window, or with the minimize and maximize butters. If those are removed when going borderless, there will need to be some kind of replacement.

What is the specific code needed to determine full screen size with/without borders?

cugone:
> What is the specific code needed to determine full screen size with/without borders?

See for SDL: https://wiki.libsdl.org/SDL_CreateWindow
Windows is a little more involved. Raymond Chen posted an example on his blog 10 years ago: https://devblogs.microsoft.com/oldnewthing/20100412-00/?p=14353

None of these methods resize the backbuffers so you will get stretching.

> There would need to be a way to exit full screen. Currently window sizing is controlling by dragging the edges of the window, or with the minimize and maximize butters. If those are removed when going borderless, there will need to be some kind of replacement.

The Windows API allows for disabling resizing a window via dragging. I don't know about SDL.

Windowed Fullscreen (borderless window the size of the desktop) is the convention these days. Exclusive Fullscreen on modern Windows provides no benefit. In fact, Windows 10 just gives you a Borderless Fullscreen window even if you ask for exclusive.

By default `ALT+ENTER` asks a window to toggle full screen. Cursory google searches say SDL does the same.

> I would like to be able to play fullscreen borderless.

Me too. See above. When I check Fullscreen in the options menu Windows 10 gives me a Borderless Fullscreen Window by default (I can tell because I can set breakpoints and the game doesn't take control my computer away from me.)

DanRStevens:
Thank you for the reference links.

I don't mean to disable window resizing. I mean to enable it when there is no longer a border or buttons to click on.

Hotkeys can work, though if they're not intuitive, people could get stuck in fullscreen, and maybe they don't want to exit the game to access their desktop.

Currently on Linux the maximize button leaves the game in full screen windowed mode, and `<ALT>+<ENTER>` does nothing.

cugone:
Generally, you only go borderless in a full screen context. Otherwise you leave the window borders and titlebar. In any case, using the options menu to select an appropriate Resolution is the conventional way to change the size of the window *because* it's assumed you have fullscreen turned on and have no way to access the maximize button or borders to drag.

cugone:
Closing because this needs to be reworked. See #196 for a summary.

PR #167, opened 2020-02-02
Structure Integrity feature Stub

cugone:
Ref: https://github.com/ldicker83/OPHD/issues/20

This is an initial implementation for the feature. The functions are stubbed out but do effectively nothing. In addition, one requirement is to react to different difficulty modifiers (https://github.com/ldicker83/OPHD/issues/25), but that is a separate issue.

Further game design discussion on individual structure gameplay mechanics will be required.

### DIFF
--- a/src/Things/Structures/Structure.cpp
+++ b/src/Things/Structures/Structure.cpp
@@ -182,6 +182,14 @@ void Structure::activate()
 }
 
 
+int Structure::calculateIntegrity() {
+	return mIntegrity - calculateIntegrityDecayRate();
+}
+
+int Structure::calculateIntegrityDecayRate() {
+	return 0;
+}
+
 bool Structure::disabledByIntegrity() const
 {
 	return mIntegrity < 35;
--- a/src/Things/Structures/Structure.cpp
+++ b/src/Things/Structures/Structure.cpp
@@ -7,8 +7,10 @@
 
 #include <random>
 
-std::random_device rd;
-std::mt19937 generator(rd());
+namespace {
+	std::random_device rd;
+	std::mt19937 generator(rd());
+}
 
 /**
  * Translation table for Structure States.
--- a/src/Things/Structures/Structure.cpp
+++ b/src/Things/Structures/Structure.cpp
@@ -5,6 +5,10 @@
 
 #include "../../Constants.h"
 
+#include <random>
+
+std::random_device rd;
+std::mt19937 generator(rd());
 
 /**
  * Translation table for Structure States.
@@ -178,10 +182,33 @@ void Structure::activate()
 }
 
 
+bool Structure::disabledByIntegrity() const
+{
+	return mIntegrity < 35;
+}
+
+bool Structure::destroyedByIntegrity() const
+{
+	float destroyChance = 1.0f - mIntegrity / 25.0f;
+	// uniform_real_distributions are in a half-closed range [lower, upper).
+	// std::nextafter allows the range to be [lower, upper]
+	std::uniform_real_distribution<double> dist(0.0, std::nextafter(1.0, 2.0));
+	return dist(generator) < destroyChance;
+}
+
 void Structure::update()
 {
 	if (disabled() || destroyed()) { return; }
 	incrementAge();
+	mIntegrity = calculateIntegrity();
+	if(disabledByIntegrity())
+	{
+		disable(DISABLED_STRUCTURAL_INTEGRITY);
+	}
+	else if(destroyedByIntegrity())
+	{
+		destroy();
+	}
 }
 
 
--- a/src/Things/Structures/Structure.cpp
+++ b/src/Things/Structures/Structure.cpp
@@ -197,10 +197,8 @@ bool Structure::disabledByIntegrity() const
 
 bool Structure::destroyedByIntegrity() const
 {
-	float destroyChance = mIntegrity <= 25 ? 1.0f - mIntegrity / 25.0f : 0.0f;
-	// uniform_real_distributions are in a half-closed range [lower, upper).
-	// std::nextafter allows the range to be [lower, upper]
-	std::uniform_real_distribution<double> dist(0.0, std::nextafter(1.0, 2.0));
+	int destroyChance = mIntegrity < 25 ? 100 - (mIntegrity * 100 / 25) : 0;
+	std::uniform_int_distribution<int> dist(0, 100);
 	return dist(generator) < destroyChance;
 }
 
--- a/src/Things/Structures/Structure.cpp
+++ b/src/Things/Structures/Structure.cpp
@@ -208,7 +208,7 @@ void Structure::update()
 {
 	if (disabled() || destroyed()) { return; }
 	incrementAge();
-	mIntegrity = calculateIntegrity();
+	integrity(calculateIntegrity());
 	if(disabledByIntegrity())
 	{
 		disable(DISABLED_STRUCTURAL_INTEGRITY);
--- a/src/Things/Structures/Structure.cpp
+++ b/src/Things/Structures/Structure.cpp
@@ -197,7 +197,7 @@ bool Structure::disabledByIntegrity() const
 
 bool Structure::destroyedByIntegrity() const
 {
-	float destroyChance = 1.0f - mIntegrity / 25.0f;
+	float destroyChance = mIntegrity <= 25 ? 1.0f - mIntegrity / 25.0f : 0.0f;
 	// uniform_real_distributions are in a half-closed range [lower, upper).
 	// std::nextafter allows the range to be [lower, upper]
 	std::uniform_real_distribution<double> dist(0.0, std::nextafter(1.0, 2.0));
--- a/src/Things/Structures/Structure.h
+++ b/src/Things/Structures/Structure.h
@@ -158,6 +158,7 @@ protected:
 	void setPopulationRequirements(const PopulationRequirements& pr) { mPopulationRequirements = pr; }
 
 	virtual int calculateIntegrity() = 0;
+	virtual int calculateIntegrityDecayRate() = 0;
 private:
 	Structure() = delete;
 
--- a/src/Things/Structures/Structure.h
+++ b/src/Things/Structures/Structure.h
@@ -6,6 +6,8 @@
 #include "../../PopulationPool.h"
 #include "../../ResourcePool.h"
 
+#include <algorithm>
+
 class Structure: public Thing
 {
 public:
--- a/src/Things/Structures/Structure.h
+++ b/src/Things/Structures/Structure.h
@@ -135,6 +135,8 @@ public:
 	void update();
 	virtual void think() {}
 
+	int integrity() const { return mIntegrity; }
+	int integrityDecayRate() const { return mIntegrityDecayRate; }
 protected:
 	friend class StructureCatalogue;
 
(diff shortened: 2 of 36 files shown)


src/Things/Structures/Structure.cpp
```diff
@@ -5,6 +5,10 @@
 
 #include "../../Constants.h"
 
+#include <random>
+
+std::random_device rd;
+std::mt19937 generator(rd());
```
DanRStevens: For purposes such as network sync or game replay saving, we should be careful about uncontrolled use of random number generators. It might be best to have a central place to store a random number generator for the whole game, and to consistently use the same generator.

src/Things/Structures/Structure.cpp
```diff
@@ -178,10 +182,33 @@ void Structure::activate()
...
+{
+	return mIntegrity < 35;
+}
+
+bool Structure::destroyedByIntegrity() const
+{
+	float destroyChance = 1.0f - mIntegrity / 25.0f;
+	// uniform_real_distributions are in a half-closed range [lower, upper).
+	// std::nextafter allows the range to be [lower, upper]
+	std::uniform_real_distribution<double> dist(0.0, std::nextafter(1.0, 2.0));
```
DanRStevens: I can't help but think there must be a simpler more straightforward way to do this.

src/Things/Structures/Structure.h
```diff
@@ -157,6 +157,7 @@ class Structure: public Thing
 
 	void setPopulationRequirements(const PopulationRequirements& pr) { mPopulationRequirements = pr; }
 
+	virtual int calculateIntegrity() = 0;
```
DanRStevens: I noticed a distinct lack of other direct pure `virtual` methods being declared in this class.

Given the large number of structures, and given that many of them will have very similar behaviour, I suspect much of the building integrity behaviour will be implemented in the base class, with just a few parameter values, where the parameter values can vary slightly between classes. That may be an easier way to control behaviour. It would certainly require less duplication of code in derived classes.

I don't oppose this method being `virtual`, but I suspect we don't want it to be pure virtual. I also don't think we need or want to override it in the vast majority of classes. There probably should be a good default implementation that comes from the base class.

src/Things/Structures/Structure.cpp
cugone: I was thinking the same thing. You really shouldn't have more than one RNG per thread and it certainly shouldn't be initalized more than once (or reinitialzied, if you store the seed in the save game file.)

src/Things/Structures/Structure.cpp
cugone: This function fulfills the requirement:
 - Logic sets a destroyed state based on a chance when integrity is lower than 25%. The lower the integrity, the higher the chance of the building collapsing.

As the value gets closer to `0.0` the chance to destroy gets closer to `1.0`, giving the RNG more of the range to pick from and a higher chance to return `true`, which will set the state of the structure to destroy during `update`.

cugone: The issue with `uniform_real_distribution` being a half-open range is "because the standard says so." Can't be helped unfortunately. :/

cugone: However, the algorithm is wrong. Should be
```
float destroyChance = 1.0f - mIntegrity / 100.0f;
std::uniform_real_distribution<double> dist(0.0, std::nextafter(1.0, 2.0));
return dist(generator) < destroyChance;
```

src/Things/Structures/Structure.h
cugone: Removed Pure Virtual designation. Default behavior is to `return mIntegrity - calculateIntegrityDecayRate();` where `calculateIntegrityDecayRate` by default just returns `0`.

cugone:
I've opted to leave the derived structure overrides until the gameplay mechanics have been hashed out.

If it's decided that a specific structure should do the default (not decay), the function override can be removed.

As is, all structures already do the default in their overridden functions anyway.

src/Things/Structures/Structure.cpp
cugone: Opted for `float destroyChance = mIntegrity <= 25 ? 1.0f - mIntegrity / 25.0f : 0.0f;` instead.
This way the structures can only destroy by decay if they are actually at 25 or less integrity as the spec requires. The other algorithm above makes it so there's a chance for a perfectly healthy 99% integrity building to destroy itself, albeit a 1% chance.

Maybe that can be a feature when difficulty is implemented. 😈

DanRStevens: The half open range might not actually be a problem. The difference should amount to whether 25% integrity means the first point at which a structure can collapse, or the last point at which a structure doesn't collapse. As the point was arbitrary to begin with, the distinction is quite arbitrary. I'm kind of thinking the `std::nextafter` could just be dropped.

With the ternary, the `std::nextafter` difference amounts to the different between `<=` or strict `<`.

----

With the new formula, you might as well use `<`. The result is the same, though it avoids generating random numbers when the probability of collapse is 0.

----

Was also thinking the labelling of `destroyChance` looked a bit backwards before. Not something I'm quite worried about at this exact moment. The details of the formula used is subject to change over time.

DanRStevens:
Sounds a bit backwards. If a structure wants non-default behaviour, then it needs to override the virtual method. Structures should by default use the parent class methods, and only override if they need special rules. That should result in very few structures actually overriding that method.

cugone:
Yeah, I realize it's a little backwards, but I don't want to make another massive 34-file changeset. :P

DanRStevens:
You could run a `git revert`, which creates a new commit with the reverse changes of the original commit. Alternatively, you should be able to run a rebase and just delete the commit.

src/Things/Structures/Structure.cpp
DanRStevens: I was just looking for existing practice. The `Population` code also just creates a random number generator and uses it. As such, I would consider this an external concern, though we really should be opening an issue to track this, and coming up with a general solution so all game components have access to a random number generator.

src/Things/Structures/Structure.cpp
DanRStevens: The `Population` code generates integers in the range `0` to `100`:
```cpp
std::uniform_int_distribution<int> pop_distribution(0, 100);
```

Perhaps we should do similar for the percent chance a building will be destroyed. It would be easy enough to convert the formula to use percentage instead of a `float` value:
```cpp
int destroyChance = mIntegrity < 25 ? 100 - (mIntegrity * 100 / 25) : 0;
```

DanRStevens:
I was just thinking, until there is a way to repair buildings, we probably need to leave integrity decay disabled.

DanRStevens:
Actually, I just had a more troubling thought. The requirements for building integrity decay were never really fleshed out. As such, it's really hard to determine if the new code would support the requirements for building decay. I don't really have a clear acceptance criteria for this code. That kind of makes me feel a little irresponsible for giving the go ahead to start creating implementation. On the other hand, it gives us something to look at, which may help guide us in figuring out what the requirements should be.

We'll probably need to involve Leeor in the discussion to figure out what the requirements are in greater detail.

cugone:
> 
> 
> Actually, I just had a more troubling thought. The requirements for building integrity decay were never really fleshed out. As such, it's really hard to determine if the new code would support the requirements for building decay. I don't really have a clear acceptance criteria for this code. That kind of makes me feel a little irresponsible for giving the go ahead to start creating implementation. On the other hand, it gives us something to look at, which may help guide us in figuring out what the requirements should be.
> 
> We'll probably need to involve Leeor in the discussion to figure out what the requirements are in greater detail.
> 
> Edit: Or provide suggestions to Leeor about how we think it should perhaps work.

Yeah. That's why the code defaults integrity to 100 and reduces it by 0 every update. It effectively does nothing.

This way the code is there for when individual building decay rates are established; we only have to add the implementation for `calculateDecayRate` and it'll magically start working.

cugone:
> 
> 
> I was just thinking, until there is a way to repair buildings, we probably need to leave integrity decay disabled.

Good point.

If we're going by the original game as a metric, repair wasn't a thing. They just destroyed themselves and you had to rebuild it. :P

ldicker83:
I added a few questions and a comment regarding the stub functions that don't match the rest of the code style (brace placement, I know it's a religious thing but it should really match the rest of the codebase regardless of style preference).

I'm not sure I understand the logic of destruction chance of a structure, so if I can get that resolved, the questions regarding the calculateIntegrityXXX() function return values resolved and the stylistic choices I'll pull this locally and check numbers for it and ultimately merge.

cugone:
> 
> I added a few questions and a comment regarding the stub functions that don't match the rest of the code style (brace placement, I know it's a religious thing but it should really match the rest of the codebase regardless of style preference).

You did? I'm not seeing them.

> 
> I'm not sure I understand the logic of destruction chance of a structure, so if I can get that resolved, the questions regarding the calculateIntegrityXXX() function return values resolved and the stylistic choices I'll pull this locally and check numbers for it and ultimately merge.

The logic is basically:
1. Update the integrity by subtracting a building's decay rate from its current integrity value.
1. Clamp integrity to the closed range `[0,100]`
1. Is the integrity of a building below 35?
    1. Yes. Disable the building
1. Else: Is the integrity 25 or below?
    1. Yes.
        1. Calculate a chance to destroy the building the closer the integrity gets to zero with 25 integrity being 0% chance and 0 integrity being 100%.
        1. Was the building destroyed?
            1. Yes. Set the destroyed state.

Each building calculates its own decay rate based on its own utility. E.g. Agridome decay could be based on how long the food has been in storage; CHAP based on age; Warehouses and Storage tanks based on how full they are; Residences based on the over-crowded factor; etc. Of course, also multiplying by the difficulty factor where appropriate.

src/Things/Structures/Structure.cpp
cugone: Added https://github.com/lairworks/nas2d-core/issues/453 to track.

cugone:
I was playing OP1 last night and had a thought. This could easily be used to implement the Seed Lander/Factory destruction. Their Utility is just determining if it's past Turn 150 or not:

```
int calculateIntegrityDecayRate()
{
    int currentTurn = getCurrentTurnCount();
    return 150 < currentTurn ? 100 : 0;
}
```

cugone:
https://github.com/lairworks/nas2d-core/issues/453 needs to be addressed before checks will pass.

DanRStevens:
Looks like there is a missing include for `<algorithm>` which causes GCC to fail on [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp):
```
src/UI/../Things/Structures/Structure.h: In member function ‘void Structure::integrity(int)’:
src/UI/../Things/Structures/Structure.h:164:55: error: ‘clamp’ is not a member of ‘std’
  void integrity(int newIntegrity) { mIntegrity = std::clamp(newIntegrity, 0, 100); }
                                                       ^~~~~
```

----

In a related matter, I noticed CircleCI hasn't been running since the OPHD repository was transferred to the OutpostUniverse organization. I've just gone and enabled CircleCI for the repository's new home.

DanRStevens:
To solve the linking problem, the random number generator declaration can be wrapped in an anonymous namespace:
```cpp
namespace {
	std::random_device rd;
	std::mt19937 generator(rd());
}
```

That will allow code to link for now. We can defer until later on collecting all uses of random numbers to a single generator.

DanRStevens:
Closing for rework. I think the code is ok now. We probably want to work on the presentation a bit though. This thread has quite a bit of back and forth discussion and design changes. We should rework, and then present the final changes in a new PR.

PR #169, opened 2020-02-02
Fixes unintentional bug when MoraleModifier is given negative numbers

cugone:
It also has no reason to be a full class, a POD struct is fine.

Part of splitting up #155 into smaller commits.

### DIFF
--- a/src/Population/Morale.h
+++ b/src/Population/Morale.h
@@ -5,19 +5,16 @@
 /**
  * Morale modifier values.
  */
-class MoraleModifier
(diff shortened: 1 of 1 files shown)


src/Population/Morale.h
```diff
@@ -5,19 +5,16 @@
 /**
  * Morale modifier values.
  */
-class MoraleModifier
+struct MoraleModifier
 {
 public:
```
DanRStevens: Now that this is a struct, the use of `public` is redundant.

PR #171, opened 2020-02-02
Fixes Unreachable Code warnings

cugone:
Part of splitting up #155 into smaller commits.

Unreachable code in GraphWalker.cpp: This is not an exceptional case. Logging the problem is sufficient.
Unreachable code in RobotPool.cpp: When returning from a switch-case, break statements are unreachable.

### DIFF
--- a/src/GraphWalker.cpp
+++ b/src/GraphWalker.cpp
@@ -39,7 +39,7 @@ static bool validConnection(Structure* src, Structure* dst, Direction _d)
 	if (src == nullptr || dst == nullptr)
 	{
 		#ifdef DEBUG
-		throw std::runtime_error("GraphWalker::validConnection() was passed a NULL Pointer.");
+		std::cout << "GraphWalker::validConnection() was passed a NULL Pointer." << std::endl;
 		#endif
 		return false;
 	}
--- a/src/GraphWalker.cpp
+++ b/src/GraphWalker.cpp
@@ -38,10 +38,7 @@ static bool validConnection(Structure* src, Structure* dst, Direction _d)
 {
 	if (src == nullptr || dst == nullptr)
 	{
-		#ifdef DEBUG
-		std::cout << "GraphWalker::validConnection() was passed a NULL Pointer." << std::endl;
-		#endif
-		return false;
+		throw std::runtime_error("GraphWalker::validConnection() was passed a NULL Pointer.");
 	}
 	if (_d == DIR_UP || _d == DIR_DOWN)
 	{
--- a/src/RobotPool.cpp
+++ b/src/RobotPool.cpp
@@ -143,23 +143,22 @@ bool RobotPool::robotAvailable(RobotType _type)
 	switch (_type)
 	{
 	case ROBOT_DIGGER:
+	{
 		return getDigger() != nullptr;
-		break;
-
-	case ROBOT_DOZER:
-		return getDozer() != nullptr;
-		break;
-
-	case ROBOT_MINER:
-		return getMiner() != nullptr;
-		break;
-
-	default:
-		return false;
-		break;
 	}
-
-	return false;
+	case ROBOT_DOZER:
+	{
+		return getDozer() != nullptr;
+	}
+	case ROBOT_MINER:
+	{
+		return getMiner() != nullptr;
+	}
+	default:
+	{
+		return false;
+	}
+	}
 }
 
 


src/GraphWalker.cpp
```diff
@@ -39,7 +39,7 @@ static bool validConnection(Structure* src, Structure* dst, Direction _d)
 	if (src == nullptr || dst == nullptr)
 	{
 		#ifdef DEBUG
-		throw std::runtime_error("GraphWalker::validConnection() was passed a NULL Pointer.");
+		std::cout << "GraphWalker::validConnection() was passed a NULL Pointer." << std::endl;
```
DanRStevens: I'm not so fond of this. I think an exception would be better here.

I'm also not so fond of the surrounding `#ifdef`. I think that's the real source of the problem.

src/RobotPool.cpp
```diff
@@ -143,23 +143,22 @@ bool RobotPool::robotAvailable(RobotType _type)
...
-
+	}
 	default:
+	{
 		return false;
-		break;
 	}
-
-	return false;
+	}
```
DanRStevens: Minor musing. I'm wondering if this might be better removing the `default` case, and putting the `return false;` outside of the switch. Not sure which would be better.

Initially I missed seeing the `default` case and wondered if this might not return a value in some cases. In that sense, having it outside the `switch` made the return result more obvious. On the other hand, using `default` might say something about the cases which are specifically handled.

I rarely use `switch`, so I don't really have any guidelines on making such a decision.

src/GraphWalker.cpp
cugone: It is. But I didn't write it, and at the time I was concerned with just getting rid of the warning. :wink:

I disagree on the exception. It's not a fatal unrecoverable state that a `nullptr` is passed in; it's just another case where the connection is not valid and `false` should be returned.

src/RobotPool.cpp
cugone: Conventional guidelines are to A) Make a `case` for every value of your switch variable if possible (duplicates not withstanding, but document it; and B) Always provide a `default` case.

In fact, Visual Assist even has a refactor option "Add missing case statements" for uses with `enum` and `enum class` types that does both A and B. If you are missing a statement it adds it, and it adds the `default` case if it is missing.

By providing a `default` any value not specified in the `case` labels will jump to the default. In this case if the value is not one of `ROBOT_DOZER`, `ROBOT_DIGGER`, `ROBOT_MINER`, `false` will be returned.

Additionally, I have a personal rule that if the range of your switch variable can ever change, loudly announce with a `MessageWindow` and a fatal error/`exit(0)` in the default case it when it does without your knowledge.

In this case, the number of robot types could change (I remember there were airbots in the original manual and research tree, but they were never implemented in-game.) and there should be a loud failure that the number of the type of robots has changed and they aren't handled in a `switch` case.

See
https://github.com/cugone/Abrams2019/blob/870dcd04d6a8fa8e5132efb015126d0de6935ec0/Engine/Code/Engine/Core/KerningFont.cpp#L858
for an example of announcing that a third-party API has changed and the new value isn't being handled; and
https://github.com/cugone/Abrams2019/blob/870dcd04d6a8fa8e5132efb015126d0de6935ec0/Engine/Code/Engine/Renderer/DirectX/DX11.cpp#L673
for using sensible defaults when needed and failing when the default is reached due to required state not being met.

src/GraphWalker.cpp
DanRStevens: Well, technically, exceptions are not unrecoverable. You can catch them. But that's beside the point here.

----

Being able to check for `nullptr` and doing nothing doesn't mean there's no problem. It could indicate the calling code is doing something stupid and needs to know about it. In this case, it really doesn't make sense to try and check for a path between two structures if you're not passing two valid structures.

If this ever happens, it's quite clearly a programming error, and the code should complain loudly. An exception is much better than a relatively silent log message, which could easily be missed.

src/RobotPool.cpp
DanRStevens: I think I would prefer an exception for a `MessageWindow` or `exit(0)`, though yes, it should loudly announce when not all cases are handled. I believe D added `final` to `switch` statements as a way of enforcing all cases are handled, and would result in a compiler error if they weren't.

PR #173, opened 2020-02-02
static_cast Narrowing conversions

cugone:
Part of splitting #155 up into smaller commits

### DIFF
--- a/src/States/MainReportsUiState.cpp
+++ b/src/States/MainReportsUiState.cpp
@@ -134,13 +134,13 @@ static void drawPanel(Renderer& _r, Panel& _p)
 
 		if (_p.UiPanel) { _p.UiPanel->update(); }
 
-		_r.drawText(*BIG_FONT_BOLD, _p.Name, _p.TextPosition.x(), _p.TextPosition.y(), 185, 185, 0);
-		_r.drawImage(*_p.Img, _p.IconPosition.x(), _p.IconPosition.y(), 1.0f, 185, 185, 0, 255);
+		_r.drawText(*BIG_FONT_BOLD, _p.Name, static_cast<float>(_p.TextPosition.x()), static_cast<float>(_p.TextPosition.y()), 185, 185, 0);
+		_r.drawImage(*_p.Img, static_cast<float>(_p.IconPosition.x()), static_cast<float>(_p.IconPosition.y()), 1.0f, 185, 185, 0, 255);
 	}
 	else
 	{
-		_r.drawText(*BIG_FONT_BOLD, _p.Name, _p.TextPosition.x(), _p.TextPosition.y(), 0, 185, 0);
-		_r.drawImage(*_p.Img, _p.IconPosition.x(), _p.IconPosition.y(), 1.0f, 0, 185, 0, 255);
+		_r.drawText(*BIG_FONT_BOLD, _p.Name, static_cast<float>(_p.TextPosition.x()), static_cast<float>(_p.TextPosition.y()), 0, 185, 0);
+		_r.drawImage(*_p.Img, static_cast<float>(_p.IconPosition.x()), static_cast<float>(_p.IconPosition.y()), 1.0f, 0, 185, 0, 255);
 	}
 }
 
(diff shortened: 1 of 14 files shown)


src/States/MapViewStateHelper.cpp
```diff
@@ -187,8 +187,9 @@ bool validLanderSite(Tile* t)
 	}
 
 	// bleh, direct copy from Tile::distanceTo()
-	int _x = t->x() - ccLocationX(), _y = t->y() - ccLocationY();
-	float _dist = sqrt((_x * _x) + (_y * _y));
+	int _x = t->x() - ccLocationX();
+	int _y = t->y() - ccLocationY();
```
DanRStevens: Technically this is sneaking in an unrelated change. This should probably have been done in a separate commit. No issue with the change though.

src/States/PlanetSelectState.cpp
```diff
@@ -36,8 +36,14 @@ class Explosion
 			mTimer.reset();
 		}
 
-		r.drawSubImageRotated(mSheet, x, y, (mFrame % 8) * 128, (mFrame / 8) * 128, 128, 128, 270.0f);
-		//r.drawSubImage(mSheet, x, y, (mFrame % 8) * 128, (mFrame / 8) * 128, 128, 128);
+		auto rasterX = static_cast<float>(x);
+		auto rasterY = static_cast<float>(y);
+		auto posX = static_cast<float>((mFrame % 8u) * 128.0f);
+		auto posY = static_cast<float>((mFrame / 8u) * 128.0f);
```
DanRStevens: I wonder if the scope of these `static_casts` are perhaps not quite right. The inner expression appears to be an `unsigned int` `*` `float`. Maybe the cast should be wrapping the `unsigned int` portion. Alternatively, remove the `.0f` from the `128.0f`, and maybe the `u` as well, and just do the whole internal calculation as an integer.

src/UI/Core/TextField.cpp
```diff
@@ -139,13 +139,13 @@ void TextField::onTextInput(const std::string& _s)
 
 	if (mMaxCharacters > 0 && text().length() == mMaxCharacters) { return; }
 
-	int prvLen = text().length();
+	std::size_t prvLen = text().length();
```
DanRStevens: No issue with this change. I actually prefer this sort of solution. Just want to point out that many other places currently using `static_cast` may be due to an improper type being used. We should at some point look at the uses of `static_cast` and see if a change in type would perhaps work better.

DanRStevens: Just noticed this change is also included in #174. It maybe fits better in the other PR.

src/States/PlanetSelectState.cpp
cugone: I have no idea why I did that. With proper use of parenthesis Integer-to-float promotion should take over: `auto posX = (mFrame % 8u) * 128.0f` and `auto posY = (mFrame / 8u) * 128.0f`

cugone: Fixed

src/UI/Core/TextField.cpp
cugone: Done

cugone:
Moved `TextField.cpp` change to #174.

src/States/MainReportsUiState.cpp
```diff
@@ -206,7 +206,7 @@ void MainReportsUiState::initialize()
 	Panels[PANEL_SPACEPORT].Name = "Space Ports";
 
 	Renderer& r = Utility<Renderer>::get();
-	setPanelRects(r.width());
+	setPanelRects(static_cast<int>(std::floor(r.width())));
```
DanRStevens: Ahh, this caused the build to break on Clang since there is no `<cmath>` include for `std::floor`.

DanRStevens: Btw, why was `std::floor` added here?

cugone: I haven't the foggiest idea why I added it.

DanRStevens: Fixed in #180.

PR #177, opened 2020-02-02
MapViewStateDraw refactor

cugone:
Refactors MapViewStateDraw file to be more readable and reduce duplicate function calls

### DIFF
--- a/src/States/MapViewStateDraw.cpp
+++ b/src/States/MapViewStateDraw.cpp
@@ -254,62 +254,37 @@ void MapViewState::drawNavInfo()
 {
 	Renderer& r = Utility<Renderer>::get();
 
-	// Up / Down
-	if (isPointInRect(MOUSE_COORDS, MOVE_DOWN_ICON))
-	{
-		r.drawSubImage(mUiIcons, MOVE_DOWN_ICON.x(), MOVE_DOWN_ICON.y(), 64, 128, 32, 32, 255, 0, 0, 255);
-	}
-	else
-	{
-		r.drawSubImage(mUiIcons, MOVE_DOWN_ICON.x(), MOVE_DOWN_ICON.y(), 64, 128, 32, 32);
-	}
+	NAS2D::Rectangle_2d currentIconBounds = MOVE_DOWN_ICON;
+	bool isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
+	NAS2D::Color iconColor{255, 255, 255, 255};
+	NAS2D::Color iconHighlightColor{255, 0, 0, 255};
+	NAS2D::Color color = isMouseInIcon ? iconHighlightColor : iconColor;
+	r.drawSubImage(mUiIcons, currentIconBounds.x(), currentIconBounds.y(), 64, 128, 32, 32, color.red(), color.green(), color.blue(), color.alpha());
 
-	if (isPointInRect(MOUSE_COORDS, MOVE_UP_ICON))
-	{
-		r.drawSubImage(mUiIcons, MOVE_UP_ICON.x(), MOVE_UP_ICON.y(), 96, 128, 32, 32, 255, 0, 0, 255);
-	}
-	else
-	{
-		r.drawSubImage(mUiIcons, MOVE_UP_ICON.x(), MOVE_UP_ICON.y(), 96, 128, 32, 32);
-	}
+	currentIconBounds = MOVE_UP_ICON;
+	isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
+	color = isMouseInIcon ? iconHighlightColor : iconColor;
+	r.drawSubImage(mUiIcons, currentIconBounds.x(), currentIconBounds.y(), 96, 128, 32, 32, color.red(), color.green(), color.blue(), color.alpha());
 
-	// East / West / North / South
-	if (isPointInRect(MOUSE_COORDS, MOVE_EAST_ICON))
-	{
-		r.drawSubImage(mUiIcons, MOVE_EAST_ICON.x(), MOVE_EAST_ICON.y(), 32, 128, 32, 16, 255, 0, 0, 255);
-	}
-	else
-	{
-		r.drawSubImage(mUiIcons, MOVE_EAST_ICON.x(), MOVE_EAST_ICON.y(), 32, 128, 32, 16);
-	}
+	currentIconBounds = MOVE_EAST_ICON;
+	isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
+	color = isMouseInIcon ? iconHighlightColor : iconColor;
+	r.drawSubImage(mUiIcons, currentIconBounds.x(), currentIconBounds.y(), 32, 128, 32, 16, color.red(), color.green(), color.blue(), color.alpha());
 
-	if (isPointInRect(MOUSE_COORDS, MOVE_WEST_ICON))
-	{
-		r.drawSubImage(mUiIcons, MOVE_WEST_ICON.x(), MOVE_WEST_ICON.y(), 32, 144, 32, 16, 255, 0, 0, 255);
-	}
-	else
-	{
-		r.drawSubImage(mUiIcons, MOVE_WEST_ICON.x(), MOVE_WEST_ICON.y(), 32, 144, 32, 16);
-	}
+	currentIconBounds = MOVE_WEST_ICON;
+	isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
+	color = isMouseInIcon ? iconHighlightColor : iconColor;
+	r.drawSubImage(mUiIcons, currentIconBounds.x(), currentIconBounds.y(), 32, 144, 32, 16, color.red(), color.green(), color.blue(), color.alpha());
 
-	if (isPointInRect(MOUSE_COORDS, MOVE_NORTH_ICON))
-	{
-		r.drawSubImage(mUiIcons, MOVE_NORTH_ICON.x(), MOVE_NORTH_ICON.y(), 0, 128, 32, 16, 255, 0, 0, 255);
-	}
-	else
-	{
-		r.drawSubImage(mUiIcons, MOVE_NORTH_ICON.x(), MOVE_NORTH_ICON.y(), 0, 128, 32, 16);
-	}
+	currentIconBounds = MOVE_NORTH_ICON;
+	isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
+	color = isMouseInIcon ? iconHighlightColor : iconColor;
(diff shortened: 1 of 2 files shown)


src/States/MapViewStateDraw.cpp
```diff
@@ -251,17 +247,15 @@ void MapViewState::drawNavInfo()
 	if (isPointInRect(MOUSE_COORDS, MOVE_DOWN_ICON))
 	{
 		r.drawSubImage(mUiIcons, MOVE_DOWN_ICON.x(), MOVE_DOWN_ICON.y(), 64, 128, 32, 32, 255, 0, 0, 255);
-	}
-	else
+	} else
```
DanRStevens: This new code formatting style is not used anywhere else in the project. I would recommend leaving this change out for consistency. A quick search shows that everywhere else in the code base, the preferred syntax is:
```cpp
}
else
{
```

----

I actually do prefer a more compact brace style, however this really doesn't seem like the right PR to be making this change.

cugone:
Oh my god. The master merge screwed everything up.

cugone:
I hate it that it just merges files that don't conflict all willy-nilly and it's definition of "don't conflict" is specific lines changed in both files. If only one file changes them it deems them safe to merge even if it overwrites a newer file.

ldicker83:
So I haven't looked over all the changes but I noticed a number of places where a division by two in floating point that was originally static_cast'd to int were reverted back to floating point. The cast to int was intentional to avoid 'fuzzy' bilinear filtering in non-whole positions for text/lines. By truncating the fractional part of these divisions we force pixel alignment and avoid triggering filtering. I haven't gotten too deep into the changes but it's going to take some time to test all of this and see how it goes.

Dan also mentioned that there is some formatting changes that don't jive with the rest of the styling of the code.

```
if {
} else {
}
```

Style coding boils my blood. It's a stylistic choice, I know, and I know that a lot of industry standards use the above styling but I do not and I'm not willing to change that style. Just FYI for the future, this is a change I will reject every time.

---

TL;DR:

I've been struggling with a lot of personal stuff lately so I haven't had the time or energy to look at the massive number of proposed changes. It will take some time for me to get through it all, so I ask that you be patient and know that I appreciate the efforts.

cugone:
> 
> So I haven't looked over all the changes but I noticed a number of places where a division by two in floating point that was originally static_cast'd to int were reverted back to floating point. The cast to int was intentional to avoid 'fuzzy' bilinear filtering in non-whole positions for text/lines. By truncating the fractional part of these divisions we force pixel alignment and avoid triggering filtering. I haven't gotten too deep into the changes but it's going to take some time to test all of this and see how it goes.

I noticed in the renderer you've defaulted to linear MIN/MAG/MIP sampling. For a pixelated look this is undesirable. It is trivial to change this to point filtering and as a consequence it removes the bilinear fuzzing and you get pixel-perfect images.

> 
> Dan also mentioned that there is some formatting changes that don't jive with the rest of the styling of the code.
> 
> ```
> if {
> } else {
> }
> ```
> 
> Style coding boils my blood. It's a stylistic choice, I know, and I know that a lot of industry standards use the above styling but I do not and I'm not willing to change that style. Just FYI for the future, this is a change I will reject every time.
> 

This was an issue with the clang-format file, an unintentional change, future commits fixed the issue but then the merge with master re-applied these changes.

ldicker83:
Eeesh, well, at least we know the culprit.

As for the pixel alignment, I made mention in another comment that this could be handled by control but I like your idea better. Admittedly I'm not very proficient with OpenGL so would the change in filtering cause aliasing artifacts for images that aren't pixel aligned?

I'm game for making that a middleware change versus forcing it here.

As for the stylistic changes, I'm currently fighting with vcpkg and some poor cmake porting on the part of the library uploaders... the release builds of SDL_TTF and SDL_Mixer link against debug builds of SDL2... go figure.

cugone:
I undid the merge so the single file I changed was put back to what I intended.

Files I did not change are being marked as conflicted. This is starting to get frustrating.

cugone:
Dropped first commit. Things look much better.

src/States/MapViewStateDraw.cpp
```diff
@@ -92,19 +92,13 @@ void MapViewState::drawMiniMap()
...
-			r.drawSubImage(mUiIcons, _mine.x() + mMiniMapBoundingBox.x() - 2, _mine.y() + mMiniMapBoundingBox.y() - 2, 16.0f, 0.0f, 7.0f, 7.0f);
-		}
-
+		bool isMineActive = mine->active();
+		bool isMineUnused = !isMineActive;
+		bool isMineExhausted = mine->exhausted();
+		bool isMineProducing = !isMineExhausted;
+		bool isMineActiveAndProducing = isMineActive && isMineProducing;
+		float mineBeaconStatusOffsetX = isMineUnused ? 0.0f : (isMineActiveAndProducing ? 8.0f : (isMineExhausted ? 16.0f : 0.0f));
+		r.drawSubImage(mUiIcons, _mine.x() + mMiniMapBoundingBox.x() - 2, _mine.y() + mMiniMapBoundingBox.y() - 2, mineBeaconStatusOffsetX, 0.0f, 7.0f, 7.0f);
```
DanRStevens: I general, I like how some complicated expressions are pulled out into named variables. In this particular case, I felt it was maybe going a tad bit far. I didn't feel the names here really helped explain what were already pretty simple and easy to understand expressions.

As much as I love the ternary operator, I generally advise against nesting it multiple levels deep. The cascaded `if`/`else if` can start to be easier to read.

Embedded in this is a small change in behaviour, which is a positive. Previously, if none of the `if` cases held, the image was not drawn. Now it is always drawn, but in a modified way. That's probably a safer way to handle this situation, and should be friendlier to updates involving possible mine states.

----

I would suggest using fewer named variables here. You don't really need a new variable name just to hold the result of a named function, nor just to negate something. There are 3 conditions, so 3 boolean variables max should suffice.

src/States/MapViewStateDraw.cpp
```diff
@@ -254,62 +254,37 @@ void MapViewState::drawNavInfo()
...
+	bool isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
+	NAS2D::Color iconColor{255, 255, 255, 255};
+	NAS2D::Color iconHighlightColor{255, 0, 0, 255};
+	NAS2D::Color color = isMouseInIcon ? iconHighlightColor : iconColor;
+	r.drawSubImage(mUiIcons, currentIconBounds.x(), currentIconBounds.y(), 64, 128, 32, 32, color.red(), color.green(), color.blue(), color.alpha());
+
+	currentIconBounds = MOVE_UP_ICON;
+	isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
+	color = isMouseInIcon ? iconHighlightColor : iconColor;
+	r.drawSubImage(mUiIcons, currentIconBounds.x(), currentIconBounds.y(), 96, 128, 32, 32, color.red(), color.green(), color.blue(), color.alpha());
```
DanRStevens: I like what you're going for here. It seems a bit off though. It's kind of hard to read. There's a lot of duplication. Variables are being re-used in different contexts. This seems like a sign the code can be extracted to a function.

The magic numbers can probably be extracted to `Rectangle` structs. It looks like they are associated with an icon. A function could then take each icon `Point` and associated "magic number" `Rectangle`, as well as the current mouse `Point`, and draw it with any appropriate highlight. Not sure if you want the highlight colors to be parameters too.

src/States/MapViewStateDraw.cpp
cugone: I was taught that readability trumps brevity. In this case, each variable denotes a state that the mine could be in, even though logically they only negate a previous state that negated stated doesn't mean "the opposite" of the other state. Since they all ultimately only affect one thing: the X offset of the status image; I opted for a ternary operator to reduce the use of `if/else` chains. Otherwise I would have written it like this:
```
float mineBeaconStatusOffsetX = 0.0f;
if(isMineUnused) { mineBeaconStatusOffsetX = 0.0f; }
else if(isMineActiveAndProducing) { mineBeaconStatusOffsetX = 8.0f; }
else if(isMineExhausted) { mineBeaconStatusOffsetX = 16.0f; }
else { mineBeaconStatusOffsetX = 0.0f; } //always include an else in an if-else chain!
```

src/States/MapViewStateDraw.cpp
```diff
@@ -254,37 +266,12 @@ void MapViewState::drawNavInfo()
...
-	currentIconBounds = MOVE_SOUTH_ICON;
-	isMouseInIcon = isPointInRect(MOUSE_COORDS, currentIconBounds);
-	color = isMouseInIcon ? iconHighlightColor : iconColor;
-	r.drawSubImage(mUiIcons, currentIconBounds.x(), currentIconBounds.y(), 0, 144, 32, 16, color.red(), color.green(), color.blue(), color.alpha());
+	drawNavIcon(r, MOVE_DOWN_ICON, NAS2D::Rectangle_2d{64, 128, 32, 32}, NAS2D::Color::White, NAS2D::Color::Red);
+	drawNavIcon(r, MOVE_UP_ICON, NAS2D::Rectangle_2d{96, 128, 32, 32}, NAS2D::Color::White, NAS2D::Color::Red);
+	drawNavIcon(r, MOVE_EAST_ICON, NAS2D::Rectangle_2d{32, 128, 32, 16}, NAS2D::Color::White, NAS2D::Color::Red);
+	drawNavIcon(r, MOVE_WEST_ICON, NAS2D::Rectangle_2d{32, 144, 32, 16}, NAS2D::Color::White, NAS2D::Color::Red);
+	drawNavIcon(r, MOVE_NORTH_ICON, NAS2D::Rectangle_2d{0, 128, 32, 16}, NAS2D::Color::White, NAS2D::Color::Red);
+	drawNavIcon(r, MOVE_SOUTH_ICON, NAS2D::Rectangle_2d{0, 144, 32, 16}, NAS2D::Color::White, NAS2D::Color::Red);
```
DanRStevens: Nice. That really improved the code.

cugone: I don't know why I didn't do it earlier. I have a personal rule, "If you repeat code more than twice, make it a function." I was clearly repeating code *six* times.

src/States/MapViewStateDraw.cpp
DanRStevens: I think the issue I have is that the variable names don't tell us anything that code doesn't already tell us. Compare the left and right sides:
```cpp
bool isMineActive = mine->active();
bool isMineExhausted = mine->exhausted();
bool isMineActiveAndProducing = isMineActive && isMineProducing;
```

cugone: True, but the variables also reduce cognitive load. I don't have to think about order of operations, the meaning of `&&`, is the use of `&&` a mistake, etc. I just have to be able to answer "is the mine active? is it exhausted? is it active and producing?"

cugone: I've pushed a commit to address the issue, but I still think it's a bad idea. Now the function calls are being duplicated whereas before they were being called exactly once.

cugone:
Poking the thread to re-request a review of recent changes.

PR #178, opened 2020-02-02
Formatting changes to MainReportsUiState.cpp

cugone:


src/States/MapViewStateDraw.cpp
```diff
@@ -151,57 +157,47 @@ void MapViewState::drawResourceInfo()
 	updateGlowTimer();
 
 	// Common Metals
-	r.drawSubImage(mUiIcons, x, y , 64, 16, constants::RESOURCE_ICON_SIZE, constants::RESOURCE_ICON_SIZE);
-	if (mPlayerResources.commonMetals() <= 10) { r.drawText(*MAIN_FONT, string_format("%i", mPlayerResources.commonMetals()), x + margin, textY, 255, GLOW_STEP, GLOW_STEP); }
-	else { r.drawText(*MAIN_FONT, string_format("%i", mPlayerResources.commonMetals()), x + margin, textY, 255, 255, 255); }
+	r.drawSubImage(mUiIcons, x, y, 64, 16, constants::RESOURCE_ICON_SIZE, constants::RESOURCE_ICON_SIZE);
+	if (mPlayerResources.commonMetals() <= 10) { r.drawText(*MAIN_FONT, string_format("%i", mPlayerResources.commonMetals()), x + margin, textY, 255, GLOW_STEP, GLOW_STEP); } else { r.drawText(*MAIN_FONT, string_format("%i", mPlayerResources.commonMetals()), x + margin, textY, 255, 255, 255); }
```
DanRStevens: I don't like these formatting changes that put large `if` and `else` blocks on the same line.

cugone: This issue is fixed in #177

DanRStevens:
The first commit here is the same as on #177, and is hard to review. Suggestions on that other PR apply here.

src/States/MainMenuState.cpp
```diff
@@ -104,17 +104,17 @@ void MainMenuState::positionButtons()
...
-	btnQuit.position(start_x, start_y + 140);
+	btnContinueGame.position(start_x, start_y + 35.0f);
+	btnOptions.position(start_x, start_y + 70.0f);
+	btnHelp.position(start_x, start_y + 105.0f);
+	btnQuit.position(start_x, start_y + 140.0f);
 
-	mFileIoDialog.position(static_cast<int>(r.center_x() - mFileIoDialog.width() / 2), static_cast<int>(r.center_y() - mFileIoDialog.height() / 2));
-	dlgOptions.position(static_cast<int>(r.center_x() - dlgOptions.width() / 2), static_cast<int>(r.center_y() - dlgOptions.height() / 2));
+	mFileIoDialog.position(r.center_x() - mFileIoDialog.width() / 2.0f, r.center_y() - mFileIoDialog.height() / 2.0f);
+	dlgOptions.position(r.center_x() - dlgOptions.width() / 2.0f, r.center_y() - dlgOptions.height() / 2.0f);
```
ldicker83: Not truncating the floating point's to ints here causes alignment issues. Windows and other UI elements really need to be pixel aligned in order to look correct.

src/States/MapViewStateDraw.cpp
```diff
@@ -251,17 +247,15 @@ void MapViewState::drawNavInfo()
 	if (isPointInRect(MOUSE_COORDS, MOVE_DOWN_ICON))
 	{
 		r.drawSubImage(mUiIcons, MOVE_DOWN_ICON.x(), MOVE_DOWN_ICON.y(), 64, 128, 32, 32, 255, 0, 0, 255);
-	}
-	else
+	} else
```
ldicker83: This styling does not conform to project styling.

src/States/MapViewStateUi.cpp
```diff
@@ -77,26 +77,26 @@ void MapViewState::initUi()
 	mDiggerDirection.directionSelected().connect(this, &MapViewState::diggerSelectionDialog);
 	mDiggerDirection.hide();
 
-	mTileInspector.position(static_cast<int>(r.center_x() - mTileInspector.width() / 2), static_cast<int>(r.height() / 2 - 175));
+	mTileInspector.position(r.center_x() - mTileInspector.width() / 2.0f, r.height() / 2.0f - 175.0f);
 	mTileInspector.hide();
 
-	mStructureInspector.position(static_cast<int>(r.center_x() - mStructureInspector.width() / 2), static_cast<int>(r.height() / 2 - 175));
+	mStructureInspector.position(r.center_x() - mStructureInspector.width() / 2.0f, r.height() / 2.0f - 175.0f);
```
ldicker83: Same as above, the cast was done intentionally to force pixel alignment.

src/States/MapViewStateUi.cpp
```diff
@@ -208,15 +208,15 @@ void MapViewState::setupUiPositions(int w, int h)
...
-	mGameOverDialog.position(centerWindowWidth(mGameOverDialog.width()), centerWindowHeight(mGameOverDialog.height()) - 100);
-	mAnnouncement.position(centerWindowWidth(mAnnouncement.width()), centerWindowHeight(mAnnouncement.height()) - 100);
-	mGameOptionsDialog.position(centerWindowWidth(mGameOptionsDialog.width()), centerWindowHeight(mGameOptionsDialog.height()) - 100);
+	mFileIoDialog.position(centerWindowWidth(mFileIoDialog.width()), 50.0f);
+	mGameOverDialog.position(centerWindowWidth(mGameOverDialog.width()), centerWindowHeight(mGameOverDialog.height()) - 100.0f);
+	mAnnouncement.position(centerWindowWidth(mAnnouncement.width()), centerWindowHeight(mAnnouncement.height()) - 100.0f);
+	mGameOptionsDialog.position(centerWindowWidth(mGameOptionsDialog.width()), centerWindowHeight(mGameOptionsDialog.height()) - 100.0f);
 
-	mDiggerDirection.position(centerWindowWidth(mDiggerDirection.width()), static_cast<int>(h / 2) - 125);
+	mDiggerDirection.position(centerWindowWidth(mDiggerDirection.width()), static_cast<int>(h / 2.0f) - 125.0f);
```
ldicker83: And again.

To be clear, the casting is done whenever a division is used to determine placement of a UI element and to force pixel alignment. I suppose this could be moved up to Control to not make it the users problem but that would be a different change set.

I'm open to discussion on this -- maybe it would make sense to open an issue and have the pixel alignment handled directly in Control's position assignment operations.

DanRStevens: Maybe we should be using [`std::floor`](https://en.cppreference.com/w/cpp/numeric/math/floor) if what we really want is rounding. I assume in this case the type is unchanged, so both pre-rounded and post-rounded values are `float`.

More long term, I think we should be using integer coordinates for UI component positions, particularly if the values basically represent pixels anyway.

cugone: In my experience, using percentages and unit offsets for UI alignment performs better on a wider variety of screen resolutions. So say you want a control to appear in the center of its parent, then offset by a few units down. The call would be: `setPosition(0.5f, 0.5f, 0.0f, 2.0f)`. As a bonus, this is resolution independent. It will look the same for any screen.

This is most definitely long term though as it involves setting up a matrix transform hierarchy and we don't have support for matrices yet.

cugone:
Voting to close. This needs less work than actually done due to comments by @ldicker83. It's easier to start over.

PR #184, opened 2020-02-03
Removes duplication from the MineVein declarations by using auto

### LINKED ISSUE #176: Shorten MineVein declarations
Ref: https://github.com/ldicker83/OPHD/pull/170#discussion_r373844454
> 
> I believe this could be shortened further to:
> 
> ```c++
> MineVein _mv{0, 0, 0, 0};
> ```

cugone:
Closes #176

### DIFF
--- a/src/Mine.cpp
+++ b/src/Mine.cpp
@@ -149,7 +149,7 @@ void Mine::miningRareMinerals(bool _b)
  */
 void Mine::increaseDepth()
 {
-	MineVein vein;
+	auto vein = MineVein{0,0,0,0};
 
 	switch (productionRate())
 	{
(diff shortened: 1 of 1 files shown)


src/Mine.cpp
```diff
@@ -346,7 +346,7 @@ void Mine::deserialize(NAS2D::Xml::XmlElement* _ti)
 	mVeins.resize(depth);
 	for (XmlNode* vein = _ti->firstChild(); vein != nullptr; vein = vein->nextSibling())
 	{
-		MineVein _mv = MineVein{0, 0, 0, 0};
+		auto _mv = MineVein{0, 0, 0, 0};
```
DanRStevens: Was thinking that since this is a declaration, it could use initialization:
```cpp
MineVein vein{0, 0, 0, 0};
```

That might be a better general solution than default initialization, followed by assignment to overwrite the default values. Not that it matters since `MineVein` is just a `typedef` for `std::array<int, 4>`, so default initialization should just do nothing and leave it uninitialized, which is then overwritten by the assignment.

Same thought for both updated lines.

cugone: The language specifies that declaration of the form `Foo a = Foo()` or `Foo a = Foo{}` is not an assignment. It is initialization. The copy constructor is never called.

See: https://en.cppreference.com/w/cpp/language/initialization

DanRStevens: Right. So style choice then.

cugone: Yes. "Almost always auto". :)

PR #186, opened 2020-02-04
Adds UI RadioButton

cugone:
Adds a radio button to the UI system.

Radio buttons are like check boxes except only one in a UIContainer may be checked at a time.

If not added specifically to a Panel (a child UIContainer of a window), the window becomes the UIContainer parent of the radio button.

### DIFF
--- a/src/UI/Core/RadioButton.cpp
+++ b/src/UI/Core/RadioButton.cpp
@@ -75,9 +75,9 @@ const std::string& RadioButton::text() const
  */
 RadioButton::ClickCallback& RadioButton::click()
 {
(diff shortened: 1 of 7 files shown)


src/UI/Core/RadioButton.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+	return mLabel.text();
+}
+
+/**
+ *
+ */
+RadioButton::ClickCallback& RadioButton::click()
+{
+	for(auto* sibling : mParentContainer->controls())
```
DanRStevens: Spacing: `for (`

PR #187, opened 2020-02-04
Adds difficulty modifier with main menu dialog

### LINKED ISSUE #25: Add Difficulty Selection and Global Modifier
Difficulty selection will affect the following values:
- Resource availability and yield at Mines
- Morale effects (both negative and positive)
- Number of colonists available at the beginning
- Birth and Death rates
- Planet variety

cugone:
Closes #25

Requires #186

 - Adds a global difficulty modifier to `Common.h`
 - Adds a dialog on New Game to select it.

Currently the difficulty is not saved between games because it has not been incorporated into any features where it would make a difference.

### DIFF
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -16,6 +16,42 @@
 
 using namespace NAS2D;
 
+std::string StringFromDifficulty(const Difficulty& difficulty)
+{
+	switch(difficulty)
+	{
+    case Difficulty::Beginner:	return constants::DIFFICULTY_BEGINNER;
+    case Difficulty::Easy:		return constants::DIFFICULTY_EASY;
+    case Difficulty::Medium:	return constants::DIFFICULTY_MEDIUM;
+    case Difficulty::Hard:		return constants::DIFFICULTY_HARD;
+	default:					return constants::DIFFICULTY_EASY;
+	}
+}
+Difficulty DifficultyFromString(std::string difficultyStr)
+{
+	difficultyStr = NAS2D::toLowercase(difficultyStr);
+	if(difficultyStr == "beginner")
+	{
+		return Difficulty::Beginner;
+	}
+    else if(difficultyStr == "easy")
+    {
+		return Difficulty::Easy;
+    }
+    else if(difficultyStr == "medium")
+    {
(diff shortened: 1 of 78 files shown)


src/UI/DifficultySelect.cpp
```diff
@@ -107,9 +107,9 @@ void DifficultySelect::init()
 }
 
 void DifficultySelect::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandler::KeyModifier mod, bool repeat) {
-    if(key == NAS2D::EventHandler::KeyCode::KEY_ESCAPE) {
+	if(key == NAS2D::EventHandler::KeyCode::KEY_ESCAPE) {
```
DanRStevens: Spacing: `if (`

src/Common.cpp
```diff
@@ -19,13 +19,13 @@ using namespace NAS2D;
...
-	{
+    {
     case Difficulty::Beginner:  return constants::DIFFICULTY_BEGINNER;
     case Difficulty::Easy:      return constants::DIFFICULTY_EASY;
     case Difficulty::Medium:    return constants::DIFFICULTY_MEDIUM;
     case Difficulty::Hard:      return constants::DIFFICULTY_HARD;
-	default:                    return constants::DIFFICULTY_EASY;
-	}
+    default:                    return constants::DIFFICULTY_EASY;
+    }
```
DanRStevens: Indentation still doesn't look right. The `switch` is using a tab, while the block contents are using spaces.

cugone: Are the tabs for github set to 5?

DanRStevens: Point is, the whole file is using tabs, except for the indented block of `case` statements.

DanRStevens: Though it would be ingenious if GitHub used a non-standard value for tabs, just to make the differences stand out more.

DanRStevens:
@ldicker83, you might want to take a look at this one. It changes the user interface, adding a difficulty selection menu between clicking new game and choosing a planet type. Maybe there's some way we can streamline it in the future, so it behaves less like a popup.

PR #199, opened 2020-02-09
Enable Windows CI builds

DanRStevens:
Add AppVeyor config file.


DanRStevens:
It seems the solution file maps the solution configuration name to individual project configurations. That's how the `x86` was translated to `Win32` for NAS2D.

Issue #206, opened 2020-02-10
Replace `isPointInRect` with `rect.contains(point)`


DanRStevens:
The `Rectangle` class now has a `contains` method, which can check if it contains a given `Point`. This should be used in place of the global function `isPointInRect`.

DanRStevens:
Adding reference to upstream change: https://github.com/lairworks/nas2d-core/pull/447

Many existing uses of `isPointInRect` will fail to compile if the upstream branch is merged, since there is no direct header include for `isPointInRect`.

As there is alternate functionality in `Rectangle::contains`, which is already being imported, it may be better to just change the method being used, rather than adjust headers.

PR #212, opened 2020-02-11
Convert `isPointInRect` to `contains` States

DanRStevens:
This is a bulk update for #206. It addresses only the changes that apply to States.

### DIFF
--- a/src/States/MainReportsUiState.cpp
+++ b/src/States/MainReportsUiState.cpp
@@ -126,7 +126,7 @@ static void setPanelRects(int width)
  */
 static void drawPanel(Renderer& _r, Panel& _p)
 {
-	if (isPointInRect(MOUSE_COORDS, _p.Rect)) { _r.drawBoxFilled(_p.Rect, 0, 185, 185, 100); }
+	if (_p.Rect.contains(MOUSE_COORDS)) { _r.drawBoxFilled(_p.Rect, 0, 185, 185, 100); }
 
 	if (_p.Selected())
 	{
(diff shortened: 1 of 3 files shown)


src/States/MainReportsUiState.cpp
```diff
@@ -269,13 +269,13 @@ void MainReportsUiState::onMouseDown(EventHandler::MouseButton button, int x, in
 {
 	if (!active()) { return; }
 
-	if (!isPointInRect(x, y, 0, 0, static_cast<int>(Utility<Renderer>::get().width()), 40)) { return; } // ignore clicks in the UI area.
+	if (!NAS2D::Rectangle{0, 0, static_cast<int>(Utility<Renderer>::get().width()), 40}.contains(Point{x, y})) { return; } // ignore clicks in the UI area.
```
cugone: This would be much easier to read if the `Rectangle` were its own variable.

Consider
```
//...
{ //constrained scope
auto screenWidth = static_cast<int>(Utility<Renderer>::get().width());
auto menuBar = NAS2D::Rectangle{0, 0, screenWidth, 40};
auto locationClicked = NAS2D::Point{x, y};
if (!menuBar.contains(locationClicked)) { return; }
}
//...
```

DanRStevens: I agree. I would like to get there. I don't know if this is the right PR to start doing too many changes like that.

The focus on this PR is swapping out an old method for a new method. It's a fairly mechanical change, and doesn't require deep understanding of the code. In particular, I want the change in so the code doesn't break when upstream change https://github.com/lairworks/nas2d-core/pull/447 is merged.

In this particular case, the original code doesn't name or describe the original rectangle, and no information is being lost by the change. My focus is on incremental improvement, which didn't include this particular dimension, where rectangles are unnamed.

More long term, I would like the `Point` and `Rectangles` to be named. In particular, I would like to change all the methods with `int x, int y` parameters to either `Point<int>` or `Vector<int>`, depending on use. That would greatly reduce need to construct objects within the methods themselves.

So my thought is yes, but defer those changes to a new PR.

cugone: Good point. Make a new issue and I'll approve this.

PR #214, opened 2020-02-11
Convert `isPointInRect` to `contains` in TileMap

DanRStevens:
This is an update for #206. It addresses only the changes that apply to `TileMap`.

### DIFF
--- a/src/Map/TileMap.cpp
+++ b/src/Map/TileMap.cpp
@@ -322,7 +322,7 @@ void TileMap::centerMapOnTile(Tile* _t)
  */
 bool TileMap::tileHighlightVisible() const
 {
-	return isPointInRect(mMapHighlight.x(), mMapHighlight.y(), 0, 0, mEdgeLength - 1, mEdgeLength - 1);
+	return NAS2D::Rectangle{0, 0, mEdgeLength - 1, mEdgeLength - 1}.contains(mMapHighlight);
 }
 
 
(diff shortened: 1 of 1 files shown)


src/Map/TileMap.cpp
```diff
@@ -607,7 +607,7 @@ Tile* TileMap::getVisibleTile(int x, int y, int level)
 
 bool TileMap::isVisibleTile(int _x, int _y, int _d) const
 {
-	if (!isPointInRect(_x, _y, mMapViewLocation.x(), mMapViewLocation.y(), mEdgeLength - 1, mEdgeLength - 1))
+	if (!NAS2D::Rectangle{mMapViewLocation.x(), mMapViewLocation.y(), mEdgeLength - 1, mEdgeLength - 1}.contains(NAS2D::Point{_x, _y}))
```
cugone: Consider making this a variable for readability. I don't know what the rectangle represents.

DanRStevens: Neither do I, which is one reason why I left it. The original code doesn't give any indication as to what those coordinates represent.

cugone: Looking at it more closely, `mMapViewLocation` is probably a topleft camera position and mEdgeLength is probably extents of said view.

cugone: That said, since it's culling anything outside that area: `cullBounds` is appropriate

cugone: In world space of course (i.e., tile coordinates)

PR #220, opened 2020-02-12
Remove pointInRect_f functions

DanRStevens:
Remove `pointInRect_f` functions. Instead consider using the `Rectangle::contains` method. The `Rectangle` methods are more explicit about type conversions, making the code easy to reason about, and generally quite compact when working with already packed values of the correct type.

Edit: In the spirit up https://github.com/lairworks/nas2d-core/pull/457, I thought I'd continue that line of thought in OPHD.

### DIFF
--- a/src/UI/Core/Window.cpp
+++ b/src/UI/Core/Window.cpp
@@ -57,8 +57,8 @@ void Window::onMouseDown(EventHandler::MouseButton button, int x, int y)
 
 	UIContainer::onMouseDown(button, x, y);
 
-	const auto titleBarRect = NAS2D::Rectangle{mRect.x(), mRect.y(), mRect.width(), sWindowTitleBarHeight};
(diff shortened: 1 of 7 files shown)


src/UI/Core/Window.cpp
```diff
@@ -57,7 +57,8 @@ void Window::onMouseDown(EventHandler::MouseButton button, int x, int y)
 
 	UIContainer::onMouseDown(button, x, y);
 
-	mMouseDrag = (button == EventHandler::MouseButton::BUTTON_LEFT && pointInRect_f(x, y, rect().x(), rect().y(), rect().width(), sWindowTitleBarHeight));
+	const auto titleBarRect = NAS2D::Rectangle{mRect.x(), mRect.y(), mRect.width(), sWindowTitleBarHeight};
```
cugone: `titleBarBounds` is a better name.

PR #221, opened 2020-02-13
Use packed `Sprite::color` in Structure

DanRStevens:
Removes references to the unpacked `Sprite::color` overload. This allows the unpacked version to potentially be removed.

### DIFF
--- a/src/Things/Structures/Structure.cpp
+++ b/src/Things/Structures/Structure.cpp
@@ -85,7 +85,7 @@ Structure::~Structure()
 void Structure::disable(DisabledReason reason)
 {
 	sprite().pause();
-	sprite().color(255, 0, 0, 185);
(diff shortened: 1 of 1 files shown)


src/Things/Structures/Structure.cpp
```diff
@@ -105,7 +105,7 @@ void Structure::enable()
 	}
 
 	sprite().resume();
-	sprite().color(255, 255, 255, 255);
+	sprite().color(NAS2D::Color{255, 255, 255, 255});
```
cugone: Can also use `NAS2D::Color::White` here.

DanRStevens: Good point. It would be nice to switch to named constants. Will probably need to do another pass over the code to handle cases like that.

PR #222, opened 2020-02-14
Simplify signal declarations

DanRStevens:
Simplifies `Signal` declarations by removing explicit `void` for no parameters, and removing the explicit count of the number of parameters. Parameter count is automatically detected by the updated template `Signal` code.


DanRStevens:
Adding reference to upstream change:
https://github.com/lairworks/nas2d-core/pull/470

PR #225, opened 2020-02-15
Fix Pather warnings

### LINKED ISSUE #128: GCC Warning comparison between signed and unsigned
GCC warnings for signed/unsigned comparisons:
```
src/MicroPather/micropather.cpp: In member function 'void micropather::PathCache::AddItem(const micropather::PathCache::Item&)':
src/MicroPather/micropather.cpp:804:13: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (index == allocated)

       ~~~~~~^~~~~~~~~~~~

src/MicroPather/micropather.cpp: In member function 'const micropather::PathCache::Item* micropather::PathCache::Find(void*, void*)':
src/MicroPather/micropather.cpp:826:13: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (index == allocated)

       ~~~~~~^~~~~~~~~~~~

src/UI/Core/ListBox.cpp: In member function 'virtual void ListBox::update()':
src/UI/Core/ListBox.cpp:302:9: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (i == mCurrentHighlight)
       ~~^~~~~~~~~~~~~~~~~~~~

src/UI/IconGrid.cpp: In member function 'void IconGrid::raiseChangedEvent()':
src/UI/IconGrid.cpp:380:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  if (mCurrentSelection != constants::NO_SELECTION)
      ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~

src/UI/IconGrid.cpp: In member function 'virtual void IconGrid::update()':
src/UI/IconGrid.cpp:428:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  if (mCurrentSelection != constants::NO_SELECTION)
      ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~

src/UI/IconGrid.cpp:438:22: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  if (mHighlightIndex != constants::NO_SELECTION)
      ~~~~~~~~~~~~~~~~^~~~~~~~~~~~

```

DanRStevens:
Closes #128

### DIFF
--- a/src/MicroPather/micropather.cpp
+++ b/src/MicroPather/micropather.cpp
@@ -687,8 +687,7 @@ void PathNodePool::AllStates(unsigned frame, std::vector<void*>* stateVec)
 
 PathCache::PathCache(int _allocated)
 {
-	mem = new Item[_allocated];
-	memset(mem, 0, sizeof(*mem) * _allocated);
+	mem = new Item[_allocated]{};
 	allocated = _allocated;
 	nItems = 0;
 	hit = 0;
--- a/src/MicroPather/micropather.cpp
+++ b/src/MicroPather/micropather.cpp
@@ -705,7 +705,7 @@ void PathCache::Reset()
 {
 	if (nItems)
 	{
-		memset(mem, 0, sizeof(*mem) * allocated);
+		mem = {};
 		nItems = 0;
 		hit = 0;
 		miss = 0;
--- a/src/MicroPather/micropather.cpp
+++ b/src/MicroPather/micropather.cpp
@@ -831,7 +831,7 @@ const PathCache::Item* PathCache::Find(void* start, void* end)
 
 void MicroPather::GetCacheData(CacheData* data)
 {
-	memset(data, 0, sizeof(*data));
+	*data = {};
 
 	if (pathCache)
 	{
--- a/src/MicroPather/micropather.h
+++ b/src/MicroPather/micropather.h
@@ -349,8 +349,8 @@ namespace micropather
 		const Item* Find(void* start, void* end);
 
 		Item* mem = nullptr;
-		int allocated = 0;
-		int nItems = 0;
+		unsigned allocated = 0;
+		unsigned nItems = 0;
 	};
 
 	struct CacheData


src/MicroPather/micropather.cpp
```diff
@@ -706,7 +705,7 @@ void PathCache::Reset()
 {
 	if (nItems)
 	{
-		memset(mem, 0, sizeof(*mem) * allocated);
+		mem = {};
```
cugone: This doesn't do what you think it does and does not do the equivalent to what it did before.

`mem` is a dynamic C-array masquerading as a pointer, assigning it via braced initialization assigns it `nullptr`. The previous version zeroed out the memory (`Item` is a POD, so this is safe). This does not.

Conceptually, since `mem` is a dynamic C-array, that's just a `std::vector`

Change `Item* mem;` to `std::vector<Item> mem;`.

The only lines that will need to change by changing the type is 708 and 822 and the c'tor/d'tor:
708 from:
```
mem = {};
```
to
``
//std::vector::clear() does not reduce the capacity, no de-allocation is performed.
mem.clear();
```
and 822 from:
```
return mem + index;
```
to
```
return mem.data() + index;
```
and the constructor/destructor of `PathCache` to:
```
PathCache::PathCache(int _allocated)
{
        mem = std::vector<Item>(_allocated, Item{});
        allocated = _allocated;
        nItems = 0;
        hit = 0;
        miss = 0;
}
```
and the destructor can be `=default`ed or implemented to specifically clear and shrink `mem`:
``
PathCache::~PathCache()
{
        mem.clear();
        mem.shrink_to_fit();
}
```

src/MicroPather/micropather.h
```diff
@@ -349,8 +349,8 @@ namespace micropather
 		const Item* Find(void* start, void* end);
 
 		Item* mem = nullptr;
-		int allocated = 0;
-		int nItems = 0;
+		unsigned allocated = 0;
```
cugone: Personally, I hate this style. Yes, I know it defaults to `unsigned int`, but ["Specifics, Bob"](https://www.youtube.com/watch?v=qHDSNs9wBpU)

cugone:
Level 4 Warnings reveal two instances of `signed/unsigned mismatch`:

804 and 826: `if (index == allocated)` should be `if (index == static_cast<unsigned int>(allocated))`

src/MicroPather/micropather.cpp
DanRStevens: Careless oversight from programming while overly tired. The code should have read:
```cpp
*mem = {};
```

DanRStevens: Oh, the and type is not a POD by virtue of not being an aggregate type. Hence the warning about it.

src/MicroPather/micropather.h
DanRStevens: I hate that style too. I chose it to be consistent with the rest of the code in that file.

cugone: Fair enough.

src/MicroPather/micropather.cpp
cugone: Just poking the thread. Any update to this?

DanRStevens:
I'm going to close this for now. I created an issue to track the problem, with a reference to this PR, so we can easily find the suggested fix. We can open a new PR when the fix is made.

PR #226, opened 2020-02-15
Set OPHD VS project to Warning Level 4

cugone:
Sets the warning level to 4.

VS-only change.


cugone:
Whups. Forgot to set All Configurations.

PR #227, opened 2020-02-15
Fixes TextArea warnings

cugone:
Fixes int to uint8_t conversion, possible loss of data

### DIFF
--- a/src/UI/Core/TextArea.h
+++ b/src/UI/Core/TextArea.h
@@ -14,7 +14,7 @@ public:
 	TextArea() = default;
 	virtual ~TextArea() = default;
 
-	void textColor(int r, int g, int b, int a = 255) { mTextColor(r, g, b, a); }
+	void textColor(const NAS2D::Color& color) { mTextColor = color; }
 
 	void font(const std::string&, size_t);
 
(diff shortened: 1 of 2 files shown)


src/UI/Core/TextArea.h
```diff
@@ -14,6 +14,7 @@ class TextArea : public Control
 	TextArea() = default;
 	virtual ~TextArea() = default;
 
+	void textColor(int r, int g, int b, int a = 255) { textColor(NAS2D::Color{static_cast<uint8_t>(r), static_cast<uint8_t>(g), static_cast<uint8_t>(b), static_cast<uint8_t>(a)}); };
```
DanRStevens: Is there a way to remove the use of `static_cast`?

cugone: No. char/unsigned char/signed char all require down casts.

I wanted to remove it entirely, but that broke a call in `FactoryProduction.cpp`. Decided to leave it as a separate issue.

DanRStevens: Yes, I was thinking update the call. Is it just one place?

cugone: As far as the compiler warned me it is.

PR #236, opened 2020-02-19
Add more CircleCI builds

DanRStevens:
Add more CircleCI builds:
- gcc
- gcc-8
- clang
- mingw-w64 (partial support, compile without linking)

Previously, there was only a `gcc` build.


DanRStevens:
Linux only change.

Issue #246, opened 2020-02-23
Add lint rules to makefile


DanRStevens:
Would be handy to add lint rules to the `makefile`, so things like `cppclean` and `cppcheck` can be run easily. It would also provide some documentation that these tools can be used, and how to run them.

It would also be possible to setup a CI build that would run the lint tools on the code.

cugone:
Some minimal research says that clang-tidy is [built-in to VS 2019 16.4 or later](https://devblogs.microsoft.com/cppblog/code-analysis-with-clang-tidy-in-visual-studio/).

DanRStevens:
I had kind of thought we'd already enabled `Code Analysis` for Visual Studio.

----

Interesting point from the link you gave:
> Code Analysis defaults to the respective tool as dictated by your platform toolset and compiler: Microsoft Code Analysis if using MSVC (“Visual Studio 2019”) and clang-tidy if using LLVM/clang-cl.

It seems it changes the code analysis tools based on which compiler toolset you're using.

PR #247, opened 2020-02-23
Adds lint rules to MSVC

cugone:
MSVC version of #246

Adds Code Analysis rules (lint rules) for MSVC by adding checks for all the Core Guidelines. The custom `.ruleset` file is required to select multiple rule sets to follow.

There are also `Microsoft Native Minimum Rules` and `Microsoft Native Recommended Rules`, the `Minimum` is recommended for all custom rulesets and `Recommended` is for use with Visual Studio Professional Edition. They are Microsoft-specific so I decided to defer using them.


cugone:
@DanRStevens 

The default CA settings and rules are `Microsoft Minimum Rules` for Debug builds and `Disabled` for non-Debug builds. I've replaced them with All Core Guidelines for all builds.

Should I turn on `Minimum` for all builds as well as the Core Guidelines?

DanRStevens:
Woah, the builds just took 13-14 minutes per configuration, and produced over 7000 lines of output. That's overwhelming. AppVeyor only shows the most recent 2000 lines, so we're missing most of the output.

I would like to use increased linting support, but I think we're going to have to roll this one out in stages. It's not practical to merge as is.

Maybe we can experiment with enabling just one or two sets of warning in a branch, and focus on fixing those.

----

I'm uncertain about replacing `Microsoft Native Minimum Rules`. These might be a different set than the core guidelines. Maybe we'd like to enable both sets of warnings? (Assuming they don't conflict with each other, each requiring something the other will warn about).

proj/vs2019/ophd.vcxproj
```diff
@@ -89,7 +93,7 @@
       <PrecompiledHeader />
       <WarningLevel>Level4</WarningLevel>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
-      <EnablePREfast>false</EnablePREfast>
+      <EnablePREfast>true</EnablePREfast>
```
DanRStevens: For reference:
[VCCLCompilerTool.EnablePREfast Property](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.vcprojectengine.vcclcompilertool.enableprefast?view=visualstudiosdk-2017)

> true to enable code analysis (/analyze).

cugone:
The warnings for the Core Guidelines are just too numerous to fix in one go. For now I've just enabled `Microsoft Native Recommended`. It gives good coverage of problem areas and doesn't add much overhead in the way of warning count.

cugone:
Recent warnings fixes and potential additions to the list of warnings to-be-added make this current PR obsolete. Closing to remake.

PR #249, opened 2020-02-25
Rename `onMouseMotion` to `onMouseMove`

DanRStevens:
This makes the event handler name consistent across the code base. Previously it was a toss up between either `onMouseMove` or `onMouseMotion`.


DanRStevens:
Indeed. That should be updated as well.

One caveat is the `EventHandler` event name comes from NAS2D. It would need to be updated in a PR on that repo first.

PR #256, opened 2020-03-01
Add Vector direction constants

DanRStevens:
Add `Vector<int>` direction constants, and iteration offset array constants.

----

The named direction constants allows a `Point` to be easily offset in any named direction:
```cpp
const auto point = NAS2D::Point<int>{x, y};
const auto offsetPoint = point + DirectionNorth;
```

----

The iteration arrays allows easy iteration over surrounding points.

This can be done as a 3x3 scan, top to bottom, left to right (NW, N, NE, W, Center, E, SW, S, SE):
```cpp
const auto point = NAS2D::Point<int>{x, y};
for (const auto offset : DirectionScan3x3) {
    doSomeProcessing(point + offset);
}
```

This can be done as an 8 direction clockwise scan (N, NE, E, SE, S, SW, W, NW):
```cpp
const auto point = NAS2D::Point<int>{x, y};
for (const auto offset : DirectionClockwise8) {
    doSomeProcessing(point + offset);
}
```

This can be done as a 4 direction clockwise scan (N, E, S, W):
```cpp
const auto point = NAS2D::Point<int>{x, y};
for (const auto offset : DirectionClockwise4) {
    doSomeProcessing(point + offset);
}
```

### DIFF
--- a/src/DirectionOffset.h
+++ b/src/DirectionOffset.h
@@ -1,0 +1,47 @@
+#pragma once
+
+#include "NAS2D/Renderer/Vector.h"
+
+#include <array>
+
+
+constexpr auto DirectionNorthWest = NAS2D::Vector<int>{-1, -1};
+constexpr auto DirectionNorth     = NAS2D::Vector<int>{ 0, -1};
+constexpr auto DirectionNorthEast = NAS2D::Vector<int>{ 1, -1};
+constexpr auto DirectionWest      = NAS2D::Vector<int>{-1,  0};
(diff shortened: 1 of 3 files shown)


DanRStevens:
I disagree. These constants are dependent on the coordinate system, and so should be in to program rather than the library. In particular, OPHD uses screen coordinates, rather than Cartesian coordinates, and so this affects the direction of North and South along the y-axis.

These are not aliases for `Vector` per say, but rather particular instances of `Vector`.

I had given some thought towards wrapping them in some kind of a namespace. I was tempted to use an `enum class`, but unfortunately those are restricted to integral base types.

PR #258, opened 2020-03-07
Rename short parameter names UI Core

DanRStevens:
More explicit longer parameter names are easier to understand than highly abbreviated ones.

Additionally, single letter variable names do not allow highlighting of matching names in the Atom editor.

### DIFF
--- a/src/UI/Core/UIContainer.cpp
+++ b/src/UI/Core/UIContainer.cpp
@@ -35,27 +35,27 @@ UIContainer::~UIContainer()
 /**
  * Adds a Control to the UIContainer.
  */
-void UIContainer::add(Control* c, float x, float y)
+void UIContainer::add(Control* control, float x, float y)
 {
-	if(c == nullptr)
+	if(control == nullptr)
 	{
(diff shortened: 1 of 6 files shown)


src/UI/Core/UIContainer.cpp
```diff
@@ -35,27 +35,27 @@ UIContainer::~UIContainer()
 /**
  * Adds a Control to the UIContainer.
  */
-void UIContainer::add(Control* c, float x, float y)
+void UIContainer::add(Control* control, float x, float y)
```
cugone: Another PR changing the signature to `add(Control* control, Point<float> position)` would be helpful

DanRStevens: Agreed.

I would also like to change the first parameter to `Control&`, since it is non-nullable. That affects a lot of calling code though.

cugone: Actually, now that I think about it, a `Vector<float>` might be a good thing to add as well:

```
add(Control* control, Point<float> absolutePosition);
add(Control* control, Vector<float> normalizedPosition, Vector<float> offset);
```

PR #259, opened 2020-03-07
Refactor window stack

DanRStevens:
A bit of light refactoring for `WindowStack`.

### DIFF
--- a/src/UI/Core/WindowStack.cpp
+++ b/src/UI/Core/WindowStack.cpp
@@ -45,7 +45,7 @@ void WindowStack::removeWindow(Window* window)
  */
 bool WindowStack::pointInWindow(int x, int y) const
 {
-	for (auto window : mWindowList)
+	for (auto* window : mWindowList)
 	{
 		if (window->visible() && window->rect().to<int>().contains(NAS2D::Point{x, y}))
 		{
@@ -62,7 +62,7 @@ bool WindowStack::pointInWindow(int x, int y) const
  */
 void WindowStack::updateStack(int x, int y)
 {
-	for (auto window : mWindowList)
+	for (auto* window : mWindowList)
 	{
 		if (window->visible() && window->rect().to<int>().contains(NAS2D::Point{x, y}))
 		{
--- a/src/UI/Core/WindowStack.cpp
+++ b/src/UI/Core/WindowStack.cpp
@@ -84,7 +84,7 @@ void WindowStack::bringToFront(Window* window)
 		return;
 	}
 
-	(*mWindowList.begin())->hasFocus(false);
+	mWindowList.front()->hasFocus(false);
 
 	mWindowList.remove(window);
 	mWindowList.push_front(window);
--- a/src/UI/Core/WindowStack.cpp
+++ b/src/UI/Core/WindowStack.cpp
@@ -45,9 +45,8 @@ void WindowStack::removeWindow(Window* window)
  */
 bool WindowStack::pointInWindow(int x, int y) const
 {
-	for (auto it = mWindowList.begin(); it != mWindowList.end(); ++it)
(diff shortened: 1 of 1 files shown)


src/UI/Core/WindowStack.cpp
```diff
@@ -45,10 +45,9 @@ void WindowStack::removeWindow(Window* window)
  */
 bool WindowStack::pointInWindow(int x, int y) const
 {
-	for (auto it = mWindowList.begin(); it != mWindowList.end(); ++it)
+	for (auto window : mWindowList)
```
cugone: Explicity declare this as a pointer: `auto*`.

src/UI/Core/WindowStack.cpp
```diff
@@ -63,13 +62,11 @@ bool WindowStack::pointInWindow(int x, int y) const
  */
 void WindowStack::updateStack(int x, int y)
 {
-	for (auto it = mWindowList.begin(); it != mWindowList.end(); ++it)
+	for (auto window : mWindowList)
```
cugone: Explicitly declare this as a pointer: `auto*`

src/UI/Core/WindowStack.cpp
```diff
@@ -78,13 +75,18 @@ void WindowStack::updateStack(int x, int y)
...
+	if (windowPosition == mWindowList.begin())
+	{
+		return;
+	}
 
-	(*mWindowList.begin())->hasFocus(false);
+	mWindowList.front()->hasFocus(false);
 
 	mWindowList.remove(window);
 	mWindowList.push_front(window);
```
cugone: Consider changing `mWindowList` to a `std::vector` instead. The number of windows on the stack is likely small enough that using a `std::list` is a waste due to cache locality misses.

Additionally, if using vector, consider swapping the target and front window.

If order matters, consider bubble sort to preserve the order by swapping items above to bubble up the target to the front. There is likely to be a small enough set size that the complexity of bubble sort won't matter.

DanRStevens: Funny, I was just thinking this sort of operation could be done more efficiently with a linked list rather than a vector. I wasn't aware it already was a linked list.

At any rate, this gets into the realm of micro optimizations. I wouldn't want to change it without some data to back it up.

cugone: >...I wouldn't want to change it without some data to back it up.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slcon2-prefer-using-stl-vector-by-default-unless-you-have-a-reason-to-use-a-different-container

:) Using `std::vector` instead of `std::list` for middle-insertions is explicitly mentioned for small (up to a few KB in size) containers. I highly doubt there will be a stack of 512 windows in use at once (assuming 4-byte integers and pointers being the same size; and the definition of "few" means more than 2.)

Like you said, by using a vector you can also increase efficiency by only calling `push_back` and reading the vector front-to-back with `.back()` being the "top" of the stack.

Removal is also much faster by using the swap-and-pop idiom. Swap the item to be removed with the last item in the list and call `pop_back`. Small container sizes can easily bubble the swapped item back to the top if order does matter.

DanRStevens: This is maybe getting a bit off topic for this PR.

I'm aware of that recommendation, and generally do use `std::vector` by default. Still, we have no actual performance data to say which would be better here. Asymptotically, a linked list is better, though as you've pointed out, we may never reach a large enough size for it to matter. But that's also a reason to not really care, since we'll never reach a large enough size for performance to matter.

Swap and pop is not suitable here, since you'd want to maintain the window ordering. Using swap and pop would cause the clicked on window to exchange places with the top most window, rather than rotate them in order.

PR #264, opened 2020-03-10
Add UI constructor param for text

DanRStevens:
Add constructor parameter for initial text to several UI controls.


DanRStevens:
Agreed. It was more done because I was tired and wanted something simple to work on at the time. Plus, I wanted to become more familiar with the UI code for other future changes, such as the layout changes.

PR #272, opened 2020-03-27
Makefile warning support

DanRStevens:
Improves `makefile` warning support, and enables a bunch of warnings flags used by NAS2D.

A few warnings from NAS2D were omitted here. They are:
- `-Wextra`
- `-Wpedantic`
- `-Wzero-as-null-pointer-constant`
- `-Wold-style-cast`
- `-Wdouble-promotion`
- `-Wmissing-declarations`
- `-Wswitch-enum`
- `-Wswitch-default`
- `-Wredundant-decls`

Some work will need to be done to cleanup these warnings before we enable these flags.

----

Linux only change.

PR #274, opened 2020-03-27
Unsigned compare less than 0

DanRStevens:
Reference: #272 (`-Wextra`)

Fix comparisons of `unsigned` values against less than 0.

### DIFF
--- a/src/UI/Core/ListBoxBase.cpp
+++ b/src/UI/Core/ListBoxBase.cpp
@@ -156,11 +156,6 @@ void ListBoxBase::onMouseMove(int x, int y, int /*relX*/, int /*relY*/)
 
 	mCurrentHighlight = (y - static_cast<int>(positionY()) + mCurrentOffset) / mItemHeight;
 
-	if (mCurrentHighlight < 0)
-	{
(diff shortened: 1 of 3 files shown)


src/UI/Core/ListBoxBase.cpp
```diff
@@ -125,7 +125,7 @@ void ListBoxBase::onMouseDown(EventHandler::MouseButton button, int x, int y)
 	// A few basic checks
 	if (!rect().contains(mMousePosition) || mCurrentHighlight == constants::NO_SELECTION) { return; }
 	if (mSlider.visible() && mSlider.rect().contains(Point{x, y})) { return; }
-	if (mCurrentHighlight < 0 || static_cast<size_t>(mCurrentHighlight) >= mItems.size()) { return; }
+	if (static_cast<size_t>(mCurrentHighlight) >= mItems.size()) { return; }
```
cugone: As a matter of style, I would prefer `operator<`.

DanRStevens: You mean comparison against `size()`? I suppose I am more used to seeing `<` in that context. Though I suppose this one is inverted for an early return.

PR #275, opened 2020-03-27
Fix zero as nullptr const

DanRStevens:
Reference: #272 (`-Wzero-as-null-pointer-constant`)

Fix use of `0` as a `nullptr` constant.


DanRStevens:
Ehh, we've already fixed other stuff in that code. Plus it takes quite a while for the author to accept changes.

Issue #217, opened 2020-02-11
Use Point/Vector for `(x, y)` parameters


DanRStevens:
A lot of UI code contains unpacked parameters such as: `(int x, int y)`

We should consider instead passing `Point<int>` or `Vector<int>` objects, depending on usage. Absolute coodinates should use `Point`. Sizes and offsets should use `Vector`. Relative coordinates are effectively offsets, so may work better as `Vector` than `Point`.

One of the major sources of these unpacked parameters are mouse coordinates. Mostly these coordinates are compared against a `Rectangle`, which is done with the `contains(Point)` method, so it easier to use when coordinates are already packed. Other uses include transformations such as `mousePosition - startPoint`, which can now be done simply with `Point` operator overloads, rather than having to repeat the operation for each component.

DanRStevens:
Somewhat related here are packed `Color` components. These values have many of the same concerns as `Point` and `Vector`.

As all such values are used by rendering code, working on using packed `{x, y}` values for `Point` and `Vector` has also updated many places to use packed `Color` components as well. As a result, there are very few places left in the code that make direct access to individual components, such as `.red`, etc.

ldicker83:
This seems to be widely addressed in a lot of places, I even committed a few changes just now that make better use of `Point<int>`. This may just be an ongoing task as we move forward.

DanRStevens:
Agreed. Lots of progress, but still lots more to do.

At some point we need to address the event handling system, which affects both NAS2D and OPHD, so the work will need to be coordinated.

ldicker83:
True. I can work on the OPHD portion of it if you want to handle NAS2D, can probably start with mouse input as it's the fewest set of related events.

DanRStevens:
Hmm, come to think of it, with the way overloads work when taking addresses of functions, I think OPHD can be updated first.

Suppose there are two overloads:
```cpp
void Callback(int x, int y);
void Callback(Point<int> point);
```

These two methods form an overload set. When you take the address with `&Callback`, which one you get depends on context:
```cpp
using UnpackedCallback = void(int, int);
using PackedCallback = void(Point<int>);

UnpackedCallback cb1 = &Callback; // Gets address of Callback(int x, int y)
PackedCallback cb2 = &Callback; // Gets address of Callback(Point<int>)
```

[...]

PR #279, opened 2020-03-27
Vectorize planet

DanRStevens:
Reference: #272 (`-Wold-style-cast`)
Reference: #217

Refactor `Planet` class. Vectorizes some of the code, and removes old C-style casts.

There's probably more that can be done, but I'm stopping for the moment, and this seemed like a good stopping point.

### DIFF
--- a/src/States/Planet.cpp
+++ b/src/States/Planet.cpp
@@ -56,8 +56,6 @@ Planet::~Planet()
 
 bool Planet::pointInCircle(NAS2D::Point<int> point)
 {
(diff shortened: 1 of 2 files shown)


DanRStevens:
That may be worth doing. There are quite a few potentially interesting geometric operations we could support. Perhaps finding if various geometric shapes intersect, or finding points or areas of intersection.

----

I suppose an integer based `Vector` would be akin to a Manhattan Distance. :wink:
Well, actually, no. I guess the sum of the components of such a `Vector` would be the Manhattan Distance.
An integer based `Point` would be akin to a particular intersection.

PR #280, opened 2020-03-28
Refactor planet select

DanRStevens:
Reference: #272 (`-Wold-style-cast`)

Refactor planet select screen to remove old C-style casts.


DanRStevens:
Yeah, it would be nice to separate those two. I suppose an initial step might be to separate some of the code within functions, even if the functions were not yet broken up. I think I saw a few functions where stuff was kind of mixed together. Or maybe it's been fixed now.

PR #285, opened 2020-03-28
Refactor glow step

DanRStevens:
Reference: #283 for some base updates that lead to this update.

### DIFF
--- a/src/States/MapViewStateDraw.cpp
+++ b/src/States/MapViewStateDraw.cpp
@@ -31,10 +31,8 @@ extern std::vector<void*> path;
 
 namespace {
 	Timer glowTimer;
-	const int GlowStepSize = 20;
-
+	int glowStepDelta = 20;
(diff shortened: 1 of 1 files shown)


cugone:
I tend to name functions that are more than just a `return` and potentially costly to `Calc...` instead of `get...`

src/States/MapViewStateDraw.cpp
```diff
@@ -30,31 +30,29 @@ extern std::vector<void*> path;
...
 namespace {
-	Timer glowTimer;
-	const int GlowStepSize = 20;
-
-	int glowStep;
-	int glowStepDirection = 1;
-
-
-	void updateGlowTimer()
+	uint8_t getGlowIntensity()
```
cugone: Consider 'calcGlowIntensity` because this is doing more than just "getting" a value.

PR #287, opened 2020-03-29
Refactor MapViewStateDraw drawMiniMap

DanRStevens:


src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
 	bool isHeightmapToggled = mBtnToggleHeightmap.toggled();
 	renderer.drawImage(isHeightmapToggled ? mHeightMap : mMapDisplay, miniMapBoxFloat.startPoint());
 
-	if (ccLocationX() != 0 && ccLocationY() != 0)
+	const auto miniMapOffset = mMiniMapBoundingBox.startPoint() - NAS2D::Point<int>{0, 0};
```
cugone: `x - 0` is still `x`.

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
...
+	const auto ccPosition = ccLocation();
+	if (ccPosition != NAS2D::Point<int>{0, 0})
 	{
-		renderer.drawSubImage(mUiIcons, ccLocationX() + mMiniMapBoundingBox.x() - 15, ccLocationY() + mMiniMapBoundingBox.y() - 15, 166, 226, 30, 30);
-		renderer.drawBoxFilled(ccLocationX() + mMiniMapBoundingBox.x() - 1, ccLocationY() + mMiniMapBoundingBox.y() - 1, 3, 3, 255, 255, 255);
+		const auto ccOffsetPosition = ccPosition + miniMapOffset;
+		const auto ccCommRangeImageRect = NAS2D::Rectangle<int>{166, 226, 30, 30};
+		renderer.drawSubImage(mUiIcons, ccOffsetPosition - ccCommRangeImageRect.size() / 2, ccCommRangeImageRect);
+		renderer.drawBoxFilled(NAS2D::Rectangle<int>::Create(ccOffsetPosition - NAS2D::Vector<int>{1, 1}, NAS2D::Vector<int>{3, 3}), NAS2D::Color::White);
 	}
```
cugone: This can be a new function:
```cpp
if(ccPosition != NAS2D::Point<int>::Zero)
{
    DrawMinimapCommRangeAt(ccPosition);
}
```

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
...
 	{
-		if (_tower->operational())
+		if (commTower->operational())
 		{
-			Tile* t = Utility<StructureManager>::get().tileFromStructure(_tower);
-			renderer.drawSubImage(mUiIcons, t->x() + mMiniMapBoundingBox.x() - 10, t->y() + mMiniMapBoundingBox.y() - 10, 146, 236, 20, 20);
+			const auto commTowerPosition = structureManager.tileFromStructure(commTower)->position();
+			const auto commTowerRangeImageRect = NAS2D::Rectangle<int>{146, 236, 20, 20};
+			renderer.drawSubImage(mUiIcons, commTowerPosition + miniMapOffset - commTowerRangeImageRect.size() / 2, commTowerRangeImageRect);
 		}
```
cugone: Consider
```cpp
for (auto commTower : ...)
{
    DrawMinimapCommRange(commTower);
}
```

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
...
 	}
 
-	for (auto _mine : mTileMap->mineLocations())
+	for (auto minePosition : mTileMap->mineLocations())
```
cugone: This is an unnecessary copy of a variable that isn't changed. Use `const auto& minePosition`.

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
...
-		else if (!mine->exhausted()) { mineBeaconStatusOffsetX = 8.0f; }
-		else { mineBeaconStatusOffsetX = 16.0f; }
+		auto mineBeaconStatusOffsetX = 0;
+		if (!mine->active()) { mineBeaconStatusOffsetX = 0; }
+		else if (!mine->exhausted()) { mineBeaconStatusOffsetX = 8; }
+		else { mineBeaconStatusOffsetX = 16; }
 
-		renderer.drawSubImage(mUiIcons, _mine.x() + mMiniMapBoundingBox.x() - 2, _mine.y() + mMiniMapBoundingBox.y() - 2, mineBeaconStatusOffsetX, 0.0f, 7.0f, 7.0f);
+		const auto mineImageRect = NAS2D::Rectangle<int>{mineBeaconStatusOffsetX, 0, 7, 7};
+		renderer.drawSubImage(mUiIcons, minePosition + miniMapOffset - NAS2D::Vector<int>{2, 2}, mineImageRect);
```
cugone: More functions for reducing cognitive load:
```cpp
for (const auto& minePosition : mTileMap->mineLocations())
{
    DrawMinimapMineAt(minePosition);
}
```
And possibly the above suggestion inside:
```cpp
DrawMinimapMines() const;
```

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
...
 	}
 
-	for (auto _tile : path)
+	for (auto tile : path)
 	{
-		const auto tilePosition = static_cast<Tile*>(_tile)->position();
-		renderer.drawPoint(tilePosition.x() + mMiniMapBoundingBox.x(), tilePosition.y() + mMiniMapBoundingBox.y(), NAS2D::Color::Magenta);
+		const auto tilePosition = static_cast<Tile*>(tile)->position();
+		renderer.drawPoint(tilePosition + miniMapOffset, NAS2D::Color::Magenta);
 	}
```
cugone:
```cpp
DrawMinimapPath(path) const;
```

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
...
 
-	for (auto _robot : mRobotList)
+	for (auto robotEntry : mRobotList)
 	{
-		renderer.drawPoint(_robot.second->x() + mMiniMapBoundingBox.x(), _robot.second->y() + mMiniMapBoundingBox.y(), NAS2D::Color::Cyan);
+		const auto robotPosition = robotEntry.second->position();
+		renderer.drawPoint(robotPosition + miniMapOffset, NAS2D::Color::Cyan);
 	}
```
cugone: Consider wrapping this in a function:
```cpp
DrawMinimapRobots() const;
```

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
 	bool isHeightmapToggled = mBtnToggleHeightmap.toggled();
 	renderer.drawImage(isHeightmapToggled ? mHeightMap : mMapDisplay, miniMapBoxFloat.startPoint());
```
cugone: I just thought of this after the below changes.

---
Consider
```cpp
DrawMinimapImage() const;
```

src/States/MapViewStateDraw.cpp
```diff
@@ -69,49 +69,60 @@ void MapViewState::drawMiniMap()
...
 
 	const auto& viewLocation = mTileMap->mapViewLocation();
+	const auto edgeLength = mTileMap->edgeLength();
+	const auto viewBoxSize = NAS2D::Vector<int>{edgeLength, edgeLength};
+	const auto viewBoxPosition = viewLocation + miniMapOffset;
 
-	renderer.drawBox(mMiniMapBoundingBox.x() + viewLocation.x() + 1, mMiniMapBoundingBox.y() + viewLocation.y() + 1, mTileMap->edgeLength(), mTileMap->edgeLength(), 0, 0, 0, 180);
-	renderer.drawBox(mMiniMapBoundingBox.x() + viewLocation.x(), mMiniMapBoundingBox.y() + viewLocation.y(), mTileMap->edgeLength(), mTileMap->edgeLength(), 255, 255, 255);
+	renderer.drawBox(NAS2D::Rectangle<int>::Create(viewBoxPosition + NAS2D::Vector<int>{1, 1}, viewBoxSize), NAS2D::Color{0, 0, 0, 180});
+	renderer.drawBox(NAS2D::Rectangle<int>::Create(viewBoxPosition, viewBoxSize), NAS2D::Color::White);
```
cugone: Function all the things! \o.
```cpp
DrawMinimapBorder() const;
```

src/States/MapViewStateDraw.cpp
DanRStevens: It's a type conversion. `Point - Point = Vector`. I needed a `Vector`. That was the easiest way I could think of to convert.

src/States/MapViewStateDraw.cpp
DanRStevens: Mmm, not sure. The `Point` class only has two fields, so a copy may be preferred to indirection. Plus, it seems quite likely the compiler could optimize away the copy.

DanRStevens:
I agree there is certainly more that can be done to reduce the cognitive load. Functions may be a part of that. I was also experimenting with a more data driven design in the later PR work I pushed afterwards.

With that said, the PR was already getting kind of large, and I felt I needed to limit the scope. We can certainly look into wrapping things into functions, though that might be better tackled as a follow up PR.

cugone:
Well, you weren't specific as to what your goals were in the PR description so I was going off the title. 😉 

To me when you set out to refactor a function, that includes breaking it up into smaller functions.

src/States/MapViewStateDraw.cpp
cugone: Well, at least mark it `const`.

src/States/MapViewStateDraw.cpp
cugone: This is why the Mathematician definition of a point can be defined as "A vector with no direction nor magnitude" and why I was advocating in the beginning that the distinction should not be made as two separate classes. With the appropriate variable names just using `Vector` when referring to points and `Vector - Vector` when referring to directions removes confusions like this and is a 30-year-old convention. Don't try to buck the trend now. You're just going to confuse people who know what they're doing.

DanRStevens:
Well I suppose refactoring may or may not involve that.

My criteria for things like this is:
Does it improve the code?

I'm generally looking for incremental improvement, not so much reaching perfection all in one go.

Assuming the direction of changes is good, blocking issues are generally newly introduced problems. Everything else is a comment. Some of them strongly worded. Often they're about small opportunities that were missed. Such as missing `const`. Most likely the author would want to make those updates before merging.

src/States/MapViewStateDraw.cpp
DanRStevens: I think this may be a symptom of the design lacking clarification, or perhaps a weakness in the API being offered. Personally I like being able to get errors when I try to do something stupid like adding two `Point`s together.

In this case, there isn't much clarity on what the fixed basis is, and what is a relative offset. There's a bit of ambiguity here, in that you could go either way. Screen coordinates might be considered absolute, while map positions are relative, or map positions are absolute, while where things appear on the screen are relative. I'm inclined to go more with the later.

Keep in mind this code was originally written before there was a `Point`/`Vector` distinction.

----

In terms of improving the API, I think it would be good if the `MiniMap` could be rendered using it's own viewport, with it's own origin, and size matching the map. That would be a way to draw the minimap without having to manually offset every drawing call.

Anyway, that's an issue for another day.

cugone:
> Well I suppose refactoring may or may not involve that.

It *does* involve that. See https://sourcemaking.com/refactoring re: ["Long Method"](https://sourcemaking.com/refactoring/smells/long-method). It's also why any refactoring tool on the planet, including Visual Assist have an "Extract Method" function. For breaking up a big function into multiple smaller manageable ones.

It most certainly "improves the code" as well.

DanRStevens:
What I meant was, there are many ways to refactor code. Extracting code to a function is just one of the possibilities.

src/States/MapViewStateDraw.cpp
cugone: At the very least, an off-screen memory texture that is then drawn in the correct place.

src/States/MapViewStateDraw.cpp
DanRStevens: Came across some interesting stuff on the topic:
[Auto Type Deduction in Range-Based For Loops](https://blog.petrzemek.net/2016/08/17/auto-type-deduction-in-range-based-for-loops/)
StackOverflow: [C++ auto& vs auto](https://stackoverflow.com/questions/29859796/c-auto-vs-auto)

The note on copy elision was interesting:

> Use auto when you need a local copy. This will never produce a reference. The copy (or move) constructor must exist, but it might not get called, due to the copy elision optimization.


----

At any rate, I just noticed the use of `auto` was pre-existing, and not something I had changed here. It's perhaps a bit out of scope for this PR. Lacking a clear reason to change this, I'm going to leave it for now.

PR #288, opened 2020-03-29
Refactor MapViewStateDraw drawResourceInfo

DanRStevens:


DanRStevens:
I would like to push that off to a future update. It may be worth wrapping with functions. It may be worth using a more data driven design. I want to get some of the basic cleanup in before tackling a higher level issue like that.

PR #289, opened 2020-03-29
Refactor MapViewStateDraw drawRobotInfo

DanRStevens:


src/States/MapViewStateDraw.cpp
```diff
@@ -221,30 +221,38 @@ void MapViewState::drawResourceInfo()
 void MapViewState::drawRobotInfo()
 {
 	// CC hasn't been placed yet.
-	if (ccLocationX() == 0) { return; }
+	if (ccLocation() == NAS2D::Point<int>{0, 0}) { return; }
```
cugone: Just a thought. Why is the `ccLocation` defaulting to a valid tile? Why not default it to `{-1,-1}`?

src/States/MapViewStateDraw.cpp
```diff
@@ -221,30 +221,38 @@ void MapViewState::drawResourceInfo()
...
+	const auto diggerImageRect = NAS2D::Rectangle<int>{181, 18, 25, 25};
+	const auto robotSummaryImageRect = NAS2D::Rectangle<int>{231, 43, 25, 25};
+
+	struct IconPartsTotal {
+		const NAS2D::Rectangle<int>& imageRect;
+		int parts;
+		std::size_t total;
+	};
+
+	std::array<IconPartsTotal, 4> icons = {{
```
cugone:
```cpp
const std::array<IconPartsTotal, 4> icons
```

src/States/MapViewStateDraw.cpp
DanRStevens: Yeah, I kind of wondered that too. We should set a constant somewhere to represent that invalid position.

Probably a separate issue though. There are likely going to be a few places in the code where we need to make that change.

PR #290, opened 2020-03-29
Refactor MapViewStateDraw drawNavInfo

DanRStevens:


src/States/MapViewStateDraw.cpp
```diff
@@ -277,9 +277,9 @@ void MapViewState::drawNavInfo()
 	{
 		const auto levelString = (i == 0) ? std::string{"S"} : std::to_string(i); // Set string for current level
 		bool isCurrentDepth = i == mTileMap->currentDepth();
-		NAS2D::Color color = isCurrentDepth ? NAS2D::Color{255, 0, 0, 255} : NAS2D::Color{200, 200, 200, 255}; // red for current depth : white for others
+		NAS2D::Color color = isCurrentDepth ? NAS2D::Color::Red : NAS2D::Color{200, 200, 200}; // red for current depth : white for others
```
cugone: Note to self: Add `Color::LightGray` (and `Color::LightGrey` for the brits)

DanRStevens: By that token, we'd also need a `Colour` alias :wink:

(On a more serious note, I'm actually against adding regional spelling in source code. Pick a standard and stick with it. As many companies exporting software APIs are American, I generally choose American spelling for source code).

cugone: Point.

I like at least the grey/gray distinction because half the time I forget which one is the American and which one is the British spelling. (I had to look it up for my original comment!) Without intellisense or code completion I'd constantly get it wrong and things wouldn't compile.

DanRStevens: Admittedly I never remember which is which for that spelling either. But hey, if one fails to compile, try the other. :stuck_out_tongue:

cugone: Also, stupid question, Why does the comment say "use white" but the color is actually super light gray?

DanRStevens: I noticed that too. Probably because super light gray still looks white.

cugone: I could understand using `{254, 254, 254}` when interacting with a shader like "I want white to show up but I don't want the shader to mess with it." As Full White has a special meaning of "do nothing" to shaders. But 200? Weird. Probably should fix the inconsistency.

The Allegro Library had the same problem. It would handle transparency in the shader by skipping pixels that were exactly `{255, 0, 255}`. Using `{254, 0, 254}` alleviated the problem and to the human eye, still looked the same.

Issue #266, opened 2020-03-12
Remove uses of `string_format`


DanRStevens:
We should transition away from `string_format`. For simple cases of formatting integers with `"%i"`, we could consider using [`std::to_string`](https://en.cppreference.com/w/cpp/string/basic_string/to_string) instead.

Reference: https://github.com/lairworks/nas2d-core/issues/507

DanRStevens:
I notice there are a lot of `string_format` calls with very similar format strings, such as:
- `"%i/%i"`
- `"%i / %i"`
- `"%i of %i"`
- `"%i, %i"`

Although possible to replace each with two calls to `std::string`, it may be less code, and easier to read code, if common formatting was extracted to a function. For example:
```cpp
std::string formatPartsTotal(int parts, int total)
{
    return std::to_string(parts) + " / " + std::to_string(total);
}
```

It might be possible to modify some of the strings so they are an exact match, so a common formatting function could be used:
`"%i/%i"` => `"%i / %i"`

It might also be possible to pass a separator to the format function:
- `"/"`
- `" / "`
- `" of "`
- `", "`

Either way, perhaps such a common formatting method could go in `Common.h`/`Common.cpp`.

PR #296, opened 2020-03-30
Refactor FactoryProduction

DanRStevens:
Reference: #266

A bit of refactoring for the `FactoryProduction` class. Allows for higher level text output. Removes use of `string_format`.

### DIFF
--- a/src/UI/FactoryProduction.cpp
+++ b/src/UI/FactoryProduction.cpp
@@ -220,7 +220,7 @@ void FactoryProduction::update()
 
 	Renderer& renderer = Utility<Renderer>::get();
 
-	const auto drawTitleText = [&renderer](NAS2D::Point<int> position, std::string title, std::string text) {
+	const auto drawTitleText = [&renderer](NAS2D::Point<int> position, const std::string& title, const std::string& text) {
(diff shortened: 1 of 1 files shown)


src/UI/FactoryProduction.cpp
```diff
@@ -218,20 +218,26 @@ void FactoryProduction::update()
 
 	Window::update();
 
-	Renderer& r = Utility<Renderer>::get();
+	Renderer& renderer = Utility<Renderer>::get();
 
-	r.drawText(*FONT_BOLD, "Turns Completed:", rect().x() + constants::MARGIN * 2 + mProductGrid.width(), rect().y() + 25.0f, 255, 255, 255);
-	r.drawText(*FONT, string_format("%i of %i", mFactory->productionTurnsCompleted(), mProductCost.turnsToBuild()), rect().x() + constants::MARGIN * 2 + mProductGrid.width() + 120, rect().y() + 25.0f, 255, 255, 255);
+	const auto drawTitleText = [&renderer](NAS2D::Point<int> position, std::string title, std::string text) {
```
cugone: Because you aren't changing their values, consider changing the `std::string` parameters to `const std::string&` or `std::string_view`.

PR #297, opened 2020-03-30
Refactor SliderDraw

DanRStevens:
Reference: #266

A bit of refactoring for one block within the `draw` method of the `Slider` class. Use vectorized draw calls. Removes use of `string_format`.

### DIFF
--- a/src/UI/Core/Slider.cpp
+++ b/src/UI/Core/Slider.cpp
@@ -417,7 +417,7 @@ void Slider::draw()
 
 	if (mDisplayPosition && mMouseHoverSlide)
 	{
-		std::string textHover = string_format("%i / %i", static_cast<int>(thumbPosition()), static_cast<int>(mLength));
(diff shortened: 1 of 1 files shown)


src/UI/Core/Slider.cpp
```diff
@@ -365,8 +365,6 @@ void Slider::update()
 void Slider::draw()
 {
 	Renderer& r = Utility<Renderer>::get();
-	std::string textHover;
-	int _x = 0, _y = 0, _w = 0, _h = 0;
 	float _thumbPosition = 0.0f;
```
cugone: If you're gonna removed the leading underscores on 85% of the variables, might as well go for that 100%. :)

DanRStevens: ... Oh, because I wasn't touching that part of the function. I was only really touching the last code block, to which the two lines of moved variables actually belonged.

It maybe seems odd to update variables in the rest of the function when I wasn't actually looking at updating any of the code there.

cugone: Okay. :/

PR #298, opened 2020-03-30
Refactor WarehouseInspector draw

DanRStevens:
Reference: #266

A bit of refactoring for the `WarehouseInspector` class. Allows for higher level text output. Removes use of `string_format`.

### DIFF
--- a/src/UI/WarehouseInspector.cpp
+++ b/src/UI/WarehouseInspector.cpp
@@ -93,7 +93,7 @@ void WarehouseInspector::update()
 	};
 
 	auto position = rect().startPoint() + NAS2D::Vector{constants::MARGIN, 25};
-	drawTitleText(position, "Storage", string_format("%i / %i", pool.availableStorage(), pool.capacity()), 0);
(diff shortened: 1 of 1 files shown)


src/UI/WarehouseInspector.cpp
```diff
@@ -82,19 +82,25 @@ void WarehouseInspector::update()
...
 
-	Renderer& r = Utility<Renderer>::get();
+	Renderer& renderer = Utility<Renderer>::get();
 
-	ProductPool& _pool = mWarehouse->products();
+	ProductPool& pool = mWarehouse->products();
 
-	r.drawText(*FONT_BOLD, "Storage", rect().x() + constants::MARGIN, rect().y() + 25.0f, 255, 255, 255);
-	r.drawText(*FONT, string_format("%i / %i", _pool.availableStorage(), _pool.capacity()), rect().x() + constants::MARGIN + FONT_BOLD->width("Storage") + 20, rect().y() + 25.0f, 255, 255, 255);
+	const auto drawTitleText = [&renderer](NAS2D::Point<int> position, std::string title, std::string text, int offset = 100) {
```
cugone: Consider changing the `std::string` parameter types to `const std::string&` or `std::string_view` because they don't change within the function.

PR #299, opened 2020-03-30
Refactor StructureInspector::drawPopulationRequirements

DanRStevens:
Reference: #266

A bit of refactoring for `StructureInspector::drawPopulationRequirements`. Reduced redundancy in text output code. Removes use of string_format.

### DIFF
--- a/src/UI/StructureInspector.cpp
+++ b/src/UI/StructureInspector.cpp
@@ -107,9 +107,9 @@ void StructureInspector::drawPopulationRequirements()
 		const auto& populationAvailable = mStructure->populationAvailable();
 		if (populationRequirements[populationType] > 0)
 		{
-			std::string format = populationTypes[populationType] + ": " + std::to_string(populationAvailable[populationType]) + "/" + std::to_string(populationRequirements[populationType]);
+			std::string text = populationTypes[populationType] + ": " + std::to_string(populationAvailable[populationType]) + "/" + std::to_string(populationRequirements[populationType]);
 			Color color = populationAvailable[populationType] >= populationRequirements[populationType] ? Color::White : Color::Red;
-			renderer.drawText(*FONT, format, position, color);
+			renderer.drawText(*FONT, text, position, color);
 			position.y() += 10;
 		}
 	}
--- a/src/UI/StructureInspector.cpp
+++ b/src/UI/StructureInspector.cpp
@@ -96,20 +96,20 @@ void StructureInspector::drawPopulationRequirements()
 	auto position = rect().startPoint() + NAS2D::Vector{10, 85};
 	r.drawText(*FONT_BOLD, "Population Required", position, NAS2D::Color::White);
 
-	position.y() += 20;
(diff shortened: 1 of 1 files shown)


src/UI/StructureInspector.cpp
```diff
@@ -91,28 +91,28 @@ void StructureInspector::btnCloseClicked()
...
-		std::string format = string_format("Workers: %i/%i", mStructure->populationAvailable()[0], mStructure->populationRequirements()[0]);
-		Color color = mStructure->populationAvailable()[0] >= mStructure->populationRequirements()[0] ? Color::White : Color::Red;
-		r.drawText(*FONT, format, posX, posY, color.red(), color.green(), color.blue());
-		posY += 10;
+	Renderer& renderer = Utility<Renderer>::get();
+
+	auto position = rect().startPoint() + NAS2D::Vector{10, 85};
+	renderer.drawText(*FONT_BOLD, "Population Required", position, NAS2D::Color::White);
+
+	const std::array<std::string, 2> populationTypes = {{
```
cugone: What happens if you declare this as:
```cpp
const std::array<std::string, 2> populationTypes{
"Workers",
"Scientists"
};
```
?

src/UI/StructureInspector.cpp
```diff
@@ -91,28 +91,28 @@ void StructureInspector::btnCloseClicked()
...
+		"Workers",
+		"Scientists"
+	}};
+
+	position.y() += 20;
+	for (std::size_t populationType = 0; populationType < populationTypes.size(); ++populationType) {
```
cugone: Consider `for(auto populationType = 0u; ...` or `for(auto populationType = std::size_t{0u} ...` if it gets deduced to the wrong type.

src/UI/StructureInspector.cpp
DanRStevens: Seems to work just fine.

src/UI/StructureInspector.cpp
DanRStevens: Using `auto` with `0u` would deduce the wrong type. It would give `unsigned int` rather than `std::size_t`. Depending on the platform, those types may be different.

I prefer putting the type on the left in the particular case. I don't see any clear benefit to the `= std::size_t{0u}` syntax. Seems to just add syntactic noise.

cugone: I was in the "almost always auto" mindset. This mostly applies to generic code or situations where the type is repeated. I'm okay with leaving it as `std::size_t populationType = 0`.

DanRStevens: Yeah. Counting `for` loops seems to be one of the areas where `auto` hasn't managed to penetrate well. I think there were supposed to be some improvements coming to make counting `for` loops easier to write, and play nicer with `auto`. I believe the change involved Ranges.

cugone: ...and iterators: `for(auto iter = std::begin(vecVar) ... `

PR #300, opened 2020-03-30
Refactor structure inspector update

DanRStevens:
A bit of refactoring for `StructureInspector::update`. Extract low level drawing details to a lambda.

### DIFF
--- a/src/UI/StructureInspector.cpp
+++ b/src/UI/StructureInspector.cpp
@@ -157,6 +157,7 @@ void StructureInspector::update()
 	}
 
 	drawPopulationRequirements();
-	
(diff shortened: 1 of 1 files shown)


src/UI/StructureInspector.cpp
```diff
@@ -124,34 +124,40 @@ void StructureInspector::update()
 	if (!visible()) { return; }
 	Window::update();
 
-	Renderer& r = Utility<Renderer>::get();
+	Renderer& renderer = Utility<Renderer>::get();
+
+	const auto drawTitleText = [&renderer](NAS2D::Point<int> position, std::string title, std::string text) {
```
cugone: Consider changing the `std::string` parameters to `const std::string&` or `std::string_view` because the function doesn't change them.

PR #305, opened 2020-03-30
Fix MSVC warnings in Pather

DanRStevens:
Use `std::size_t` for array indexes and allocation counts. Set a named constant for `Unset` (`-1`) values, of type `std::size_t`.

Fix MSVC warnings:
> C:\projects\ophd\src\MicroPather\micropather.cpp(622,42): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data [C:\projects\ophd\proj\vs2019\ophd.vcxproj]
>   C:\projects\ophd\src\MicroPather\micropather.cpp(629,57): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data [C:\projects\ophd\proj\vs2019\ophd.vcxproj]
>   C:\projects\ophd\src\MicroPather\micropather.cpp(629,57): warning C4267: 'initializing': conversion from 'size_t' to 'const unsigned int', possible loss of data [C:\projects\ophd\proj\vs2019\ophd.vcxproj]
>   C:\projects\ophd\src\MicroPather\micropather.cpp(642,7): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data [C:\projects\ophd\proj\vs2019\ophd.vcxproj]
>

### DIFF
--- a/src/MicroPather/micropather.cpp
+++ b/src/MicroPather/micropather.cpp
@@ -44,6 +44,9 @@ distribution.
 
 using namespace micropather;
 
+constexpr auto Unset = int(-1);
+
+
 class OpenQueue
 {
 public:
@@ -233,7 +236,7 @@ PathNodePool::~PathNodePool()
 
 bool PathNodePool::PushCache(const NodeCost* nodes, int nNodes, int* start)
 {
-	*start = -1;
+	*start = Unset;
 	if (nNodes + cacheSize <= cacheCap)
 	{
 		for (int i = 0; i < nNodes; ++i)
@@ -468,8 +471,8 @@ void PathNode::Init(unsigned _frame,
 void PathNode::Clear()
 {
 	memset(this, 0, sizeof(PathNode));
-	numAdjacent = -1;
-	cacheIndex = -1;
+	numAdjacent = Unset;
+	cacheIndex = Unset;
 }
 
 
--- a/src/MicroPather/micropather.cpp
+++ b/src/MicroPather/micropather.cpp
@@ -604,7 +604,7 @@ void MicroPather::GetNodeNeighbors(PathNode* node, std::vector< NodeCost >* pNod
 		// it has no neighbors.
 		pNodeCost->resize(0);
 	}
-	else if (node->cacheIndex < 0)
+	else if (node->cacheIndex == Unset)
 	{
 		// Not in the cache. Either the first time or just didn't fit. We don't know
 		// the number of neighbors and need to call back to the client.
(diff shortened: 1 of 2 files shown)


src/MicroPather/micropather.cpp
```diff
@@ -44,6 +45,9 @@ distribution.
 
 using namespace micropather;
 
+constexpr std::size_t Unset = std::numeric_limits<std::size_t>::max();
```
cugone: Setting an `unsigned` variable to `-1` has well-defined and understood wrapping behavior. Declaring it as `constexpr auto Unset = std::size_t{-1};` is shorter, less repetitive of the type, and conveys the same information.

DanRStevens: I'm afraid Clang complains about that:
```cpp
constexpr auto Unset = std::size_t{-1};
```

> error: constant expression evaluates to -1 which cannot be narrowed to type 'std::size_t' (aka 'unsigned long') [-Wc++11-narrowing]
> constexpr auto Unset = std::size_t{-1};

cugone: Try `constexpr auto Unset = std::size_t(-1u)` a non-brace-initialized version that doesn't enforce narrowing conversions.

DanRStevens: That works. Ran into this on another project, and someone resorted to using `numeric_limits` there.

DanRStevens:
I force pushed to rework this a bit. The order of commits has been changed. I think the new order makes more sense in terms of refactoring without causing addition warnings in the intermediate steps.

The `Unset` named constant is introduced and used first. It is initially an `int`, but later changed to `std::size_t` in the last commit, which is now the one that changes a bunch of `int` and `unsigned int` to `std::size_t`.

Overall, the result is effectively the same as before, plus the new suggestion.

src/MicroPather/micropather.cpp
cugone: Goody. That said, a comment documenting the intended behavior is probably a good idea.

DanRStevens: Ehh, I don't know there is much to comment. I think most programmers understand wraparound.

cugone: Actually, after examining the source code, there is an internal class-level maximum cache value `cacheCap` that limits the size of the path to some value "big enough" and is used everywhere `start` is used. Setting `Unset` to `cacheCap + 1` serves the same purpose and doesn't rely on wrap around.

DanRStevens: Yes, but `cacheCap` is a runtime value that can change. Much better to just use max and not worry about it.

Besides, this is for the code that was previously comparing to a hardcoded literal constant, which was previously set to `-1`. I'd rather not change the numerical value of the sentinel.

cugone: Hm. Okay.

PR #311, opened 2020-03-31
Refactor TileMap::draw

DanRStevens:
A bit of refactoring for `TileMap::draw`. This method had a warning fixed yesterday, though I didn't have time to refactor the rest of the method until today.

There's probably more that can be done, but this seems like a good stopping point. In particular, the `loopVar++` syntax can probably be changed to `++loopVar`, which is slightly more common in other files. However, this may require an update involving more than just the one function being refactored here, so I thought I'd leave it for a blanket update.

I made use of nested ternary operators to select the `highlightColor`. Normally I would consider nested ternary to be a bad idea, and something to avoid. It's usually not very readable. I thought I'd try it out here though, since it results in quite significantly more compact code, and the ability to declare a variable `const`. Hopefully the formatting helps keep it readable.

### DIFF
--- a/src/Map/TileMap.cpp
+++ b/src/Map/TileMap.cpp
@@ -356,18 +356,18 @@ void TileMap::draw()
 				{
 					if (mShowConnections && tile->connected())
 					{
-						r.drawSubImage(mTileset, position, subImageRect, NAS2D::Color{71, 224, 146, 255});
+						r.drawSubImage(mTileset, position, subImageRect, NAS2D::Color{71, 224, 146});
 					}
 					else
 					{
-						r.drawSubImage(mTileset, position, subImageRect, NAS2D::Color{125, 200, 255, 255});
+						r.drawSubImage(mTileset, position, subImageRect, NAS2D::Color{125, 200, 255});
 					}
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


src/Map/TileMap.cpp
```diff
@@ -329,62 +329,40 @@ bool TileMap::tileHighlightVisible() const
...
-					if (mShowConnections && tile->connected())
-					{
-						r.drawSubImage(mTileset, x, y, tile->index() * TILE_WIDTH, tsetOffset, TILE_WIDTH, TILE_HEIGHT, 0, 255, 0, 255);
-					}
-					else
-					{
-						r.drawSubImage(mTileset, x, y, tile->index() * TILE_WIDTH, tsetOffset, TILE_WIDTH, TILE_HEIGHT);
-					}
-				}
+				auto position = mMapPosition + NAS2D::Vector{(col - row) * TILE_HALF_WIDTH, (col + row) * TILE_HEIGHT_HALF_ABSOLUTE};
```
cugone: I'm not seeing `position` change (this is a rendering function, so it shouldn't) so it probably can be declared `const`.

cugone:
`++loopVar` is also more performant because it doesn't copy.

Issue #307, opened 2020-03-30
Increase Clang warnings


DanRStevens:
I tried compiling with Clang using the `-Weverything` flag to see what sorts of warnings were emitted. I then went through and turned flags off until no more warnings were generated. The following command ended up being used:
```
make WARN_EXTRA="-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-global-constructors -Wno-exit-time-destructors -Wno-padded -Wno-weak-vtables -Wno-implicit-int-float-conversion -Wno-inconsistent-missing-destructor-override -Wno-missing-variable-declarations -Wno-extra-semi -Wno-extra-semi-stmt -Wno-header-hygiene -Wno-missing-prototypes -Wno-newline-eof -Wno-ctad-maybe-unsupported -Wno-switch-enum -Wno-float-equal -Wno-documentation -Wno-documentation-unknown-command -Wno-covered-switch-default -Wno-zero-as-null-pointer-constant -Wno-format-nonliteral -Wno-unreachable-code-break -Wno-unreachable-code-return -Wno-unreachable-code -Wno-deprecated-copy-dtor -Wno-comma -Wno-float-conversion -Wno-implicit-int-conversion -Wno-sign-conversion"
```

Some flags are of no interest, however many of these seems to suggest small issues that we should perhaps resolve.

cugone:
I don't know what any of those do. >.>

DanRStevens:
The following Clang warnings have been fixed:
- `-Wextra-semi`
- `-Wheader-hygiene`
- `-Wnewline-eof`
- `-Winconsistent-missing-destructor-override`
- `-Wformat-nonliteral` (#266)

The following Clang warning has had partial fixes applied:
- `-Wdocumentation-unknown-command`

The following GCC warning has been fixed:
- `-Wsuggest-override`

----

Some warnings still remain. To compile with Clang using `-Weverything`, the following command is used to silence remaining warnings:
```
[...]

PR #316, opened 2020-03-31
Fix comment

DanRStevens:
Reference: #307 (`-Wdocumentation-unknown-command`)

Fix unknown Doxygen command. Corrected associated documentation.


DanRStevens:
Probably not. Though I suppose it depends on how extensive the change is, and how involved the text is.

PR #317, opened 2020-03-31
Enum class Button::Type

DanRStevens:
Convert `Button::Type` from `enum` to `enum class`.

There are many other similar enums, which will eventually be converted. I'm starting with this one as a template. The bulk of the changes are simply adding the enum name prefix when accessing enum values. That can be done safely before switching to an `enum class`. Any code that relies on implicit conversion of `enum` to `int` or other data types will also need to be updated.

### DIFF
--- a/src/States/MapViewStateUi.cpp
+++ b/src/States/MapViewStateUi.cpp
@@ -131,11 +131,11 @@ void MapViewState::initUi()
 
 	mBtnToggleHeightmap.image("ui/icons/height.png");
 	mBtnToggleHeightmap.size(static_cast<float>(constants::MAIN_BUTTON_SIZE));
-	mBtnToggleHeightmap.type(Button::BUTTON_TOGGLE);
+	mBtnToggleHeightmap.type(Button::Type::BUTTON_TOGGLE);
 
 	mBtnToggleConnectedness.image("ui/icons/connection.png");
 	mBtnToggleConnectedness.size(static_cast<float>(constants::MAIN_BUTTON_SIZE));
-	mBtnToggleConnectedness.type(Button::BUTTON_TOGGLE);
+	mBtnToggleConnectedness.type(Button::Type::BUTTON_TOGGLE);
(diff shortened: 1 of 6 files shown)


cugone:
I'd also prefer removing the type prefix to the enum *values* as well: `Button::Type::TOGGLE` and `Button::Type::NORMAL`

A separate PR can be made to remove the ALL_CAPS naming.

DanRStevens:
As a side note, I did a search for `enum` definitions which were not `enum class` by using a negative look ahead regex:
[Regular Expression in Atom Text Editor, exclude string “abc” from beginning of string](https://stackoverflow.com/questions/40798179/regular-expression-in-atom-text-editor-exclude-string-abc-from-beginning-of-s)

Find: `enum (?!class)`
Replace: `enum class `

The find is potentially more useful than the replace, since each changes results in many errors due to the name prefix requirement of values for `enum class`.

DanRStevens:
I would also like to do some renaming, though that was something I had planned to defer on.

I figure the rename and dropping ALL_CAPS can be done at the same time.

cugone:
Why defer when you can do it all right now. You're already changing the point of use anyway.

DanRStevens:
Technically it only changes the ones that aren't already prefixed. Though that is likely to be all of them, or very nearly all of them.

Mostly though I just didn't want to think that deeply about the change, nor invest the time in renaming. I wanted something quick and mechanical that would let me know how widespread the changes would need to be. Turns out, quite a lot.

In terms of renaming the fields, if they are at least prefixed first, it makes search/replace of text safer later. Though with the prefix built into the old names, I suppose they're already pretty unique.

I was also considering changes that would make certain enums obsolete. For instance, it's not likely someone would change a Button between Normal and Toggle at runtime, after the Button was created. Perhaps it should be part of the type, with two different subclasses, rather than a parameter.

Still making decisions on where I want to go with some of these.

cugone:
Consider installing Visual Assist for Visual Studio. It makes renaming *much* easier.

DanRStevens:
That's true. I do have at least temporary access to Visual Studio now.

PR #322, opened 2020-04-01
Don't override empty destructors

DanRStevens:
I noticed this while doing some related investigation for #307, with warning flag `-Winconsistent-missing-destructor-override`. This doesn't address that issue, but was some related cleanup to reduce the work that directly affects those warnings.

Basically, if we're not actually overriding the destructor, then don't override it.

### DIFF
--- a/src/Things/Structures/MineFacility.h
+++ b/src/Things/Structures/MineFacility.h
@@ -13,7 +13,7 @@ public:
 	typedef NAS2D::Signals::Signal<MineFacility*> ExtensionCompleteCallback;
 public:
 	MineFacility(Mine* mine);
-	
+
 	void mine(Mine* mine) { mMine = mine; }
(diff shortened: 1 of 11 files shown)


src/Things/Structures/MineFacility.h
```diff
@@ -13,7 +13,7 @@ class MineFacility: public Structure
 	typedef NAS2D::Signals::Signal<MineFacility*> ExtensionCompleteCallback;
 public:
 	MineFacility(Mine* mine);
-	
+
```
DanRStevens, line 16: I noticed this a little after pushing. This was the results of resolving a merge conflict. I had started on these changes days ago, but then put them aside. Meanwhile other changes modified these lines. The difference between the two branches, was whether or not a stray tab was left around.

Considering the tab should be removed anyway, I'm just going to leave it.

PR #326, opened 2020-04-02
Refactor MapViewState::drawResourceInfo

DanRStevens:
Reference: #266

Refactor similar code into data driven loops. Replace use of `string_format` with `std::to_string`.

### DIFF
--- a/src/States/MapViewStateDraw.cpp
+++ b/src/States/MapViewStateDraw.cpp
@@ -162,11 +162,9 @@ void MapViewState::drawResourceInfo()
 		std::tuple{NAS2D::Rectangle{112, 16, iconSize, iconSize}, mPlayerResources.rareMinerals(), 0},
 	};
 
-	NAS2D::Color color;
-
 	for (const auto& [imageRect, amount, spacing] : resources) {
 		renderer.drawSubImage(mUiIcons, position, imageRect);
-		color = (amount <= 10) ? glowColor : NAS2D::Color::White;
+		const auto color = (amount <= 10) ? glowColor : NAS2D::Color::White;
 		renderer.drawText(*MAIN_FONT, std::to_string(amount), position + textOffset, color);
 		position.x() += spacing;
 	}
@@ -181,7 +179,7 @@ void MapViewState::drawResourceInfo()
 	position.x() += x + offsetX;
 	for (const auto& [imageRect, parts, total, isHighlighted] : storageCapacities) {
 		renderer.drawSubImage(mUiIcons, position, imageRect);
-		color = isHighlighted ? glowColor : NAS2D::Color::White;
+		const auto color = isHighlighted ? glowColor : NAS2D::Color::White;
 		renderer.drawText(*MAIN_FONT, NAS2D::string_format("%i/%i", parts, total), position + textOffset, color);
 		position.x() += (x + offsetX) * 2;
 	}
--- a/src/States/MapViewStateDraw.cpp
+++ b/src/States/MapViewStateDraw.cpp
@@ -180,7 +180,8 @@ void MapViewState::drawResourceInfo()
 	for (const auto& [imageRect, parts, total, isHighlighted] : storageCapacities) {
 		renderer.drawSubImage(mUiIcons, position, imageRect);
 		const auto color = isHighlighted ? glowColor : NAS2D::Color::White;
-		renderer.drawText(*MAIN_FONT, NAS2D::string_format("%i/%i", parts, total), position + textOffset, color);
+		const auto text = std::to_string(parts) + "/" + std::to_string(total);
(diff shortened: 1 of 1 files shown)


src/States/MapViewStateDraw.cpp
```diff
@@ -154,63 +154,39 @@ void MapViewState::drawResourceInfo()
...
-	renderer.drawSubImage(mUiIcons, position, commonMineralsImageRect);
-	bool shouldCommonMineralsGlow = mPlayerResources.commonMinerals() <= 10;
-	color = shouldCommonMineralsGlow ? glowColor : NAS2D::Color::White;
-	renderer.drawText(*MAIN_FONT, std::to_string(mPlayerResources.commonMinerals()), position + textOffset, color);
+	constexpr auto iconSize = constants::RESOURCE_ICON_SIZE;
+	const std::array resources{
+		std::tuple{NAS2D::Rectangle{64, 16, iconSize, iconSize}, mPlayerResources.commonMetals(), offsetX},
+		std::tuple{NAS2D::Rectangle{80, 16, iconSize, iconSize}, mPlayerResources.rareMetals(), x + offsetX},
+		std::tuple{NAS2D::Rectangle{96, 16, iconSize, iconSize}, mPlayerResources.commonMinerals(), x + offsetX},
+		std::tuple{NAS2D::Rectangle{112, 16, iconSize, iconSize}, mPlayerResources.rareMinerals(), 0},
```
cugone, line 162: Any reason you didn't use `std::make_tuple`?

src/States/MapViewStateDraw.cpp
```diff
@@ -154,63 +154,39 @@ void MapViewState::drawResourceInfo()
...
+	};
 
-	// Rare Minerals
-	position.x() += x + offsetX;
-	const auto rareMineralsImageRect = NAS2D::Rectangle<int>{112, 16, constants::RESOURCE_ICON_SIZE, constants::RESOURCE_ICON_SIZE};
-	renderer.drawSubImage(mUiIcons, position, rareMineralsImageRect);
-	bool shouldRareMineralsGlow = mPlayerResources.rareMinerals() <= 10;
-	color = shouldRareMineralsGlow ? glowColor : NAS2D::Color::White;
-	renderer.drawText(*MAIN_FONT, std::to_string(mPlayerResources.rareMinerals()), position + textOffset, color);
+	for (const auto& [imageRect, amount, spacing] : resources) {
```
cugone, line 165: For consistency with rest of project, there should be a newline between the for loop and the opening brace.
```cpp
for ...
{
```

src/States/MapViewStateDraw.cpp
```diff
@@ -154,63 +154,39 @@ void MapViewState::drawResourceInfo()
...
-	renderer.drawText(*MAIN_FONT, NAS2D::string_format("%i/%i", foodInStorage(), foodTotalStorage()), position + textOffset, color);
-
-	// Energy
-	position.x() += (x + offsetX) * 2;
-	const auto energyImageRect = NAS2D::Rectangle<int>{80, 32, constants::RESOURCE_ICON_SIZE, constants::RESOURCE_ICON_SIZE};
-	renderer.drawSubImage(mUiIcons, position, energyImageRect);
-	bool shouldEnergyGlow = mPlayerResources.energy() <= 5;
-	color = shouldEnergyGlow ? glowColor : NAS2D::Color::White;
-	renderer.drawText(*MAIN_FONT, NAS2D::string_format("%i/%i", mPlayerResources.energy(), NAS2D::Utility<StructureManager>::get().totalEnergyProduction()), position + textOffset, color);
+	for (const auto& [imageRect, parts, total, isHighlighted] : storageCapacities) {
```
cugone, line 180: Brace consistency.

src/States/MapViewStateDraw.cpp
DanRStevens: There's no reason to use `make_tuple`. The deduction guide added for `std::tuple` means you can just construct directly, rather than going through a factory function. The point of the factory function was so template type deduction works. In earlier versions of C++, template type deduction only worked for functions, not for classes.

src/States/MapViewStateDraw.cpp
DanRStevens: Hmm, this syntax has slipped through for a few other recent commits.

src/States/MapViewStateDraw.cpp
cugone: Ok.

For Reference: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t44-use-function-templates-to-deduce-class-template-argument-types-where-feasible

DanRStevens: Interesting. They added a note for C++17:

> Note that C++17 will make this rule redundant by allowing the template arguments to be deduced directly from constructor arguments

PR #330, opened 2020-04-03
Use `int` for residence capacity (percentage)

DanRStevens:
Reference: #325

Change residence capacity to use integers. (Percentage 0-100+).

The code was designed to expect whole numbers, and to round when they weren't. Might as well just use integers.

Values over 100 (%) are expected. There are likely no negatives, so unsigned may be a reasonable consideration here. Other related expressions appear to use `int` though, so I thought I'd try being consistent.

### DIFF
--- a/src/States/MapViewStateTurn.cpp
+++ b/src/States/MapViewStateTurn.cpp
@@ -151,10 +151,10 @@ void MapViewState::updateMorale()
 	mCurrentMorale -= sm.disabled();
 	mCurrentMorale -= sm.destroyed();
 
-	int residentialMoraleHit = static_cast<int>(mPopulationPanel.capacity() / 100.0f);
+	int residentialMoraleHit = mPopulationPanel.capacity() / 100;
 
 	// Ensure that there is always a morale hit if residential capacity is more than 100%.
-	if (mPopulationPanel.capacity() > 100.0f && residentialMoraleHit < constants::MINIMUM_RESIDENCE_OVERCAPACITY_HIT) { residentialMoraleHit = constants::MINIMUM_RESIDENCE_OVERCAPACITY_HIT; }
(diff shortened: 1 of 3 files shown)


cugone:
Stupid question:
Why can't it be calculated as `float resMoraleHitNormalized = totalPop / (float)resCapacity` in order to normalize the value, then just check if the result is over `1.0`? Or, better yet, a direct scalar of the default morale hit (itself based on the difficulty):

```cpp
float resOvercrowdedNormalized = totalPop / (float)resCapacity;
// truncate to make non-overcrowding (less than 1.0) not affect morale at all.
int resMoraleHit = (int)resOvercrowdedNormalized * baseMoraleHitOverCrowding;
```

The normalized value can also be used to color the text from green to yellow to red:

Pseudo-code:
```cpp
float resMoraleColor = std::clamp(resOvercrowdedNormalized, 0.0f, 1.0f);
drawText("Residence Capacity: " + std::to_string(getResidenceCapacity()), NAS2D::Color{(uint8_t)(255 * resMoraleColor), (uint8_t)(255 * (1.0f - resMoraleColor)), 0})
```

DanRStevens:
I'm not really sure there's an answer to that. Sure you can use normalized values from `0.0` to `1.0` instead of `0` to `100`. That's just not how the code is written.

I figure if you're scaling things up to larger integer values, you might as well just use integer. Besides, it saves having to do conversion to/from floating point, which is actually a somewhat expensive operation. I don't think it's enough to ever be a performance problem given the context, but ehh, if it doesn't have to be done, why bother doing it.

In terms of scaling color, you can do the same with integer math. Just use [fixed point](https://en.wikipedia.org/wiki/Fixed-point_arithmetic). In this case, it would be much more efficient, since both inputs and outputs are integer based, and fixed point uses the integer portion of the CPU. No expensive conversions necessary.

PR #331, opened 2020-04-03
Add CcNotPlaced constant with invalid tile coordinates

### LINKED ISSUE #292: Refactor COMMAND_CENTER_LOCATION to default to an invalid tile
See https://github.com/OutpostUniverse/OPHD/pull/289#discussion_r399823337

DanRStevens:
Closes #292

Refactor uses of `ccLocation`, and introduces a `CcNotPlaced` constant, with invalid tile coordinates.

----

At some point we should add `size`/`length` support to `NAS2D::Vector`.

### DIFF
--- a/src/Map/Tile.cpp
+++ b/src/Map/Tile.cpp
@@ -125,9 +125,9 @@ Robot* Tile::robot()
 /**
  * 
  */
-float Tile::distanceTo(Tile* t)
+float Tile::distanceTo(Tile* tile)
 {
-	return distanceTo(t->position());
+	return distanceTo(tile->position());
 }
 
 
--- a/src/Map/Tile.cpp
+++ b/src/Map/Tile.cpp
@@ -127,9 +127,7 @@ Robot* Tile::robot()
  */
 float Tile::distanceTo(Tile* t)
 {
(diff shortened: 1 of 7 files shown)


src/Map/Tile.cpp
```diff
@@ -125,9 +125,14 @@ Robot* Tile::robot()
...
+float Tile::distanceTo(Tile* tile)
 {
-	int x = t->x() - Tile::x();
-	int y = t->y() - Tile::y();
-	return static_cast<float>(sqrt((x * x) + (y * y)));
+	return distanceTo(tile->position());
+}
+
+
+float Tile::distanceTo(NAS2D::Point<int> point)
```
cugone, line 134: The argument name and type is unclear whether or not this is tile space (i.e. tile coords), or world space (pixel-perfect, sub-tile coordinates).

Additionally, are you counting from the corner of a tile or the center?

DanRStevens: The name isn't really the concern of this PR. It used an existing name. This should be addressed as it's own issue.

DanRStevens: And to add to that, it's on the `Tile` class, so that should give a hint.

PR #335, opened 2020-04-04
Add new `TileMap::getTile` overload, for default depth

DanRStevens:
Noticed this was missing from the API, and will be needed to refactor various functions.

### DIFF
--- a/src/Map/TileMap.h
+++ b/src/Map/TileMap.h
@@ -34,7 +34,7 @@ public:
 	Tile* getTile(int x, int y, int level);
 	Tile* getTile(int x, int y) { return getTile(x, y, mCurrentDepth); }
 	Tile* getTile(NAS2D::Point<int> position, int level) { return getTile(position.x(), position.y(), level); }
-	Tile* getTile(NAS2D::Point<int> position) { return getTile(position.x(), position.y(), mCurrentDepth); }
+	Tile* getTile(NAS2D::Point<int> position) { return getTile(position, mCurrentDepth); }
 
 	Tile* getVisibleTile(NAS2D::Point<int> position, int level);
 	Tile* getVisibleTile() { return getVisibleTile(tileMouseHover(), mCurrentDepth); }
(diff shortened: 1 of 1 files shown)


src/Map/TileMap.h
```diff
@@ -34,6 +34,7 @@ class TileMap: public micropather::Graph
 	Tile* getTile(int x, int y, int level);
 	Tile* getTile(int x, int y) { return getTile(x, y, mCurrentDepth); }
 	Tile* getTile(NAS2D::Point<int> position, int level) { return getTile(position.x(), position.y(), level); }
+	Tile* getTile(NAS2D::Point<int> position) { return getTile(position.x(), position.y(), mCurrentDepth); }
```
cugone, line 37: I prefer delegates of delegates to make sure there's only one point of failure:

```cpp
Tile* getTile(int x, int y, int level) { /* Lowest common delegate where all the work is performed */ }
Tile* getTile(int x, int y) { return getTile(x, y, mCurrentDepth); }
Tile* getTile(NAS2D::Point<int> position, int level) { return getTile(position.x(), position.y(), level); }
Tile* getTile(NAS2D::Point<int> position) { return getTile(position, mCurrentDepth); }
```

This makes sure that there is only one point of failure and if there's a problem with the code, there's only one place that needs to be fixed.

DanRStevens: I'm not sure I understand. They all call the same basic function here, as you've marked. There is only one real implementation.

cugone: Your implementation of `Tile* getTile(NAS2D::Point<int> position)` is calling `Tile* getTile(int x, int y, int level)` instead of `Tile* getTile(NAS2D::Point<int> position, int level)`

DanRStevens: Yes I see that. I don't quite understand the reasoning behind the comment though.

There is only one real implementation, and hence one "point of failure". You're proposing it be changed to go through an extra intermediary wrapper method before getting to the implementation. I don't see how that fits with what you wrote.

To be clear, I'm not opposed to changing it to your suggestion. I expect the unpacked methods may eventually be removed, so any delegating will need to go through the packed method when that happens.

cugone: > 
> To be clear, I'm not opposed to changing it to your suggestion. I expect the unpacked methods may eventually be removed, so any delegating will need to go through the packed method when that happens.

Exactly. With my suggestion, when the unpacked functions are removed the only function that has to change, besides the removals, is `getTile(Point<int> position, int level)` instead of all of them.

PR #341, opened 2020-04-05
Suggest override UI Core

DanRStevens:
Reference: #307 (`-Wsuggest-override`), #338, #322
Related: #339, #340

Add suggested uses of override to member functions in the `UI/Core/` folder.


DanRStevens:
Missed some stuff on the first go. The `-Wsuggest-override` flag is quite noisy, since it concerns header files, which have a way of bringing in other header files, and the suggestion affects many areas of the code.

PR #353, opened 2020-04-07
Deprecated copy dtor

DanRStevens:
Reference: #307 (`-Wdeprecated-copy-dtor`)

A copy or move constructor is needed for types to be usable by certain standard containers. This is needed to support operations such as `std::vector::resize`, or construction of `std::map`.

Having an explicit destructor can suppress generation of an automatic copy constructor.

Sample warning message:

> warning: definition of implicit copy constructor for 'ProductionCost' is deprecated because it has a user-declared destructor [-Wdeprecated-copy-dtor]

Reference: [Copy constructors](https://en.cppreference.com/w/cpp/language/copy_constructor)
> The generation of the implicitly-defined copy constructor is deprecated if T has a user-defined destructor or user-defined copy assignment operator.

### DIFF
--- a/OPHD/Map/Tile.h
+++ b/OPHD/Map/Tile.h
@@ -17,6 +17,7 @@ class Tile
 public:
 	Tile() = default;
 	Tile(Tile&& other);
+	Tile& operator=(Tile&& other);
 	~Tile();
 
 	int index() const { return mIndex; }
--- a/OPHD/Map/Tile.h
+++ b/OPHD/Map/Tile.h
@@ -16,6 +16,8 @@ class Tile
 {
 public:
 	Tile() = default;
+	Tile(const Tile& other) = delete;
+	Tile& operator=(const Tile& other) = delete;
 	Tile(Tile&& other);
 	Tile& operator=(Tile&& other);
 	~Tile();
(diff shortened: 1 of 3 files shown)


OPHD/Map/Tile.h
```diff
@@ -15,7 +15,8 @@
 class Tile
 {
 public:
-	Tile();
+	Tile() = default;
+	Tile(Tile&& other);
```
cugone, line 19: Rule of Six violation. Either default, explicitly define or delete a copy ctor for shallow copies, deep copies, or making `Tile` non-copyable respectively.

Right now, I would recommend a deep copy since the tile owns `mMine` but not `mThing`...which makes no sense. (See next).

It shouldn't own either (in that case, a `default` would work). There should be an owned-by-map, map-wide, collection of entities (`std::map<Point<int>,std::unique_ptr<Thing>> mThings;`) and another (optional, albeit makes referencing *just* structures easier with a range-based for) non-owning collection of `Structures*`: `std::vector<Structure*>`.

DanRStevens: The destructor tries to delete both pointers. That suggests ownership of both those pointers.

I think copy should be disallowed. A move only class would be easier to deal with, and I don't think allowing copies is intentional.

PR #357, opened 2020-04-07
ListBox clipRect bounds

DanRStevens:
As mentioned in #365, this adjusts the `clipRect` to match the bounds of the control. Previously the `clipRect` extended beyond the bounds of the control. Overflow was 1 pixel to the left, and 1 pixel to the bottom.

We should check the border boxes and highlight boxes to see if they need adjustment to stay within the bounds of the control. None of them are drawn with such offsets though, so they should be fine.


DanRStevens:
Yes, most curious. I suspect maybe a previous version drew the highlight boxes or bounding boxes around the list items, and added to the size to contain them. That doesn't seem to be the case now.

PR #359, opened 2020-04-08
Refactor ResourceBreakdownPanel::update

DanRStevens:
Reference: #217

Refactor `ResourceBreakdownPanel::update` to use packed parameter methods and a data drive loop.

### DIFF
--- a/OPHD/UI/ResourceBreakdownPanel.cpp
+++ b/OPHD/UI/ResourceBreakdownPanel.cpp
@@ -128,9 +128,9 @@ void ResourceBreakdownPanel::update()
 		r.drawSubImage(mIcons, position, imageRect);
 		r.drawText(*FONT, text, position + NAS2D::Vector{23, 0}, NAS2D::Color::White);
 		const auto valueString = std::to_string(value);
-		r.drawText(*FONT, valueString, position + NAS2D::Vector{200 - 5 - FONT->width(valueString), 0}, NAS2D::Color::White);
+		r.drawText(*FONT, valueString, position + NAS2D::Vector{195 - FONT->width(valueString), 0}, NAS2D::Color::White);
 		const auto changeIconImageRect = NAS2D::Rectangle<int>::Create(ICON_SLICE[resourceTrend], NAS2D::Vector{8, 8});
-		r.drawSubImage(mIcons, position + NAS2D::Vector{220 - 5, 3}, changeIconImageRect);
+		r.drawSubImage(mIcons, position + NAS2D::Vector{215, 3}, changeIconImageRect);
 		r.drawText(*FONT, formatDiff(value - oldValue), position + NAS2D::Vector{235, 0}, valueChangeColor);
 		position.y() += 18;
 	}
--- a/OPHD/UI/ResourceBreakdownPanel.cpp
+++ b/OPHD/UI/ResourceBreakdownPanel.cpp
@@ -107,8 +107,8 @@ void ResourceBreakdownPanel::resourceCheck()
 
 void ResourceBreakdownPanel::update()
 {
-	Renderer& r = Utility<Renderer>::get();
-	r.drawImageRect(rect(), mSkin);
+	Renderer& renderer = Utility<Renderer>::get();
+	renderer.drawImageRect(rect(), mSkin);
 
 	const auto commonMetalImageRect = NAS2D::Rectangle{64, 16, 16, 16};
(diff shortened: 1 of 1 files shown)


OPHD/UI/ResourceBreakdownPanel.cpp
```diff
@@ -107,46 +107,31 @@ void ResourceBreakdownPanel::resourceCheck()
...
-
-	fmt = formatDiff(mPlayerResources->rareMinerals() - mPreviousResources.rareMinerals());
-	r.drawText(*FONT, fmt, 235.0f, rect().y() + 59.0f, RARE_MIN_COL.red(), RARE_MIN_COL.green(), RARE_MIN_COL.blue());
+	Renderer& renderer = Utility<Renderer>::get();
+	renderer.drawImageRect(rect(), mSkin);
+
+	const auto commonMetalImageRect = NAS2D::Rectangle{64, 16, 16, 16};
+	const auto rareMetalImageRect = NAS2D::Rectangle{80, 16, 16, 16};
+	const auto commonMineralImageRect = NAS2D::Rectangle{96, 16, 16, 16};
+	const auto rareMineralImageRect = NAS2D::Rectangle{112, 16, 16, 16};
```
cugone, line 116: You can increase readability by pre-declaring the dimensions and y-position:
```cpp
const auto resourceUiYPosition = 16;
const auto resourceUiDimensions = NAS2D::Vector{16,16};
const auto commonMetalImageRect = NAS2D::Rectangle{64, resourceUiYPosition, resourceUiDimensions};
const auto rareMetalImageRect = NAS2D::Rectangle{80, resourceUiYPosition, resourceUiDimensions};
const auto commonMineralImageRect = NAS2D::Rectangle{96, resourceUiYPosition, resourceUiDimensions};
const auto rareMineralImageRect = NAS2D::Rectangle{112, resourceUiYPosition, resourceUiDimensions};
//...
```

OPHD/UI/ResourceBreakdownPanel.cpp
```diff
@@ -107,46 +107,31 @@ void ResourceBreakdownPanel::resourceCheck()
...
+
+	std::array resources{
+		std::tuple{commonMetalImageRect, "Common Metals", mPlayerResources->commonMetals(), mPreviousResources.commonMetals(), COMMON_METALS, COMMON_MET_COL},
+		std::tuple{rareMetalImageRect, "Rare Metals", mPlayerResources->rareMetals(), mPreviousResources.rareMetals(), RARE_METALS, COMMON_MIN_COL},
+		std::tuple{commonMineralImageRect, "Common Minerals", mPlayerResources->commonMinerals(), mPreviousResources.commonMinerals(), COMMON_MINERALS, RARE_MET_COL},
+		std::tuple{rareMineralImageRect, "Rare Minerals", mPlayerResources->rareMinerals(), mPreviousResources.rareMinerals(), RARE_MINERALS, RARE_MIN_COL},
+	};
```
cugone, line 123: Const-correctness can be applied here. The `std::array` definitely, Not sure about the `std::tuple` though, you can always try and see.

OPHD/UI/ResourceBreakdownPanel.cpp
DanRStevens: I had thought about that briefly, but quickly decided I didn't like the style. It tries to reduce repetition, but ends up being more verbose in the end.

I suppose you could calculate the `Rectangle` inside the loop, based on the regular change to `x`. That was another thing I had considered.

I think more long term though, I want those Rectangle values to be read from the data files. They really shouldn't be defined as literals in the source code. That's just a detail we need to deal with for now until we can get the sprite code and data refactored. I'm not quite prepared to tackle that problem yet though.

In terms of where I want to go with this long term, I think having fully formed `Rectangle`s available for use is what it's going to end up like, so trying to build them up from components maybe goes a bit against the style in the long term plan.

OPHD/UI/ResourceBreakdownPanel.cpp
DanRStevens: Yes! For some reason I keep forgetting `const` when defining these `std::array`s.

cugone: Did it work with the `std::tuple`s?

DanRStevens: Err..., no it doesn't. "Expected expression". Though I don't think it would be meaningful to try and put `const` in there anyway.

PR #365, opened 2020-04-08
Brace initialization of Color

DanRStevens:
Use brace initialization of `Color` objects.

Remove explicit full alpha channels, and use named color constants where possible.


DanRStevens:
Hmm, missed a few, where the variable name came between `Color` and the initialization values.

Issue #319, opened 2020-03-31
MSVC Code Analysis warnings suggesting `enum class`


DanRStevens:
Running MSVC's Code Analysis (Analyze -> Run Code Analysis -> On Solution) gives many warnings, most of which suggest changing `enum` to `enum class`.

Warnings filtered for `enum class` suggestions:
```
C:\Users\User\source\repos\vcpkg\installed\x64-windows\include\SDL2\SDL_rect.h(102): warning C26812: The enum type 'SDL_bool' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\nas2d-core\include\NAS2D\Signal.h(65): warning C26812: The enum type 'DiggerDirection::DiggerSelection' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Things\Structures\Structure.h(68): warning C26812: The enum type 'Structure::StructureState' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Things\Structures\Structure.h(72): warning C26812: The enum type 'DisabledReason' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Things\Structures\Structure.h(79): warning C26812: The enum type 'IdleReason' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Things\Structures\Structure.h(103): warning C26812: The enum type 'Structure::StructureClass' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Things\Structures\Structure.h(104): warning C26812: The enum type 'ConnectorDir' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Common.cpp(82): warning C26812: The enum type 'MineProductionRate' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Common.cpp(232): warning C26812: The enum type 'ProductType' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\Mine.h(37): warning C26812: The enum type 'MineProductionRate' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
Z:\OPHD\src\GraphWalker.cpp(13): warning C26812: The enum type 'Direction' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).
[...]

PR #370, opened 2020-04-09
Fix other Code Analysis warnings

### LINKED ISSUE #320: MSVC Code Analysis warnings - Other
Related: #319

Running MSVC's Code Analysis (Analyze -> Run Code Analysis -> On Solution) gives a number of warnings. Most were suggestions to use `enum class` (#319). The remainder, not related to `enum class` are:
```
Z:\OPHD\src\Map\TileMap.cpp(76): warning C26451: Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(76): warning C26451: Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(77): warning C26451: Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(78): warning C26451: Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(78): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(80): warning C26451: Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(81): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(83): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(83): warning C26451: Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(84): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(85): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(85): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(340): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\Map\TileMap.cpp(340): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\UI\Core\ListBox.cpp(244): warning C26451: Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
Z:\OPHD\src\UI\Core\ListBox.cpp(244): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\UI\Core\Slider.cpp(27): warning C26495: Variable 'Slider::mSliderType' is uninitialized. Always initialize a member variable (type.6).
Z:\OPHD\src\MicroPather\micropather.cpp(53): warning C26495: Variable 'OpenQueue::sentinelMem' is uninitialized. Always initialize a member variable (type.6).
Z:\OPHD\src\MicroPather\micropather.h(222): warning C26495: Variable 'micropather::PathNode::child' is uninitialized. Always initialize a member variable (type.6).
Z:\OPHD\src\MicroPather\micropather.cpp(298): warning C6011: Dereferencing NULL pointer 'block'. 
Z:\OPHD\src\MicroPather\micropather.cpp(848): warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).
Z:\OPHD\src\MicroPather\micropather.cpp(959): warning C28182: Dereferencing NULL pointer. 'child' contains the same NULL value as 'inOpen' did. See line 937 for an earlier location where this can occur
```

cugone:
~~Branched off #319~~ Edit: (Rebased by Dan)

Closes #320

### DIFF
--- a/OPHD/MicroPather/micropather.cpp
+++ b/OPHD/MicroPather/micropather.cpp
@@ -848,9 +848,10 @@ void MicroPather::GetCacheData(CacheData* data)
 
 		data->hit = pathCache->hit;
 		data->miss = pathCache->miss;
-		if (data->hit + data->miss)
+		const auto hitAsFloat = static_cast<float>(data->hit);
+		if (const float total = hitAsFloat + data->miss)
 		{
-			data->hitFraction = static_cast<float>(static_cast<double>(data->hit) / (static_cast<double>(data->hit) + data->miss));
+			data->hitFraction = hitAsFloat / total;
 		}
 		else
 		{
--- a/OPHD/MicroPather/micropather.cpp
+++ b/OPHD/MicroPather/micropather.cpp
@@ -850,7 +850,7 @@ void MicroPather::GetCacheData(CacheData* data)
 		data->miss = pathCache->miss;
 		if (data->hit + data->miss)
 		{
-			data->hitFraction = static_cast<float>(static_cast<double>(data->hit) / static_cast<double>(data->hit + data->miss));
+			data->hitFraction = static_cast<float>(static_cast<double>(data->hit) / (static_cast<double>(data->hit) + data->miss));
 		}
 		else
 		{
--- a/OPHD/MicroPather/micropather.cpp
+++ b/OPHD/MicroPather/micropather.cpp
@@ -73,7 +73,7 @@ private:
 	void operator=(const OpenQueue&) = delete;
 
 	PathNode* sentinel;
-	int sentinelMem[(sizeof(PathNode) + sizeof(int)) / sizeof(int)];
+	int sentinelMem[(sizeof(PathNode) + sizeof(int)) / sizeof(int)] = {0};
 	Graph* graph;	// for debugging
 };
 
--- a/OPHD/MicroPather/micropather.cpp
+++ b/OPHD/MicroPather/micropather.cpp
@@ -939,8 +939,8 @@ int MicroPather::Solve(void* startNode, void* endNode, std::vector< void* >* pat
 				PathNode* child = nodeCostVec[i].node;
 				float newCost = node->costFromStart + nodeCostVec[i].cost;
 
-				PathNode* inOpen = child->inOpen ? child : nullptr;
-				PathNode* inClosed = child->inClosed ? child : nullptr;
+				PathNode* inOpen = (child && child->inOpen) ? child : nullptr;
+				PathNode* inClosed = (child && child->inClosed) ? child : nullptr;
 				PathNode* inEither = reinterpret_cast<PathNode*>(reinterpret_cast<MP_UPTR>(inOpen) | reinterpret_cast<MP_UPTR>(inClosed));
 
 				MPASSERT(inEither != node);
@@ -961,13 +961,16 @@ int MicroPather::Solve(void* startNode, void* endNode, std::vector< void* >* pat
 					}
 				}
 				else {
-					child->parent = node;
-					child->costFromStart = newCost;
-					child->estToGoal = graph->LeastCostEstimate(child->state, endNode),
-						child->CalcTotalCost();
+					if(child)
+					{
+						child->parent = node;
+						child->costFromStart = newCost;
+						child->estToGoal = graph->LeastCostEstimate(child->state, endNode),
+							child->CalcTotalCost();
 
-					MPASSERT(!child->inOpen && !child->inClosed);
-					open.Push(child);
+						MPASSERT(!((child->inOpen) || (child->inClosed)));
+						open.Push(child);
+					}
 				}
 			}
 		}
--- a/OPHD/MicroPather/micropather.cpp
+++ b/OPHD/MicroPather/micropather.cpp
@@ -294,15 +294,20 @@ void PathNodePool::Clear()
 
 PathNodePool::Block* PathNodePool::NewBlock()
 {
-	Block* block = reinterpret_cast<Block*>(calloc(1, sizeof(Block) + sizeof(PathNode) * (allocate - 1)));
-	block->nextBlock = nullptr;
-
-	nAvailable += allocate;
-	for (std::size_t i = 0; i < allocate; ++i)
+	// assign-then-check result is non-nullptr
+	if(Block* block = reinterpret_cast<Block*>(calloc(1, sizeof(Block) + sizeof(PathNode) * (allocate - 1))))
 	{
-		freeMemSentinel.AddBefore(&block->pathNode[i]);
+		block->nextBlock = nullptr;
+
+		nAvailable += allocate;
+		for(std::size_t i = 0; i < allocate; ++i)
+		{
+			freeMemSentinel.AddBefore(&block->pathNode[i]);
+		}
 	}
-	return block;
+	// calloc returns nullptr on error so
+	// returning nullptr if calloc fails is safe.
+	return nullptr;
 }
 
 
--- a/OPHD/MicroPather/micropather.h
+++ b/OPHD/MicroPather/micropather.h
@@ -176,7 +176,7 @@ namespace micropather
 		std::size_t numAdjacent = 0;		// -1  is unknown & needs to be queried
 		std::size_t cacheIndex = 0;			// position in cache
 
-		PathNode* child[2];			// Binary search in the hash table. [left, right]
+		PathNode* child[2] = {nullptr};			// Binary search in the hash table. [left, right]
 		PathNode* next = nullptr, * prev = nullptr;	// used by open queue
 
 		bool inOpen;


DanRStevens:
These changes look like they should be fine, though they should probably each be independent PRs, branched directly off of `master`.

DanRStevens:
I was looking back at this. Most of these changes involve the pather code. If changes can be limited to just that, it may be fine to combine them into one PR.

That means stripping off the commit for the `Tile` `noexcept` change (#372). Additionally, the change to `Slider` should be stripped off as well. It appears the change to slider is the only part that depends on the base branch changes involving `enum` changes. Hence, with that part removed from this branch, the remaining pather changes would no longer depend on the base branch changes. It should be updated to remove that dependency.

DanRStevens:
Rebased changes so they are no longer dependent on the #319 base branch. Rebase also strips out changes that were peeled off into PRs #372, #376, and were already merged to `master`.

Remaining changes affect only the pather code.

cugone:
I just tried to approve this, totally forgetting this was my own PR. >.>

That's what I get for doing 4 or 5 reviews in a row...

OPHD/MicroPather/micropather.cpp
```diff
@@ -294,15 +294,20 @@ void PathNodePool::Clear()
 
 PathNodePool::Block* PathNodePool::NewBlock()
 {
-	Block* block = reinterpret_cast<Block*>(calloc(1, sizeof(Block) + sizeof(PathNode) * (allocate - 1)));
-	block->nextBlock = nullptr;
-
-	nAvailable += allocate;
-	for (std::size_t i = 0; i < allocate; ++i)
+	// assign-then-check result is non-nullptr
+	if(Block* block = reinterpret_cast<Block*>(calloc(1, sizeof(Block) + sizeof(PathNode) * (allocate - 1))))
```
DanRStevens, line 298: I think I would prefer the assignment be left outside of the `if`. Stylistically, I would like to avoid assignments within `if` expressions. Sometimes it makes sense to do, but I don't think it's needed here.

Leaving it outside would probably also make the diff look a bit cleaner.

OPHD/MicroPather/micropather.cpp
```diff
@@ -939,8 +939,8 @@ int MicroPather::Solve(void* startNode, void* endNode, std::vector< void* >* pat
 				PathNode* child = nodeCostVec[i].node;
 				float newCost = node->costFromStart + nodeCostVec[i].cost;
 
-				PathNode* inOpen = child->inOpen ? child : nullptr;
-				PathNode* inClosed = child->inClosed ? child : nullptr;
+				PathNode* inOpen = (child && child->inOpen) ? child : nullptr;
+				PathNode* inClosed = (child && child->inClosed) ? child : nullptr;
```
DanRStevens, line 943: I wonder if we should `throw`, rather than add a guard here. Technically the guard changes behaviour. Code that likely would have hard crashed before, may now continue to run with silently incorrect behaviour.

OPHD/MicroPather/micropather.cpp
```diff
@@ -850,7 +850,7 @@ void MicroPather::GetCacheData(CacheData* data)
 		data->miss = pathCache->miss;
 		if (data->hit + data->miss)
 		{
-			data->hitFraction = static_cast<float>(static_cast<double>(data->hit) / static_cast<double>(data->hit + data->miss));
+			data->hitFraction = static_cast<float>(static_cast<double>(data->hit) / (static_cast<double>(data->hit) + data->miss));
```
DanRStevens, line 853: This doesn't appear to make a whole lot of sense. Both `hit` and `miss` are declared as `int`. It seems odd to explicitly cast one of them to `double`, and then rely on the other to be implicitly converted to `double`.

Additionally, the expression `data->hit + data->miss` is re-used in the `if` condition. Maybe we should create an `auto total` temporary variable to hold the result. I believe casting the temporary to `double` would eliminate the warning, and perhaps make more sense. Granted, that is still doing what the warning was about, naming doing calculations on the narrower size, and then casting.

The other thought is the `if` condition is checking for a condition that would lead to divide by 0. We should perhaps be using the same type for the `if` check as is actually used in the calculation, otherwise we still risk division by 0 due to conversion issues.

It might also make more sense to just use `float` throughout, rather than cast between `double` and `float`.

OPHD/MicroPather/micropather.cpp
cugone: I disagree. This is a well-known C++ idiom as well as it keeps the scope of `block` as small as possible to let the compiler optimizer do its thing. Some tools will even warn `block variable scope can be reduced`. Objectively all a good thing.

OPHD/MicroPather/micropather.cpp
cugone: If you look further down, the code already asserts when `child` is `nullptr`:

```cpp
MPASSERT(!(inOpen && inClosed));
```

Prior to this change, if child is `nullptr`, so will `inOpen and `inClosed` and as such, the MPASSERT check will pass and throw an assertion.

OPHD/MicroPather/micropather.cpp
DanRStevens: Yes, though some compilers will also warn about assignment in `if` statements. They are very easily misread, and in many cases done by accident. This case seems a little better, since the variable is being declared, not simply assigned.

As for block scope optimizations, can you name any that would apply? I'm not aware of any optimizations that would kick in for this particular case.

cugone: They don't warn for pointers.

`if(ptr)` never warns.

DanRStevens: It does warn for `if (ptr = expression)`.

cugone: Yes, but that's not what `if(type* ptr = expression)` is. It's a declaration, not an assignment.

OPHD/MicroPather/micropather.cpp
DanRStevens: That's hardly a valid check though. The `inOpen` and `inClosed` checked are derived from `child`, and they do so by dereferencing it. It `child` was `nullptr`, the program would have already crashed.

Maybe we should consider adding:
```cpp
MPASSERT(child);
```

It can be placed before the block that dereferences `child` to get the `inOpen` and `inClosed` values.

Once the assert is passed, we would no longer need to check individual expressions.

----

The check for both `inOpen` and `inClosed` is separate, and still potentially needed. It's possible a node could somehow end up being classified as being in both the open and closed sets simultaneously. Some algorithms are legitimated designed to work that way. This doesn't appear to be one of those algorithms.

OPHD/MicroPather/micropather.cpp
```diff
@@ -848,9 +848,10 @@ void MicroPather::GetCacheData(CacheData* data)
 
 		data->hit = pathCache->hit;
 		data->miss = pathCache->miss;
-		if (data->hit + data->miss)
+		const auto hitAsFloat = static_cast<float>(data->hit);
+		if (const float total = hitAsFloat + data->miss)
```
DanRStevens, line 852: This still relies on implicit conversion of one operand. I find this style very unintuitive. I think it makes more sense to either case both operands, or do the addition first, and then cast the result.

----

Second thought, if we use `float` rather than double, there is no increase in size to warn about.

----

Consider:
```cpp
auto total = data->hit + data->miss;
if (total)
{
  data->hitFraction = static_cast<float>(data->hit) / static_cast<float>(total);
}
```

In regards to testing an `int` in the `if` while using a `float` in the calculation, now that I think about it, it should be impossible for a non-zero integer to convert to a floating point 0. In the upper ranges, it may lose precision, but it would be a less precise very large number, not zero. There would be no overflow or wraparound during the conversion, only rounding to less precise large values. That would happen at about 2^24.

PR #435, opened 2020-06-25
Allow loading art assets independent of environment variable settings

Brett208:
When using Visual Studio to debug, the environment path stays with the project file instead of using the executable being debugged.

I think this makes it possible (easier) to debug from within Visual Studio.

### DIFF
--- a/OPHD/main.cpp
+++ b/OPHD/main.cpp
@@ -70,7 +70,8 @@ int main(int /*argc*/, char *argv[])
 	try
 	{
 		Filesystem& f = Utility<Filesystem>::init<Filesystem>(argv[0], "OutpostHD", "LairWorks");
-		f.mount("data");
+		// Use executable's absolute directory in case environment path is not pointing to expected directory
+		f.mount(f.basePath() + "data");
 		f.mountReadWrite(f.prefPath());
 
 		if (!f.exists(constants::SAVE_GAME_PATH))
--- a/OPHD/main.cpp
+++ b/OPHD/main.cpp
@@ -69,7 +69,7 @@ int main(int /*argc*/, char *argv[])
 
 	try
 	{
-		Filesystem& f = Utility<Filesystem>::init<Filesystem>(argv[0], "OutpostHD", "LairWorks");
+		auto& f = Utility<Filesystem>::init<Filesystem>(argv[0], "OutpostHD", "LairWorks");
 		// Prioritize data from working directory, fallback on data from executable path
 		f.mountSoftFail("data");
 		f.mountSoftFail(f.basePath() + "data");
--- a/OPHD/main.cpp
+++ b/OPHD/main.cpp
@@ -69,15 +69,15 @@ int main(int /*argc*/, char *argv[])
 
 	try
 	{
-		auto& f = Utility<Filesystem>::init<Filesystem>(argv[0], "OutpostHD", "LairWorks");
+		auto& fs = Utility<Filesystem>::init<Filesystem>(argv[0], "OutpostHD", "LairWorks");
 		// Prioritize data from working directory, fallback on data from executable path
-		f.mountSoftFail("data");
-		f.mountSoftFail(f.basePath() + "data");
-		f.mountReadWrite(f.prefPath());
+		fs.mountSoftFail("data");
+		fs.mountSoftFail(fs.basePath() + "data");
+		fs.mountReadWrite(fs.prefPath());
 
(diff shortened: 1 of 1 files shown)


OPHD/main.cpp
```diff
@@ -70,7 +70,8 @@ int main(int /*argc*/, char *argv[])
 	try
 	{
 		Filesystem& f = Utility<Filesystem>::init<Filesystem>(argv[0], "OutpostHD", "LairWorks");
-		f.mount("data");
+		// Use executable's absolute directory in case environment path is not pointing to expected directory
+		f.mount(f.basePath() + "data");
```
DanRStevens, line 74: I made this change earlier, and it was undone by Leeor in:
https://github.com/OutpostUniverse/OPHD/commit/76014230ae0c2dd1027193030796ece38db05627

Might want to discuss with him about why. I'm not sure I entirely understand why it was reverted.

Brett208:
I read the referenced commit. @ldicker83 are you placing the data directory next to the project file so it maps properly when the debugger is set regardless of which configuration is built? I guess that makes sense, although not sure if it would break the game if you were to use a third party debugger post compilation or if there were other reasons to open the game with a manipulated environment variable. No issues if you want this PR closed without merging.

DanRStevens:
Actually, I think the problem might have been a hard failure if the mount path didn't exist. It used to be a soft failure, but that was changed when filesystem code was updated and we switched to exceptions instead of logging. My guess is Leeor has a custom layout for the assets, perhaps closer to final packaging.

Since that change was made, a new method `mountSoftFail` has been introduced. It was used to avoid startup failures for the graphics test project. We should probably be using that for OPHD.

Reference:
https://github.com/lairworks/nas2d-core/pull/621

The startup code looked like:
```
auto& fs = Utility<Filesystem>::init<Filesystem>(argv_0, appName, organizationName);
fs.mountSoftFail(dataPath);
fs.mountSoftFail(fs.basePath() + dataPath);
fs.mountReadWrite(fs.prefPath());
```

That way data assets can be stored relative to the current working directory or relative to the EXE.

DanRStevens:
I implemented the change I suggested and tested that it works on Linux. Maybe someone can test that it works properly on Windows.

For reference, my `data/` folder is stored relative to the root of the repository, which matches both the current working directory and executable directory when building with the `makefile`. Not sure how Windows developers are typically setup for data folder placement. I guess the data zip isn't part of the repository, so people might place it in different locations.

Brett208:
Good idea. I'll try testing tomorrow on Windows.

Previously for Visual Studio, you must place the data folder in the same folder containing the solution file. I was placing it in the compiled folder (x64, x86, etc).

ldicker83:
This is the layout in my development folder:

image

'data' is in the same folder as the solution file.

In the debugger properties the working path is set to the solution's directory -- the idea behind this is to avoid having to copy the data folder to all of the debug/release folders where the binaries are being built to. E.g., while the binary file is in `ophd/ophd/debug` the working directory is set to `ophd/` -- only one copy of `./data/` needed.

I also know that I'm weird and do things differently than other developers do so I'm still open to suggestion here in the best way to handle it. I haven't checked out this branch since I'm still working on a different branch (and I know I can stash changes and switch back and forth but ... ugh, I just get confused easily when it comes to branches and working trees, etc.)

Brett208:
@ldicker83, what you are doing makes a lot of sense once I understood.

I tested these changes on my machine and everything worked.
- data in root solution directory worked
- data next to exectuable worked when loading in VS with debugger attached
- data next to executable worked when opening the executable without debugger attached
- moving the data out of these locations, Outpost 2 didn't fail for not finding the directory, but did fail when it couldn't find an art asset before the main menu was rendered.

So, I think this looks good. It is a lot more flexible and should allow both placing the data next to the executable or at the environment path for debugging or other operations.

I would recommend merging.

-Brett

DanRStevens:
Ok, thank you for checking.

----

Thank you Leeor for details on the layout you're using. That matches the data layout I use on Linux. The output binary folder is different from Linux, but looks compatible with the proposed settings.

From the image I noticed "Working Directory" is set to `..\`. I assume that is relative to the project file? (As opposed to relative to the solution file, or relative to the executable file). If so, that makes sense, since `..\` relative to the project file should equal the solution folder.

ldicker83:
It is relative to the project file (took a moment to remember that). Going to pull this change set into the routing branch and see how it behaves.

PR #441, opened 2020-06-26
Show Residence Capacity in Structure Info Window

Brett208:
What are thoughts on how to show all the small bits of information for each structure that doesn't require specialized UI?

I took a stab at doing it procedurally here. Wasn't sure how to handle the lamda expression that I copied from the main function. If procedural seems ok, we should probably clean that up before merging.

It could also be solved with inheritance by making many structures as child classes. Maybe there is a way to give a view pane over for each structure to populate with data. Or maybe it can just be procedural for now until a better solution is designed?

-Brett

ColonistCapacity

### DIFF
--- a/OPHD/Things/Structures/Structure.h
+++ b/OPHD/Things/Structures/Structure.h
@@ -5,6 +5,9 @@
 #include "../../Common.h"
 #include "../../PopulationPool.h"
 #include "../../ResourcePool.h"
+#include <NAS2D/Renderer/Rectangle.h>
+#include <NAS2D/Renderer/Renderer.h>
+
 
 class Structure: public Thing
 {
@@ -133,6 +136,12 @@ public:
 
 	void update() override;
 	virtual void think() {}
+
+	/**
+	* Pass limited structure specific details for drawing. Use a custom UI window if needed.
+	*/
+	#pragma warning(suppress: 4100) // Unreferenced formal parameter
+	virtual void drawInspectorView(NAS2D::Renderer& renderer, const NAS2D::Rectangle<float>& windowRect) {}
 
 protected:
 	friend class StructureCatalogue;
--- a/OPHD/Things/Structures/Structure.h
+++ b/OPHD/Things/Structures/Structure.h
@@ -140,8 +140,7 @@ public:
 	/**
 	* Pass limited structure specific details for drawing. Use a custom UI window if needed.
 	*/
-	#pragma warning(suppress: 4100) // Unreferenced formal parameter
-	virtual void drawInspectorView(NAS2D::Renderer& renderer, const NAS2D::Rectangle<float>& windowRect) {}
+	virtual void drawInspectorView([[maybe_unused]] NAS2D::Renderer& renderer, [[maybe_unused]] const NAS2D::Rectangle<float>& windowRect) {}
 
 protected:
 	friend class StructureCatalogue;
--- a/OPHD/UI/StructureInspector.cpp
+++ b/OPHD/UI/StructureInspector.cpp
@@ -101,7 +101,7 @@ void StructureInspector::drawLabelAndValue(NAS2D::Point<int> position, const std
 	renderer.drawText(*FONT_BOLD, title, position, NAS2D::Color::White);
 	position.x() += FONT_BOLD->width(title);
 	renderer.drawText(*FONT, text, position, NAS2D::Color::White);
-};
+}
 
 
 void StructureInspector::drawPopulationRequirements()
--- a/OPHD/UI/StructureInspector.cpp
+++ b/OPHD/UI/StructureInspector.cpp
@@ -6,6 +6,7 @@
 #include "../Constants.h"
 #include "../FontManager.h"
 
+#include "../Things/Structures/Residence.h"
 #include "../Things/Structures/Structure.h"
 
 #include <map>
@@ -161,7 +162,34 @@ void StructureInspector::update()
 	}
 
 	drawPopulationRequirements();
+	drawStructureTypeSpecific();
 
-	position = mRect.startPoint() + NAS2D::Vector{5, static_cast<int>(mRect.height()) - FONT->height() - 5};
+	position = mRect.startPoint() + NAS2D::Vector{ 5, static_cast<int>(mRect.height()) - FONT->height() - 5 };
 	renderer.drawText(*FONT, "This window is a work in progress", position, NAS2D::Color::White);
+}
+
+
+void StructureInspector::drawStructureTypeSpecific()
+{
+	switch (mStructure->structureClass())
+	{
+	case Structure::CLASS_RESIDENCE:
+		drawResidenceText();
+		return;
+	}
+}
+
+void StructureInspector::drawResidenceText()
+{
+	auto& renderer = Utility<Renderer>::get();
+
+	const auto drawTitleText = [&renderer](NAS2D::Point<int> position, const std::string& title, const std::string& text) {
+		renderer.drawText(*FONT_BOLD, title, position, NAS2D::Color::White);
+		position.x() += FONT_BOLD->width(title);
+		renderer.drawText(*FONT, text, position, NAS2D::Color::White);
+	};
+
+	auto position = rect().startPoint() + NAS2D::Vector{ 10, 135 };
+	drawTitleText(position, "Colonist Capacity: ", std::to_string(dynamic_cast<Residence*>(mStructure)->capacity()));
+	return;
 }
--- a/OPHD/UI/StructureInspector.cpp
+++ b/OPHD/UI/StructureInspector.cpp
@@ -94,6 +94,16 @@ void StructureInspector::btnCloseClicked()
 }
 
 
+void StructureInspector::drawTitleText(NAS2D::Point<int> position, const std::string& title, const std::string& text)
+{
+	auto& renderer = Utility<Renderer>::get();
+
+	renderer.drawText(*FONT_BOLD, title, position, NAS2D::Color::White);
+	position.x() += FONT_BOLD->width(title);
+	renderer.drawText(*FONT, text, position, NAS2D::Color::White);
+};
+
+
 void StructureInspector::drawPopulationRequirements()
 {
 	auto& renderer = Utility<Renderer>::get();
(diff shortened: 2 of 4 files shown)


DanRStevens:
I think you might be on to something with this idea.

I might suggest adding a `virtual` function to the `Thing` class hierarchy that could draw details, rather than a large `switch` block. If a new structure is added that needs custom output, it can override the details window draw method, rather than forcing an update to a centralized `switch` block. It might be too easy to miss updating the `switch` block if a new `Thing` class is added later. It also means you get default base class behaviour.

DanRStevens:
As a slight addendum, layout flexibility might be aided by designing layout container classes, which has been discussed, but so far not implemented yet. That would allow controls to be added, and they would naturally flow without having to be manually positioned. That means you could add new text below existing text more easily. Call the base class method to setup initial output, then tack on additional output for the derived class at the end. A layout container class could ensure output from both parts flows together nicely.

For simple examples of layout classes, think horizontal layout, or vertical layout. Each new control is simply stacked to the right or stacked below the previous. Adding a control to a collection automatically positions it based on layout container properties, and the position of the last added control. Makes for generating menu bars and menu windows easy. Also very helpful if items are different sizes, such as the resource panel displays at the top of the screen. The pin buttons, versus the icons, versus the text don't always make for uniformly sized panels.

Brett208:
I wouldn't worry about forgetting to update a centralized switch in this case. It would be obvious in game the structure details were missing and easy to remedy. Render layout classes would be very useful, but probably not something I would work on in the near future.

I took a stab at moving the draw code into the Structure class. It seems like we may have to repeat a lot of boilerplate code in this solution, mostly getting the fonts and defining the lambda drawLabelAndValue. I think the lambda drawLabelAndValue should be moved to NAS2D as an actual function on the renderer in any case.

Treating all the warnings as errors made development difficult. Maybe I will get more used to it. 

Also, all the function definitions in structure headers has been a bit of a challenge for me, mostly with remembering to qualify namespaces and multiplying the error count when compiling if the error is in the header.

OPHD/Things/Structures/Structure.h
```diff
@@ -134,6 +137,12 @@ class Structure: public Thing
 	void update() override;
 	virtual void think() {}
 
+	/**
+	* Pass limited structure specific details for drawing. Use a custom UI window if needed.
+	*/
+	#pragma warning(suppress: 4100) // Unreferenced formal parameter
+	virtual void drawInspectorView(NAS2D::Renderer& renderer, const NAS2D::Rectangle<float>& windowRect) {}
```
DanRStevens, line 144: Another way to avoid the warning is to comment out the parameter name:
```cpp
virtual void drawInspectorView(NAS2D::Renderer& /*renderer*/, const NAS2D::Rectangle<float>& /*windowRect*/) {}
```

Parameter names are of course optional in contexts where parameters aren't used.

----

Another possibility might be the attribute [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused), though I've never experimented with it.

OPHD/UI/StructureInspector.h
```diff
@@ -23,7 +23,10 @@ class StructureInspector : public Window
 private:
 	void btnCloseClicked();
 
+	void drawLabelAndValue(NAS2D::Point<int> position, const std::string& title, const std::string& text);
 	void drawPopulationRequirements();
+	void drawStructureTypeSpecific();
+	void drawResidenceText();
```
DanRStevens, line 29: Looks like these two methods have been removed.

Brett208: Thanks for catching.

OPHD/Things/Structures/Structure.h
Brett208: I tried maybe_unused and it seemed to work well on MSVC. I wanted to leave the argument names in place for context on what the NAS2D::Rectangle referenced.

Thanks for doing the research on how to resolve the warning/error.

Brett208:
Where is this branch from a review perspective? Is it good enough to merge or does it need more work on the solution? If it is good enough to merge, I would probably create a new branch and reapply changes to clean up the commit noise since this branch has been so exploratory.

I would like to create a DrawText helper module to collocate the helper functions. Sounds good for eventual movement into NAS2D. I would probably take a stab at that when this branch closes out. I also noticed the right justification in one of the lamdas. Maybe a DrawLabelAndValueRightJustified.

-Brett

OPHD/UI/StructureInspector.cpp
```diff
@@ -161,7 +162,34 @@ void StructureInspector::update()
 	}
 
 	drawPopulationRequirements();
+	drawStructureTypeSpecific();
 
-	position = mRect.startPoint() + NAS2D::Vector{5, static_cast<int>(mRect.height()) - FONT->height() - 5};
+	position = mRect.startPoint() + NAS2D::Vector{ 5, static_cast<int>(mRect.height()) - FONT->height() - 5 };
```
DanRStevens: I think this project tends to use tight braces pretty consistently when initializing structs. I haven't really verified though, and there hasn't been a formal discussion on the style.

DanRStevens:
Small thought. If you moved the `drawLabelAndValue` method into the `Structure` class, it would be available to all derived classes. Though I suppose currently that would mean duplicating it in both `Structure` and `StructureInspector`. (We'll probably come up with something better eventually).

Brett208:
Any unmerged changes in this branch are now contained in PR #449 

I'm going to go ahead and close this PR and branch now that the other branch exists.

ldicker83:
I haven't read the whole thread but a procedural generation of the info for the structures is what I had sort of intended. A lot of these GUI elements are unfinished and unpolished, designed to make the info available while I build the core parts of the game. I also find UI development tedious so a quick and dirty "This is good enough for now" is often how I do GUI elements. I know, not the greatest way to handle it especially since there's a lot to put in there (like why a structure is Idle, for instance). Will read through the rest of the thread and comment as needed.

ldicker83:
> As a slight addendum, layout flexibility might be aided by designing layout container classes, which has been discussed, but so far not implemented yet. 

> For simple examples of layout classes, think horizontal layout, or vertical layout. Each new control is simply stacked to the right or stacked below the previous. ... /snip

I'm definitely in favor of this as it would avoid the need for hard-coding positioning. I've been resistant to taking on this task because it seemed less important than the core game-play mechanics. That's my excuse anyway, the reality is layout engines are a distraction for me (as are skinning engines) and I know it which is why I haven't spent much time on either... I want to focus on the game being a game before it being really nice looking... if anybody is interested in taking on this task I'm game!

ldicker83:
> Treating all the warnings as errors made development difficult. Maybe I will get more used to it.

Agreed.

OPHD/UI/StructureInspector.cpp
ldicker83: Personally I prefer the spacing in the updated line, for whatever reason it's easier to read. Might be worth making that consistent.

DanRStevens:
Yeah, I think Leeor's time is better spent on game mechanics. I don't have the project vision for that part. A layout manager though, I could certainly experiments with designs for that.

----

Not really sure what specific problems are being encountered with the warnings as errors settings. Ideally the CI builds should treat warnings as errors, otherwise warnings tend to go unnoticed until you're developing on that particular platform. If there was a way to disable warnings as errors locally, while still enabling it for CI builds, that might be ideal.

Related, is the increased warning level. I kind of like having an increased warning level, and being made aware of potential issues early. Also helps prevent reintroducing problems that have been fixed. I don't always like it when those warnings get in the way of compiling and testing, though I do generally try to fix any warnings before uploading stuff.

OPHD/UI/StructureInspector.cpp
DanRStevens: I did some quick grepping on this. Searching for tight braces reveals a whole lot of data structures and classes being initialized. Grepping for loose braces reveals a whole lot of inline function definitions. There were some exceptions in each case though.


For code with tight braces, I see quite a few examples in initializer lists, such as:
```cpp
PlanetSelectState::PlanetSelectState() :
	mBg{"sys/bg1.png"},
	mStarFlare{"sys/flare_1.png"},
	mDetailFlare{"sys/flare_2.png"},
	mDetailFlare2{"sys/flare_3.png"},
	mCloud1{"sys/cloud_1.png"},
	mCloud2{"sys/cloud_2.png"},
	mBgMusic{"music/menu.ogg"},
	mSelect{"sfx/click.ogg"},
	mHover{"sfx/menu4.ogg"},
	mQuit{"Main Menu"},
	mReturnState{this}
{}
```

I think previously the code used `()` in a lot of cases, but was updated to use `{}` instead.

For reference, there's hardly any code that includes a space after an opening parenthesis. Aside from the micropather code, there are only two instances of that, one of which is a `#pragma message( ...`.


I think the particular case quoted is a bit uglier because of the calculation done inside the `{}`. Most cases are much simpler, such as:
```cpp
mPlanetDescription.size({550, 200});
```

Or:
```cpp
renderer.drawImageRotated(mCloud1, {-256, -256}, rotation, NAS2D::Color{100, 255, 0, 135});
```

Not sure if we'd want to add spaces there. I suspect extra spacing around simple cases could make them harder to read.


For an example of loose braces, they're mostly used in header files for inline function defintions:
```cpp
PlanetType type() const { return mType; }
```

ldicker83:
The warnings as errors is just that -- any time there's a warning generated by the compiler (for me it's mostly unused function parameters) it's treated as an error and the build fails. Makes it difficult to do stub functions that don't actually do anything as I flesh them out.

That stated, I do like the increased warning level, seems to have highlighted potential problems in the code and especially forces me to consider how certain implicit casts are being handled.

DanRStevens:
For the Linux builds, the `-Werror` flag is passed specially for CI builds, but is not default for local builds. Local builds would just show warnings, but still successfully build.

Perhaps something similar can be done for the Windows builds. I think that would make more sense actually. I'm just not certain how to do that with Visual Studio.

----

Possibly relevant AppVeyor documentation:
[Build-mode](https://www.appveyor.com/docs/build-configuration/#build-mode)
[Build phase](https://www.appveyor.com/docs/build-phase/)
[Adding additional parameters to msbuild with a response file](https://www.appveyor.com/docs/build-phase/#adding-additional-parameters-to-msbuild-with-a-response-file)
- msbuild.rsp

That last part is probably the most relevant.

OPHD/UI/StructureInspector.cpp
Brett208: I'm not sure it is worth regulating space or no space for this situation.

It was sloppy of me to add commit noise and not catch it. I suspect Visual Studio's default formatting made the line change when I was refactoring.

Brett208:
While it is important to reduce the warnings in the code base, in my mind they are warnings and not errors for a reason.

Too bad CI doesn't show warning count at the top level so I didn't have to open the files to see if a warning was produced. I can live with this for now though and complain later if it continues to be too big a bother.

ldicker83:
I suspect cugone changed project settings for "Treat warnings as errors". I didn't revert it only because I didn't know if there was a reason it was set like that. We can safely change it back (I'll make that change in the master branch separately from other PR's and branches.

PR #443, opened 2020-06-27
MapViewState DirectionOffsets

DanRStevens:
Use named `DirectionOffset` values in `MapViewState`.


DanRStevens:
Thanks. I have to check that one carefully to make sure I didn't miss something or get any cases backwards.

PR #449, opened 2020-06-28
Consolidate text rendering

Brett208:
First attempt at consolidating some of the text rendering. I wanted to accomplish this before starting to render more of the basic structure data.


OPHD/Things/Structures/Residence.h
```diff
@@ -29,17 +25,8 @@ class Residence : public Structure
 
 	int capacity() const { return mCapacity; }
 
-	void drawInspectorView(NAS2D::Renderer& renderer, const NAS2D::Rectangle<float>& windowRect) override
+	void drawInspectorView([[maybe_unused]] NAS2D::Renderer& renderer, const NAS2D::Rectangle<float>& windowRect) override
```
DanRStevens, line 28: The `[[maybe_unused]]` here makes me a bit apprehensive. It's theoretically possible there could be two distinct rendering surfaces (this game just isn't setup to function that way). I feel like the `Renderer` should be passed through to the text rendering functions. Either that, or we omit it here to ensure consistency.

I made a suggestion to Leeor a little while back that we start passing `Renderer` parameters around explicitly in a number of places, such as to control draw functions (and that `update` and `draw` be separate methods). My preference is to use explicit parameters over `Utility`/global variables.

OPHD/UI/TextRender.cpp
```diff
@@ -19,3 +19,15 @@ void drawLabelAndValue(NAS2D::Point<int> position, const std::string& title, con
...
+{
+	auto& renderer = NAS2D::Utility<NAS2D::Renderer>::get();
+
+	NAS2D::Font* FONT = NAS2D::Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
+	NAS2D::Font* FONT_BOLD = NAS2D::Utility<FontManager>::get().font(constants::FONT_PRIMARY_BOLD, constants::FONT_PRIMARY_NORMAL);
+
+	renderer.drawText(*FONT_BOLD, title, position, color);
+	position.x() += labelWidth;
+	renderer.drawText(*FONT, text, position, color);
+};
```
DanRStevens: Missing newline.

OPHD/UI/TextRender.cpp
```diff
@@ -19,3 +19,15 @@ void drawLabelAndValue(NAS2D::Point<int> position, const std::string& title, con
 	position.x() += FONT_BOLD->width(title);
 	renderer.drawText(*FONT, text, position, NAS2D::Color::White);
 }
+
+void drawLabelAndValueLeftJustify(NAS2D::Point<int> position, const std::string& title, const std::string& text, int labelWidth, NAS2D::Color color)
```
DanRStevens, line 23: This method name kind of made me go "huh?". Though after looking at what it does, I can't think of anything better. Names are hard.

OPHD/UI/FactoryProduction.cpp
```diff
@@ -218,26 +212,24 @@ void FactoryProduction::update()
...
-		renderer.drawText(*FONT_BOLD, title, position, NAS2D::Color::White);
-		position.x() += 120;
-		renderer.drawText(*FONT, text, position, NAS2D::Color::White);
-	};
-
+	const int labelWidth = 120;
 	auto position = mRect.startPoint() + NAS2D::Vector{constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), 25};
-	drawTitleText(position, "Turns Completed:", std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild()));
+	auto value = std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild());
+	drawLabelAndValueLeftJustify(position, "Turns Completed:", value, 120);
```
DanRStevens, line 218: Passing an explicit `120` rather than `labelWidth`.

OPHD/UI/FactoryProduction.cpp
```diff
@@ -218,26 +212,24 @@ void FactoryProduction::update()
...
 
 	position.y() += 20;
-	drawTitleText(position, "Common Metals:", std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild()));
+	value = std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild());
+	drawLabelAndValueLeftJustify(position, "Common Metals:", value, labelWidth);
```
DanRStevens, line 222: I'm not super fond of re-using the variable `value` for a bunch of different unrelated values.

I was thinking, if the parameter order of `drawLabelAndValueLeftJustify` was changed to move `labelWidth` right after `position`, it might make inline values more attractive:
```cpp
drawLabelAndValueLeftJustify(position, labelWidth, "Common Metals:", std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild()));
```

OPHD/UI/Reports/FactoryReport.cpp
```diff
@@ -590,14 +586,14 @@ void FactoryReport::drawDetailPane(Renderer& renderer)
...
-		drawTitleText(position, title, std::to_string(value), defaultTextColor);
+		drawLabelAndValueLeftJustify(position, title, std::to_string(value), labelWidth, defaultTextColor);
 		position.y() += 15;
 	}
 
 	// POPULATION
 	bool isPopulationRequirementHighlighted = SELECTED_FACTORY->populationAvailable()[0] != SELECTED_FACTORY->populationRequirements()[0];
 	auto text = std::to_string(SELECTED_FACTORY->populationAvailable()[0]) + " / " + std::to_string(SELECTED_FACTORY->populationRequirements()[0]);
-	drawTitleText(position, "Workers", text, (isPopulationRequirementHighlighted ? NAS2D::Color::Red : defaultTextColor));
+	drawLabelAndValueLeftJustify(position, "Workers", text, labelWidth, (isPopulationRequirementHighlighted ? NAS2D::Color::Red : defaultTextColor));
```
DanRStevens, line 596: Noting text justification was added here, so the change may need to be visually inspected.

OPHD/UI/TextRender.h
```diff
@@ -5,5 +5,5 @@
 #include <string>
 
 
-void drawLabelAndValue(NAS2D::Point<int> position, const std::string& title, const std::string& text, NAS2D::Color color = NAS2D::Color::White);
+void drawLabelAndValue(NAS2D::Point<int> position, const std::string& title, const std::string& text, int valuePadding = 0, NAS2D::Color color = NAS2D::Color::White);
```
DanRStevens, line 8: While less accurate, I wonder if this should be handled by adding spaces at the end of the `title` string. It's starting to seem like the method is becoming burdened with too many parameters.

OPHD/UI/TextRender.cpp
```diff
@@ -30,4 +30,4 @@ void drawLabelAndValueLeftJustify(NAS2D::Point<int> position, const std::string&
 	renderer.drawText(*FONT_BOLD, title, position, color);
 	position.x() += labelWidth;
 	renderer.drawText(*FONT, text, position, color);
-};
\ No newline at end of file
+}
```
DanRStevens, line 33: Missing newline.

OPHD/Things/Structures/Residence.h
Brett208: I'd like to just pull renderer out of the drawInspectorView for now. I was thinking this as the right course after pushing the PR.

Generally agree best practice is to keep Update and Draw logic separate. Certainly could be uses for multiple renderers and I have typically passed them as suggested.

OPHD/UI/FactoryProduction.cpp
Brett208: I like the idea of moving labelWidth right after position and will implement.

The purpose of reusing value was to reduce the length of the calls to drawLabelAndValue and (hopefully) make the code easier to read. Some of the lines were getting pretty long concatenating the strings within the function. We could change to using individual variables names instead of reusing value or just accept the line length will be long for these function calls in their current format. Certainly open to future refactoring of the RenderText functions to reduce their argument count.

OPHD/UI/FactoryProduction.cpp
Brett208: Thanks for catching my sloppiness here

OPHD/UI/TextRender.h
Brett208: I put valuePadding here to allow for the current use cases in the code more than an endorsement of the idea. In an eventual string table approach, this would be replaced by a columnPadding property that would improve the multiple arguments in this function.

Looking at this closer, it doesn't seem to make sense why the padding exists. I think it would look better left justified. Maybe some judgement here, but I'll change to left justify to remove the padding use case and avoid the extra space solution.

OPHD/UI/Reports/FactoryReport.cpp
Brett208: Actually, looking at this closer, the original solution was right justified labels, not left justified labels, so I messed up a lot of the calls in the FactoryReport. I'll look at it later today and create a new drawLabelAndValueRightJustify.

Thanks for pointing out.

Brett208:
@DanRStevens, thanks for the very thorough review of the changes. 

I ran out of time to fix all, but should be able to work tomorrow. 

I would propose after removing the renderer from drawInspectorView and adding the drawLabelAndValueRightJustify, we merge this branch and open an issue for adding a StringTable class.

-Brett

OPHD/UI/FactoryProduction.cpp
```diff
@@ -215,7 +215,7 @@ void FactoryProduction::update()
 	const int labelWidth = 120;
```
ldicker83, line 215: This could/should be done as a constant in Constants/Numbers.h or Constants/Ui.h -- perhaps make it a static local?

I'm a big fan of named constants being in ALL_CAPS to make it very obvious what they are -- though a namespace like constants:: also works.

I'm open to suggestion here.

DanRStevens: I would advise against using ALL_CAPS, as that is typically reserved for macros. Macros do not respect namespaces, so you really don't want to ever have a collision between a macro and a standard variable or constant. There is no way to protect a macro name using namespaces, they are globally replaced regardless of namespacing. This has been a problem on a few occasions previously for OPHD/NAS2D.

The typical naming conventions I'm used to for `const` variables is either normal lowerCamelCase like regular variables, or UpperCamelCase, much like classes. I think I've used UpperCamelCase a few times in this project, though I don't there's been a consistently followed convention here.

----

In general, I'm not so fond of separating constants into a separate constants file. I would prefer constants affecting a certain component be defined within that component. Makes the code more standalone, and requires less searching around to find things. Maybe it's a little easier with Visual Studio and the code navigation capabilities, though I typically don't have those features available in the editors I use.

In this case, if the value is used repeatedly across multiple UI components, it may make sense to put it in a constants file. More longer term, I kind of wonder if we might come up with another way of specifying UI constants affecting the visual layout style. Perhaps something a bit more dynamic that could support skinning. That's by no means an immediate concern though, and I'd suggest sticking to constants for the time being.

I could go either way on the placement for this variable.

----

For constants initialized with literal values, I sometimes like to mark them as `constexpr`.

In the context of literal values, there should be no real difference between `const` and `constexpr`. Either way, if you never take the address of a the variable it might actually have storage space optimized away. I'm not sure if `static const` of a local would do that.

I generally use `const` for computed constants that depend on some kind of runtime parameter, such as the return value from a non-constexpr function. I try to use `constexpr` whenever the more strict rules can apply.

DanRStevens:
I would like to defer on adding a `StringTable` class for now (or a `Table` class, which might contain other things in the cells, such as `Control` instances). The idea isn't super well formed, and may change depending on how the rest of the code shapes up. I think after more cleanup we'll have a better idea of if we even need that, or if another solution might be better. Related, the `ListBox` controls (there are several specializations) have a fair bit of duplication, and addressing that may impact the `Table` idea.

Brett208:
@ldicker83,

I'm not quite following your thoughts about the WarehouseInspector and string literals. The way the code is currently written also doesn't really bother me though.

There is sort of a back and forth about the labelWidth. I suspect this will change when some sort of table materializes, so I'd be fine leaving as is and merging for now.

If this branch looks good, please merge. If preferred, I can rework this branch to be less messy as it ended up going back and forth on some of the commits.

Thanks for the thorough reviews.

-Brett

OPHD/UI/FactoryProduction.cpp
ldicker83: Like I said, I'm open to suggestion -- it's been my convention to use all caps but I'm not dead set on it. Agreed that it's not consistent across the project as that hasn't been something I've enforced.

I would think UpperCamelCase makes sense here as it looks/feels like a proper name which is how I view constants.

The option to put them in a namespace and in a singular place is to make it easy to find where they are. This is really mostly for constants that are used around the code, namely string literals and values that make sense to keep in one spot for tweaking (e.g., simulation parameters, UI literals, etc). They all do follow the ALL_CAPS_WITH_UNDERSCORE convention but I'm not opposed to changing this to UpperCamelCase -- should be in a different change-set me thinks.

Anyway, I digress -- I don't have a whole lot more to comment here other than things are looking good and the attention to this is much appreciated!

ldicker83:
Looks good to me, I'll be merging this soon.

As for what I was mentioning, it's a different changeset and is the best explanation I could muster with the worst headache I've had all year (lasted three days now, I can't focus on anything).

DanRStevens:
`FactoryProduction` and `FactoryReport` seem to have their left/right justification swapped.

Originally `FactoryProduction` was left justified, while `FactoryReport` was right justified.

I'm merely noting this without looking at the interface, nor making any judgement on what would be better where.

ldicker83:
We can adjust it as we go forward if needed. I didn't notice that but I'm not too worried about it atm.

OPHD/UI/WarehouseInspector.cpp
```diff
@@ -84,25 +78,19 @@ void WarehouseInspector::update()
...
-	const auto drawTitleText = [&renderer](NAS2D::Point<int> position, const std::string& title, const std::string& text, int offset = 100) {
-		renderer.drawText(*FONT_BOLD, title, position, NAS2D::Color::White);
-		position.x() += offset ? offset : (FONT_BOLD->width(title) + 20);
-		renderer.drawText(*FONT, text, position, NAS2D::Color::White);
-	};
+	const int labelWidth = 100;
 
 	auto position = mRect.startPoint() + NAS2D::Vector{constants::MARGIN, 25};
-	drawTitleText(position, "Storage", std::to_string(pool.availableStorage()) + " / " + std::to_string(pool.capacity()), 0);
+	drawLabelAndValueLeftJustify(position, labelWidth, "Storage:", std::to_string(pool.availableStorage()) + " / " + std::to_string(pool.capacity()));
```
DanRStevens, line 86: Small note that this changes the positioning of the value and adds a ":" to the title. Probably a good change. Just noting it for visual inspection.

Brett208: That was on purpose on my part. Could have been a separate commit for clarity.

PR #455, opened 2020-06-30
Refactor tile highlight

DanRStevens:
Represent highlight position internally as an absolute coordinate `Point`, rather than an offset relative to `mMapViewLocation`.

### DIFF
--- a/OPHD/Map/TileMap.h
+++ b/OPHD/Map/TileMap.h
@@ -49,7 +49,6 @@ public:
 	void mapViewLocation(NAS2D::Point<int> point);
 	void centerMapOnTile(Tile*);
 
-	const NAS2D::Point<int>& tileHighlight() const { return mMapHighlight; }
 	bool tileHighlightVisible() const;
 
 	int tileMouseHoverX() const { return mMapHighlight.x() + mMapViewLocation.x(); }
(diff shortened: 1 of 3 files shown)


OPHD/Map/TileMap.h
```diff
@@ -137,7 +137,7 @@ class TileMap: public micropather::Graph
 	NAS2D::Timer mTimer;
 
 	NAS2D::Point<int> mMousePosition; /**< Current mouse position. */
-	NAS2D::Point<int> mMapHighlight; /**< Tile the mouse is pointing to. */
+	NAS2D::Vector<int> mMapHighlight; /**< Tile the mouse is pointing to. */
```
ldicker83: Curious why the change here was made? It would seem a Point is a more natural representation of a specific tile?

OPHD/Map/TileMap.h
```diff
@@ -134,7 +134,7 @@ class TileMap: public micropather::Graph
 	NAS2D::Timer mTimer;
 
 	NAS2D::Point<int> mMousePosition; /**< Current mouse position. */
-	NAS2D::Vector<int> mMapHighlight; /**< Tile the mouse is pointing to. */
+	NAS2D::Point<int> mMapHighlight; /**< Tile the mouse is pointing to. */
```
ldicker83: Seems this has been addressed. Carry on! :D

PR #462, opened 2020-07-02
TileMap random mine placement refactor

DanRStevens:
Moves random number objects from global to function local. These are used once per map generation.


DanRStevens:
@ldicker83, you might find this one interesting. It contains some of the changes were discussed.

PR #466, opened 2020-07-04
String table

### LINKED ISSUE #458: StringTable
Looking to add a StringTable that would allow organizing text into a table format for render. Below is prototype sketch for initial commit. Could be expanded on for more features in future.

This would be the time to introduce a third party library instead of home-rolling UIs...

EDIT: Fixed some syntax issues in code below and standardized function name casing to match program style
EDIT2: Added const to appropriate functions and used size_t for index variables and arguments

```C++
#pragma once

#include <NAS2D/Renderer/Renderer.h>
#include <NAS2D/Renderer/Color.h>
#include <NAS2D/Renderer/Point.h>
#include <string>
#include <vector>
#include <cstddef>

// Draw a 2 dimensional table of text. Determine cell size based on inserted text, font, and padding. Only allows one line of text per cell.
class StringTable
{
public:
    using CellCoordinate = NAS2D::Point<std::size_t>;

    enum class Justification
    {
        Left,
        Right
    };

    // Set default fonts in constructor
    StringTable(std::size_t columns, std::size_t rows);

    void draw(NAS2D::Renderer& renderer) const;

    // Call after updating table properties to recompute cell positions
    void computeRelativeCellPositions();

    void SetPosition(NAS2D::Point<float> position);
    NAS2D::Point<float> GetPosition() const;

    void setDefaultFont(NAS2D::Font* font);
    void setDefaultTitleFont(NAS2D::Font* font);
    void setDefaultTextColor(NAS2D::Color textColor);

    void setHorizontalPadding(float horizontalPadding);
    void setVerticalPadding(float verticalPadding);

    void setCellText(const CellCoordinate& cellCoordinate, std::string text);
    // Override default font settings
    void setCellFont(const CellCoordinate& cellCoordinate, NAS2D::Font* font);
    void setCellJustification(const CellCoordinate& cellCoordinate, Justification justification);
    // Override default Color setting
    void setCellTextColor(const CellCoordinate& cellCoordinate, NAS2D::Color textColor);

private:
    struct Cell
    {
        // Set textColor to ColorEmpty to indicate cell should use default StringTable color
        static const NAS2D::Color ColorEmpty;

        // Use defaultFont if not set
        NAS2D::Font* font = nullptr;
        std::string text;
        Justification justification = Justification::Left;
        NAS2D::Color textColor = ColorEmpty;
        // Position relative to the StringTable's position
        NAS2D::Point<float> textRelativePosition;
    };

    std::vector<Cell> cells;
    const std::size_t columns;
    const std::size_t rows;
    NAS2D::Point<float> position;
    NAS2D::Font* defaultFont;
    NAS2D::Font* defaultTitleFont;
    NAS2D::Color defaultTextColor = NAS2D::Color::White;
    float horizontalPadding = 0;
    float verticalPadding = 0;

    std::size_t getCellIndex(const CellCoordinate& cellCoordinate) const;
    CellCoordinate getCellCoordinate(std::size_t index) const;
    void checkCellIndex(const CellCoordinate& cellCoordinate) const;

    NAS2D::Font* getCellFont(std::size_t index) const;

    std::vector<float> computeColumnWidths();
    void computeCellPosition(std::size_t index, float columnWidth);
    void accountForCellJustification(std::size_t index, float columnWidth);
    bool isFirstColumn(std::size_t index) const;
};
```

Brett208:
Looking for feedback if I'm close enough to the mark on code style in addition to the logic of StringTable.

Included FactoryProduction window as a sample for how StringTable would be used. 

The idea of StringTable is that you can initialize all the data and then just call the Draw function many times. This would reduce CPU time recalculating the layout of each cell. The underlying game framework tends to call update and draw code from same function, so at this state doesn't buy anything.

Actual dimensions are not calculated until the function computeRelativeCellPositions is called. This allows for manipulation of the data without forcing recalculation of the dimensions on every call to a function in StringTable that could change layout. This call could be moved inside the draw function to encapsulate the logic, but may make a performance hit long term. Since there isn't currently a performance problem, I would be completely fine encapsulating it this way.

I standardized to (row, column) because that seems to match up with (x,y) type logic on a graph. However, intuitively, it might be better to insert data in (column, row) format since I tend to fill one row at a time. 

The cells are stored in a 1D array. There may be some sort of bit twiddling or better math to optimize some of the calls? Maybe best to leave alone unless performance is needed though if it harms readability.

Closes #458 

FactoryProductionSample
Screenshot showing I think it looks close to the same as before

---

PS: How do I access the FactoryReport UI from in game? I'm having trouble finding it for some reason...

### DIFF
--- a/OPHD/UI/FactoryProduction.cpp
+++ b/OPHD/UI/FactoryProduction.cpp
@@ -213,7 +213,7 @@ void FactoryProduction::update()
 	Window::update();
 
 	StringTable stringTable(2, 5);
-	stringTable.setPosition({ mRect.x() + constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), mRect.y() + 25 });
+	stringTable.setPosition(mRect.startPoint() + NAS2D::Vector<float>{ constants::MARGIN * 2 + mProductGrid.width(), 25 });
 	stringTable.setColumnJustification(1, StringTable::Justification::Right);
 
 	stringTable.setCellText(0, 0, "Turns Completed:");
--- a/OPHD/UI/FactoryProduction.cpp
+++ b/OPHD/UI/FactoryProduction.cpp
@@ -213,7 +213,7 @@ void FactoryProduction::update()
 	Window::update();
 
 	StringTable stringTable(2, 5);
-	stringTable.setPosition(mRect.startPoint() + NAS2D::Vector{constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), 25});
+	stringTable.setPosition({ mRect.x() + constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), mRect.y() + 25 });
 	stringTable.setColumnJustification(1, StringTable::Justification::Right);
 
 	stringTable.setCellText(0, 0, "Turns Completed:");
--- a/OPHD/UI/FactoryProduction.cpp
+++ b/OPHD/UI/FactoryProduction.cpp
@@ -3,7 +3,7 @@
 
 #include "FactoryProduction.h"
 
-#include "TextRender.h"
+#include "StringTable.h"
 #include "../Constants.h"
 
 using namespace NAS2D;
--- a/OPHD/UI/FactoryProduction.cpp
+++ b/OPHD/UI/FactoryProduction.cpp
@@ -212,19 +212,25 @@ void FactoryProduction::update()
 
 	Window::update();
 
-	const int labelWidth = 120;
-	auto position = mRect.startPoint() + NAS2D::Vector{constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), 25};
-	drawLabelAndValueRightJustify(position, labelWidth, "Turns Completed:", std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild()));
+	StringTable stringTable(2, 5);
+	stringTable.setPosition(mRect.startPoint() + NAS2D::Vector{constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), 25});
+	stringTable.setColumnJustification(1, StringTable::Justification::Right);
 
-	position.y() += 20;
-	drawLabelAndValueRightJustify(position, labelWidth, "Common Metals:", std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild()));
+	stringTable.setCellText(0, 0, "Turns Completed:");
+	stringTable.setCellText(1, 0, std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild()));
 
-	position.y() += 10;
-	drawLabelAndValueRightJustify(position, labelWidth, "Common Minerals:", std::to_string(mProductCost.commonMinerals() * mProductCost.turnsToBuild()));
+	stringTable.setCellText(0, 1, "Common Metals:");
+	stringTable.setCellText(1, 1, std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild()));
 
-	position.y() += 10;
-	drawLabelAndValueRightJustify(position, labelWidth, "Rare Metals:", std::to_string(mProductCost.rareMetals() * mProductCost.turnsToBuild()));
+	stringTable.setCellText(0, 2, "Common Minerals:");
+	stringTable.setCellText(1, 2, std::to_string(mProductCost.commonMinerals() * mProductCost.turnsToBuild()));
 
-	position.y() += 10;
-	drawLabelAndValueRightJustify(position, labelWidth, "Rare Minerals:", std::to_string(mProductCost.rareMinerals() * mProductCost.turnsToBuild()));
+	stringTable.setCellText(0, 3, "Rare Metals:");
+	stringTable.setCellText(1, 3, std::to_string(mProductCost.rareMetals() * mProductCost.turnsToBuild()));
+
+	stringTable.setCellText(0, 4, "Rare Minerals:");
+	stringTable.setCellText(1, 4, std::to_string(mProductCost.rareMinerals() * mProductCost.turnsToBuild()));
+
+	stringTable.computeRelativeCellPositions();
+	stringTable.draw(Utility<Renderer>::get());
 }
--- a/OPHD/UI/FactoryProduction.cpp
+++ b/OPHD/UI/FactoryProduction.cpp
@@ -216,20 +216,20 @@ void FactoryProduction::update()
 	stringTable.setPosition(mRect.startPoint() + NAS2D::Vector<float>{ constants::MARGIN * 2 + mProductGrid.width(), 25 });
 	stringTable.setColumnJustification(1, StringTable::Justification::Right);
 
-	stringTable.setCellText(0, 0, "Turns Completed:");
-	stringTable.setCellText(1, 0, std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild()));
+	stringTable.at(0, 0).text = "Turns Completed:";
+	stringTable.at(1, 0).text = std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild());
 
-	stringTable.setCellText(0, 1, "Common Metals:");
-	stringTable.setCellText(1, 1, std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild()));
+	stringTable.at(0, 1).text = "Common Metals:";
+	stringTable.at(1, 1).text = std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild());
 
-	stringTable.setCellText(0, 2, "Common Minerals:");
-	stringTable.setCellText(1, 2, std::to_string(mProductCost.commonMinerals() * mProductCost.turnsToBuild()));
+	stringTable.at(0, 2).text = "Common Minerals:";
+	stringTable.at(1, 2).text = std::to_string(mProductCost.commonMinerals() * mProductCost.turnsToBuild());
 
-	stringTable.setCellText(0, 3, "Rare Metals:");
-	stringTable.setCellText(1, 3, std::to_string(mProductCost.rareMetals() * mProductCost.turnsToBuild()));
+	stringTable.at(0, 3).text = "Rare Metals:";
+	stringTable.at(1, 3).text = std::to_string(mProductCost.rareMetals() * mProductCost.turnsToBuild());
 
-	stringTable.setCellText(0, 4, "Rare Minerals:");
-	stringTable.setCellText(1, 4, std::to_string(mProductCost.rareMinerals() * mProductCost.turnsToBuild()));
+	stringTable.at(0, 4).text = "Rare Minerals:";
+	stringTable.at(1, 4).text = std::to_string(mProductCost.rareMinerals() * mProductCost.turnsToBuild());
 
 	stringTable.computeRelativeCellPositions();
 	stringTable.draw(Utility<Renderer>::get());
--- a/OPHD/UI/StringTable.cpp
+++ b/OPHD/UI/StringTable.cpp
@@ -89,7 +89,7 @@ void StringTable::setCellJustification(const CellCoordinate& cellCoordinate, Jus
 
 void StringTable::setColumnJustification(std::size_t column, Justification justification)
 {
-	for (int row = 0; row < rows; ++row)
+	for (std::size_t row = 0; row < rows; ++row)
 	{
 		setCellJustification(CellCoordinate(column, row), justification);
 	}
--- a/OPHD/UI/StringTable.cpp
+++ b/OPHD/UI/StringTable.cpp
@@ -216,7 +216,7 @@ void StringTable::checkCellIndex(const CellCoordinate& cellCoordinate) const
 
 NAS2D::Font* StringTable::getCellFont(std::size_t index) const
 {
-	NAS2D::Font* font = cells[index].font;// = index % columns == 0 ? defaultTitleFont : defaultFont;
+	NAS2D::Font* font = cells[index].font;
 
 	if (font == nullptr) {
 		font = isFirstColumn(index) ? defaultTitleFont : defaultFont;
--- a/OPHD/UI/StringTable.cpp
+++ b/OPHD/UI/StringTable.cpp
@@ -21,18 +21,18 @@ void StringTable::draw(NAS2D::Renderer& renderer) const
 		const Cell& cell = cells.at(i);
 
 		NAS2D::Color textColor = cell.textColor != Cell::ColorEmpty ? cell.textColor : defaultTextColor;
-		NAS2D::Point<float> drawPosition = { position.x() + cell.textRelativePosition.x(), position.y() + cell.textRelativePosition.y() };
+		const auto drawPosition = position + cell.textRelativePosition;
 
-		renderer.drawText(*getCellFont(i), cell.text, drawPosition, textColor);
+		renderer.drawText(*getCellFont(i), cell.text, NAS2D::Point<float>(drawPosition.x, drawPosition.y), textColor);
 	}
 }
 
-void StringTable::setPosition(NAS2D::Point<float> tablePosition)
+void StringTable::setPosition(NAS2D::Vector<float> tablePosition)
 {
 	this->position = tablePosition;
 }
 
-NAS2D::Point<float> StringTable::getPosition() const
+NAS2D::Vector<float> StringTable::getPosition() const
 {
 	return position;
 }
@@ -137,7 +137,7 @@ void StringTable::accountForCellJustification(std::size_t index, float columnWid
 	case (Justification::Left):
 		return; // No modification required for left justifited
 	case (Justification::Right):
-		cell.textRelativePosition.x() += columnWidth - getCellFont(index)->width(cell.text);
+		cell.textRelativePosition.x += columnWidth - getCellFont(index)->width(cell.text);
 		return;
 	default:
 		return;
--- a/OPHD/UI/StringTable.cpp
+++ b/OPHD/UI/StringTable.cpp
@@ -18,7 +18,7 @@ void StringTable::draw(NAS2D::Renderer& renderer) const
 {
 	for (std::size_t i = 0; i < cells.size(); ++i)
 	{
-		const Cell& cell = cells.at(i);
+		const auto& cell = cells.at(i);
 
 		NAS2D::Color textColor = cell.textColor != Cell::ColorEmpty ? cell.textColor : defaultTextColor;
 
@@ -124,7 +124,7 @@ void StringTable::computeRelativeCellPositions()
 
 void StringTable::accountForCellJustification(std::size_t index, float columnWidth)
 {
-	Cell& cell = cells[index];
+	auto& cell = cells[index];
 
 	switch (cell.justification)
 	{
--- a/OPHD/UI/StringTable.cpp
+++ b/OPHD/UI/StringTable.cpp
@@ -6,6 +6,16 @@
 
 const NAS2D::Color StringTable::Cell::ColorEmpty = NAS2D::Color::NoAlpha;
 
+StringTable::Cell& StringTable::operator[](const CellCoordinate& coordinate)
+{
+	return cells[getCellIndex(coordinate)];
+}
+
+StringTable::Cell& StringTable::at(std::size_t column, std::size_t row)
+{
+	return cells[getCellIndex(CellCoordinate(column, row))];
+}
+
 StringTable::StringTable(std::size_t columns, std::size_t rows) : columnCount(columns), rowCount(rows)
 {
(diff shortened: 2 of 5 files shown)


OPHD/UI/StringTable.h
```diff
@@ -0,0 +1,84 @@
...
+        // Set textColor to ColorEmpty to indicate cell should use default StringTable color
+        static const NAS2D::Color ColorEmpty;
+
+        // Use defaultFont if not set
+        NAS2D::Font* font = nullptr;
+        std::string text;
+        Justification justification = Justification::Left;
+        NAS2D::Color textColor = ColorEmpty;
+        // Position relative to the StringTable's position
+        NAS2D::Point<float> textRelativePosition;
```
DanRStevens, line 61: A "relative" position should generally be a `Vector` rather than a `Point`.

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,236 @@
...
+}
+
+void StringTable::draw(NAS2D::Renderer& renderer) const
+{
+	for (std::size_t i = 0; i < cells.size(); ++i)
+	{
+		const Cell& cell = cells.at(i);
+
+		NAS2D::Color textColor = cell.textColor != Cell::ColorEmpty ? cell.textColor : defaultTextColor;
+		NAS2D::Point<float> drawPosition = { position.x() + cell.textRelativePosition.x(), position.y() + cell.textRelativePosition.y() };
```
DanRStevens, line 24: If the relative position was changed to a `Vector`, this could be written as:
```cpp
const auto drawPosition = position + cell.textRelativePosition;
```

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+	return position;
+}
+
+void StringTable::setDefaultFont(NAS2D::Font* font)
+{
+	if (font == nullptr)
+	{
+		throw std::runtime_error("Default font may not be a nullptr");
+	}
```
DanRStevens, line 45: Since there is a non-null assumption here, you could potentially change the parameter type to a reference. References are non-nullable. That would eliminate the need for a check.

DanRStevens:
Leeor had a few saved games packaged with the data assets. If you load one of them, you should be able to double click on a factory (or other building?) to popup a report screen. Factory is one of the tabs on that screen.

OPHD/UI/StringTable.h
Brett208: Okay. I could use some help understanding the difference and use cases. I was having trouble with Point when trying to add 2 points together.

Brett208:
Force push to remove line of test code accidently committed

OPHD/UI/FactoryProduction.cpp
```diff
@@ -212,19 +212,25 @@ void FactoryProduction::update()
 
 	Window::update();
 
-	const int labelWidth = 120;
-	auto position = mRect.startPoint() + NAS2D::Vector{constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), 25};
-	drawLabelAndValueRightJustify(position, labelWidth, "Turns Completed:", std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild()));
+	StringTable stringTable(2, 5);
+	stringTable.setPosition({ mRect.x() + constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), mRect.y() + 25 });
```
Brett208, line 216: I wanted to remove the static_cast on this line, but it is only tangentially related to StringTable. Is there any reason to keep this cast in place, it seemed not necessary to me.

OPHD/UI/StringTable.h
```diff
@@ -0,0 +1,85 @@
...
+        Left,
+        Right
+    };
+
+    // Set default fonts in constructor
+    StringTable(std::size_t columns, std::size_t rows);
+
+    void draw(NAS2D::Renderer& renderer) const;
+
+    void setPosition(NAS2D::Vector<float> position);
```
Brett208, line 28: I noticed in other NAS2D classes, Set/Get Position is completed as a function overload:

```
void position(NAS2D::Vector<float> position)
NAS2D::Vector<float> position()
```

Is the overload preferred to get/set here?

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,237 @@
...
+		const Cell& cell = cells.at(i);
+
+		NAS2D::Color textColor = cell.textColor != Cell::ColorEmpty ? cell.textColor : defaultTextColor;
+		const auto drawPosition = position + cell.textRelativePosition;
+
+		renderer.drawText(*getCellFont(i), cell.text, NAS2D::Point<float>(drawPosition.x, drawPosition.y), textColor);
+	}
+}
+
+void StringTable::setPosition(NAS2D::Vector<float> tablePosition)
```
Brett208, line 30: Would it be preferred to make all these short get/set functions as inline in the header file?

OPHD/UI/StringTable.h
DanRStevens: A `Point` represents a position, while a `Vector` represents a displacement.

Addition is supported in the following way:
- `Vector + Vector ==> Vector`
- `Point + Vector ==> Point` (equivalently `Vector + Point ==> Point`, assuming it's defined)
- `Point + Point` invalid

Similarly, for subtraction:
- `Vector - Vector ==> Vector`
- `Point - Vector ==> Point`
- `Point - Point ==> Vector`
- `Vector - Point` invalid (unless you define `-Point` and treat it as `-Point + Vector`)

Of course it does make geometric sense to average points:
- `(Point + Point) / 2 ==> Point`
- `(Point + Point + Point) / 3  => Point`
- ...

----

One way to represent this is to add a 3rd coordinate, which is 0 for vectors and 1 for points:
- `(x1, y1, 0) + (x2, y2, 0) ==> (x1 + x2, y1 + y2, 0)` (Vector)
- `(x1, y1, 1) + (x2, y2, 0) ==> (x1 + x2, y1 + y2, 1)` (Point)
- `(x1, y1, 1) + (x2, y2, 1) ==> (x1 + x2, y1 + y2, 2)` (invalid)

Of course, since that 3rd coordinate is basically encoded in the type system, and statically checked by the compiler, there is no need to allocate any memory for it, nor spend any CPU cycles calculating it.

Works for point averaging as well:
- `((x1, y1, 1) + (x2, y2, 1)) / 2  ==> ((x1 + x2) / 2, (y1 + y2) / 2, 1) ` (Point)

DanRStevens:
Just noticed the new files in this branch are using space for indentation, while most of the project is using tabs for indentation.

OPHD/UI/FactoryProduction.cpp
DanRStevens: Try it and see.

----

Often it's necessary for type conversion in cases like this. The temporary being passed as the parameter is initialized with `{}`, which does not allow narrowing conversions (of non-literals). That means the types in the expressions have to match exactly to be able to construct an object with the correct `BaseType`.

The constructor for `NAS2D::Vector<BaseType>` is defined to take two `BaseType` parameters. For the `{}` brace initialization to be valid, the types of the expressions need to match exactly. If one expression is of type `int`, while the other is of type `float`, it will be an error. The `BaseType` can not be correctly deduced in that case. Sometimes you need to add in a value that's a different type, and so you need to convert it to prevent it from changing the type of the expression. In particular, the `float` type tends to dominate, so you can end up with expressions of the type: `float + int ==> float`.

In this case, it appears the target type is `float`, so casting to `int` does indeed seem a bit odd here. You're probably right to remove it.

----

The other possibility, is that a lot of the code in this project will cast to `int` as a way to force rounding of a value. What the code really means in such case is `std::floor(value)`. If we find that is the case, and the rounding is needed, replacing the cast with `std::floor` makes more sense when you want to round a `float` type and end up with a final expression that's still of `float` type. The might be the intention here.

OPHD/UI/StringTable.h
DanRStevens: It would be stylistically closer to existing code.

Though with that said, the refactoring has been making some of those overloads disappear. I'm not sure it's a style we necessarily want to maintain, though I'm also a bit hesitant to deviate from it.

Of course, if `StringTable` were to derive from `Control`, it would obtain such methods through inheritance.

OPHD/UI/StringTable.cpp
DanRStevens: Uncertain.

Sometimes it feels lazy to define methods in the header file. Under some conditions that can lead to duplicate code generation. On the other hand, having inline methods can aid inlining as a compiler optimization. I'm not sure if recent compilers need inline code to inline a method though. I believe such a transformation might fall under link time optimization, which is something older compilers didn't really have.

In this particular case, it might make sense to get the method by deriving from `Control`.

OPHD/UI/StringTable.h
```diff
@@ -0,0 +1,85 @@
...
+        Justification justification = Justification::Left;
+        NAS2D::Color textColor = ColorEmpty;
+        // Position relative to the StringTable's position
+        NAS2D::Vector<float> textRelativePosition;
+    };
+
+    std::vector<Cell> cells;
+    const std::size_t columnCount;
+    const std::size_t rowCount;
+    NAS2D::Vector<float> position;
```
DanRStevens, line 68: A "position" field should generally use `Point` rather than `Vector`.

- `NAS2D::Point<float> position`
- `NAS2D::Vector<float> textRelativePosition`

Result:
- `const auto textPosition = position + textRelativePosition; // Type deduced as Point`

OPHD/UI/StringTable.h
Brett208: Ah yes, you cannot add two locations together as that wouldn't make any since. Thanks for the math refresher, I'm a bit embarrassed I didn't recognize this up front.

In MonoGame (XNA), this behaviour is allowed:

```CSharp
Point point1 = new Point(0, 1);
Point point2 = new Point(1, 1);
Point point3 = point1 + point2;
```

I guess while logically incorrect to add two points (location), the end result works for displaying graphics on the screen, collision avoidance, etc. I had just never realized I was doing it incorrectly in other projects.

I'll try to fix this up today and use NAS2D::Point for the position and NAS2D::Vector for the relative position (displacement) of each cell.

OPHD/UI/FactoryProduction.cpp
Brett208: OK, I'll try removing it. Agreed that std::floor would indicate intent better than a static_cast in these situations.

I have found with experience that while you can draw sprites on screen using floats, the image isn't as sharp when fractional positions are provided.

-Brett

OPHD/UI/StringTable.h
Brett208: OK, I'll leave it up to you or @ldicker83 if you would prefer StringTable to inherit from 'Control'. I wasn't sure it needed full support for things like visibility, focus, textchange events, etc. 

Otherwise, I'll try to match the existing terminology. It is easy to change the function names at a later date.

-Brett

OPHD/UI/StringTable.cpp
Brett208: Lets just leave the definitions in the cpp file for now. I suspect sense we are using C++17 compliant compilers, it is probably already optimizing well and no reason to try and push the compiler for inlining if there isn't a specific reason.'

I used to try to inline all the smaller calls, but I sort of like the cleaner header feel if you don't see inline functions. It makes it easier to read through the different function and variable names involved in the class. (I still prefer C# where you do not have to duplicate the prototype and definition for every class, but oh well.)

Brett208:
@DanRStevens thanks for noticing use of spaces in the header file. Not sure why when I created the cpp file it used tabs and the header used spaces. I'll standardize the header to tabs soon.

OPHD/UI/StringTable.cpp
Brett208: See comment above about control interface.

OPHD/UI/StringTable.h
DanRStevens: This reminds me, I would like to pare down the functionality for `Control`. It contains a few things that aren't truly generic, such as text change. Not all controls have text or focus.

----

I've noticed a lot of existing code prefixes member variables names with "m", which is one way to avoid name collisions. In my own projects I usually try to avoid name prefixes, though it's pretty widespread in this project.

ldicker83:
I've been following along, I don't have a whole lot to contribute so if the code looks good and all the checks pass (they passed as of 30 something minutes ago) I'm game for merging this.

The overall look in the screenshot is pretty good -- I see this beginning very useful later on especially if we can do cell highlighting (different change set me thinks).

OPHD/UI/StringTable.h
```diff
@@ -0,0 +1,86 @@
...
+	enum class Justification
+	{
+		Left,
+		Right
+	};
+
+	struct Cell
+	{
+		// Set textColor to ColorEmpty to indicate cell should use default StringTable color
+		static const NAS2D::Color ColorEmpty;
```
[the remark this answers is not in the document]
Brett208: I'm ok with it in either location. I debated both location without a strong preference. It ended up inside Cell since it typically shouldn't be applied to the Table's DefaultTextColor (although nothing would stop you from applying it as such I suppose.

OPHD/UI/StringTable.h
```diff
@@ -0,0 +1,86 @@
...
+
+		// Use defaultFont if not set
+		NAS2D::Font* font = nullptr;
```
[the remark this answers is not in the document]
Brett208: My purpose in using nullptr for font and ColorEmpty in the cell is to allow dynamically changing the text and color of the cell after adding data. If you specifically set the text color or font of the cell, but used what was the default value, the cell would retain this value even if the actual default was changed. 

Also, setting default font and text color of a cell does not require pulling the default table values, since you can just use nullptr or ColorEmpty. 

Maybe this behaviour is too complex or subtle though? We could mandate that default color and font are set in the constructor and is considered nonmutable.

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,221 @@
...
+			auto index = getCellIndex(CellCoordinate(column, row));
+			float cellHeight = static_cast<float>(getCellFont(index)->height());
+
+			if (cellHeight > rowHeight)
+			{
+				rowHeight = cellHeight;
+			}
+		}
+
+		rowHeights.push_back(rowHeight);
```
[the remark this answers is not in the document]
Brett208: I like the idea of using std::max as suggested here.

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,221 @@
...
+void StringTable::accountForCellJustification(std::size_t index, float columnWidth)
+{
+	auto& cell = cells[index];
+
+	switch (cell.justification)
+	{
+	case (Justification::Left):
+		return; // No modification required for left justifited
+	case (Justification::Right):
+		cell.textRelativePosition.x += columnWidth - getCellFont(index)->width(cell.text);
```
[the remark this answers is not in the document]
Brett208: Correct, since the cell purposely does not know what the StringTable's default font is, the cell's font should not be directly queried, but StringTable::getCellFont should be called instead. See comment above that we could just initialize the default on construction of StringTable if this behaviour is not appreciated.

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+void StringTable::computeRelativeCellPositions()
+{
+	// Must be at least 1 row and 1 column to compute position values
+	if (columnCount == 0 || rowCount == 0)
+	{
+		return;
+	}
```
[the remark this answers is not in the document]
Brett208: Yeah, I just checked and you are correct. This was not the case in an earlier iteration when I was writing the code. We can remove this call now though.

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+	defaultFont = NAS2D::Utility<FontManager>::get().font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
+	defaultTitleFont = NAS2D::Utility<FontManager>::get().font(constants::FONT_PRIMARY_BOLD, constants::FONT_PRIMARY_NORMAL);
+}
+
+void StringTable::draw(NAS2D::Renderer& renderer) const
+{
+	for (std::size_t i = 0; i < cells.size(); ++i)
+	{
+		const auto& cell = cells.at(i);
```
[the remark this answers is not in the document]
Brett208: See comments above on if it is desirable to give cell struct knowledge of default font, default title font or color.

OPHD/UI/StringTable.h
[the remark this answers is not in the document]
DanRStevens: Ahh, that explains it. Yeah, I guess I would have gone with a non-mutable assumption for the defaults. Would seem odd to be changing those dynamically after construction. Maybe create some sort of disco table.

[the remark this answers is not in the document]
Brett208: Yeah, could be a disco table. :)

OPHD/UI/StringTable.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+StringTable::Cell& StringTable::operator[](const CellCoordinate& coordinate)
+{
+	return cells[getCellIndex(coordinate)];
+}
+
+StringTable::Cell& StringTable::at(std::size_t column, std::size_t row)
+{
+	return cells[getCellIndex(CellCoordinate(column, row))];
+}
```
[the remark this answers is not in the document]
Brett208: I would like help with the design here. I was researching the operator[] and it seems a bit complicated to setup something like operator[][]. Also, I don't know that you can overload the operator[].

@DanRStevens, do you think it would be better to just have an overloaded 'at' function? This would be easiest to implement on my part.

OPHD/UI/FactoryProduction.cpp
```diff
@@ -212,19 +212,25 @@ void FactoryProduction::update()
...
-	auto position = mRect.startPoint() + NAS2D::Vector{constants::MARGIN * 2 + static_cast<int>(mProductGrid.width()), 25};
-	drawLabelAndValueRightJustify(position, labelWidth, "Turns Completed:", std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild()));
+	StringTable stringTable(2, 5);
+	stringTable.setPosition(mRect.startPoint() + NAS2D::Vector<float>{ constants::MARGIN * 2 + mProductGrid.width(), 25 });
+	stringTable.setColumnJustification(1, StringTable::Justification::Right);
 
-	position.y() += 20;
-	drawLabelAndValueRightJustify(position, labelWidth, "Common Metals:", std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild()));
+	stringTable.at(0, 0).text = "Turns Completed:";
+	stringTable.at(1, 0).text = std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild());
```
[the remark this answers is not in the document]
Brett208: I like this idea. I would have named these functions 'SetColumn' though. I'll try to look into StringValue.

Brett208:
Thanks for reviewing Dan and Leeor. 

@DanRStevens, I haven't tried merging any code yet, but probably have permissions.

OPHD/UI/StringTable.cpp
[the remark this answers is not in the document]
DanRStevens: Hmm, seems you can't overload `operator[]` to take two comma separated parameters. I wouldn't advise trying to use `[][]` syntax.

----

I suppose you could just manually pack the parameters using `{}`, and then get rid of the `at` method. Calling code would look something like:
```cpp
	stringTable[{0, 0}].text = "Turns Completed:";
	stringTable[{1, 0}].text = std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild());

	stringTable[{0, 1}].text = "Common Metals:";
	stringTable[{1, 1}].text = std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild());

	stringTable[{0, 2}].text = "Common Minerals:";
	stringTable[{1, 2}].text = std::to_string(mProductCost.commonMinerals() * mProductCost.turnsToBuild());

	stringTable[{0, 3}].text = "Rare Metals:";
	stringTable[{1, 3}].text = std::to_string(mProductCost.rareMetals() * mProductCost.turnsToBuild());

	stringTable[{0, 4}].text = "Rare Minerals:";
	stringTable[{1, 4}].text = std::to_string(mProductCost.rareMinerals() * mProductCost.turnsToBuild());
```

There is one complication internally with the line:
```cpp
		this->at(column, row).justification = justification;
```

This can be replaced with:
```cpp
		this->operator[]({column, row}).justification = justification;
```
Or with:
```cpp
		(*this)[{column, row}].justification = justification;
```

OPHD/UI/FactoryProduction.cpp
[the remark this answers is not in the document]
DanRStevens: Ahh, right. I guess I was thinking `(row, column)` order, which is reversed from `(x, y) = (column, row)`. I was trying to work backwards from the assignments: `stringTable.at(0, 1).text = "Common Metals:";`

PR #476, opened 2020-07-06
Add StringTable::SetColumnText & setRowText

Brett208:
Took a stab at allowing setting an entire row or column of text.

### DIFF
--- a/OPHD/UI/FactoryProduction.cpp
+++ b/OPHD/UI/FactoryProduction.cpp
@@ -216,20 +216,23 @@ void FactoryProduction::update()
 	stringTable.position(mRect.startPoint() + NAS2D::Vector<float>{ constants::MARGIN * 2 + mProductGrid.width(), 25 });
 	stringTable.setColumnJustification(1, StringTable::Justification::Right);
 
-	stringTable.at(0, 0).text = "Turns Completed:";
-	stringTable.at(1, 0).text = std::to_string(mFactory->productionTurnsCompleted()) + " of " + std::to_string(mProductCost.turnsToBuild());
+	stringTable.setColumnText(0,
+		{
+			"Turns Completed:",
+			"Common Metals:",
+			"Common Minerals:",
+			"Rare Metals:",
+			"Rare Minerals:"
+		});
 
-	stringTable.at(0, 1).text = "Common Metals:";
-	stringTable.at(1, 1).text = std::to_string(mProductCost.commonMetals() * mProductCost.turnsToBuild());
-
-	stringTable.at(0, 2).text = "Common Minerals:";
-	stringTable.at(1, 2).text = std::to_string(mProductCost.commonMinerals() * mProductCost.turnsToBuild());
-
-	stringTable.at(0, 3).text = "Rare Metals:";
-	stringTable.at(1, 3).text = std::to_string(mProductCost.rareMetals() * mProductCost.turnsToBuild());
-
-	stringTable.at(0, 4).text = "Rare Minerals:";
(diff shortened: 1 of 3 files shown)


OPHD/UI/StringTable.cpp
```diff
@@ -72,6 +72,26 @@ void StringTable::setVerticalPadding(float padding)
 	mVerticalPadding = padding;
 }
 
+void StringTable::setColumnText(std::size_t column, const std::vector<NAS2D::StringValue>& rows)
+{
+	checkCellIndex({ column, rows.size() - 1});
+
+	for (std::size_t row = 0; row < rows.size(); ++row)
+	{
+		mCells[getCellIndex({ column, row })].text = rows[row].to<std::string>();
```
DanRStevens, line 81: You can potentially shorten the `rows[row].to<std::string>()` to `rows[row].value`.

I guess you could say `StringValue` is basically already a `std::string`, with some conversion functionality added.

OPHD/UI/StringTable.cpp
```diff
@@ -72,6 +72,26 @@ void StringTable::setVerticalPadding(float padding)
...
+	}
+}
+
+void StringTable::setRowText(std::size_t row, const std::vector<NAS2D::StringValue>& columns)
+{
+	checkCellIndex({ columns.size() - 1, row });
+
+	for (std::size_t column = 0; column < columns.size(); ++column)
+	{
+		mCells[getCellIndex({ column, row })].text = columns[column].to<std::string>();
```
DanRStevens, line 91: Ditto.

OPHD/UI/FactoryProduction.cpp
```diff
@@ -216,20 +216,23 @@ void FactoryProduction::update()
...
-	stringTable.at(0, 4).text = "Rare Minerals:";
-	stringTable.at(1, 4).text = std::to_string(mProductCost.rareMinerals() * mProductCost.turnsToBuild());
+	stringTable.setColumnText(0,
+		{
+			"Turns Completed:",
+			"Common Metals:",
+			"Common Minerals:",
+			"Rare Metals:",
+			"Rare Minerals:"
+		});
```
DanRStevens, line 226: I find this style of indentation slightly odd. I guess I'm a little uncertain of the rules being followed here.

Generally when I spill arguments, I try to spill all arguments, that way arguments at the same level are indented to the same position (without excessive spacing to try and line things up at the first argument). Here the `0` is at a different indentation from the `std::vector` argument, despite being at the same nesting level. By spilling it to the next line, while adding a single level of indentation, both arguments will have matching indentation.

The close at `});` doesn't seem to match up well with the open, where `(` and `{` are at different levels.

Granted, the style I typically use here is influenced by the [1TBS](https://en.wikipedia.org/wiki/Indentation_style#Variant:_1TBS_(OTBS)), which is not the style Leeor follows for these projects.

Anyway, not a strong objection to the style, just thought I'd bring it up since I'm uncertain of established practice here.

----

For reference, my original comment with an alternate style can be found at:
https://github.com/OutpostUniverse/OPHD/pull/466#discussion_r449983538

OPHD/UI/StringTable.cpp
Brett208: Ah, thanks. I didn't see the value option for StringValue.

OPHD/UI/FactoryProduction.cpp
Brett208: Ok. I'm happy with either syntax.

PR #487, opened 2020-07-08
Agridome Refactor

Brett208:
I'm a little confused on how local vs global storage works, but figured those concerns could be separate from adding basic UI to the Agridome Structure Inspector.

Agridome

### DIFF
--- a/OPHD/Things/Structures/Agridome.h
+++ b/OPHD/Things/Structures/Agridome.h
@@ -17,6 +17,19 @@ public:
 		requiresCHAP(true);
 	}
 
+	StringTable createInspectorViewTable()
+	{
+		StringTable stringTable(2, 2);
+
+		stringTable[{0, 0}].text = "Food Stored:";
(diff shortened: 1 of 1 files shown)


OPHD/Things/Structures/Agridome.h
```diff
@@ -54,4 +42,25 @@ class Agridome : public Structure
...
+		{
+			return 0;
+		}
+
+		if (AGRIDOME_CAPACITY < storage().food() + AGRIDOME_BASE_PRODUCUCTION)
+		{
+			return AGRIDOME_CAPACITY - storage().food();
+		}
+
+		return AGRIDOME_BASE_PRODUCUCTION;
```
DanRStevens, line 72: This could be simplified as:
```cpp
return std::min(AGRIDOME_BASE_PRODUCUCTION, AGRIDOME_CAPACITY - storage().food());
```

Brett208: Oh nice idea.

Issue #479, opened 2020-07-07
Remove ListBox duplication


DanRStevens:
There are two independent classes:
- `ListBox`
- `ListBoxBase`

They are highly similar, though don't derive from each other. We should seek to eliminate one of them.

Currently `ListBox` is used by:
- `ComboBox`
- `FileIo`
- `FactoryReport`

Currently `ListBoxBase` is used as a base class by:
- `FactoryListBox`
- `ProductListBox`
- `StructureListBox`

It seems `ListBox` may be slightly more feature complete, so that may be the candidate class to keep. To that end, we should investigate using `ListBox` as a base class instead of `ListBoxBase`. It may be some features will need to be added before this can be done.

DanRStevens:
One of the notable differences is support for sorting:
```cpp
	void sorted(bool isSorted) { mSorted = isSorted; }
	bool sorted(bool) const { return mSorted; }

	void sort() { if (mSorted) { std::sort(mItems.begin(), mItems.end()); } }
```

I found this implementation a little non-intuitive. Calling `sort()` does nothing unless the `sorted` property is first set to true. I was thinking the `ListBox` should sort the contents whenever `sort()` is called. There doesn't seem to be a clear reason to first require setting a `sorted` property.

Additionally the sorted property may be misleading, in that it could be taken to mean the contents are current sorted. However, having the `sorted` property set doesn't mean the contents are actually sorted. The `sort()` call may not have been called yet, or a new item may have been added since last calling `sort()`.

DanRStevens:
To add to the `sort` topic:

The `ComboBox` explicitly sets the `sorted` property to `false`, but this is the default value, so does nothing.

The `FileIo` control calls `.sort()` manually, but does not set the `sorted` property, which defaults to `false`. Technically this means the `.sort()` call from `FileIo` does nothing. This is probably unintentional.

The use of `ListBox` from `FactoryReport` does not make any use of the `sort()` or `sorted` features. The default is to not sort.

In all cases, no sorting is done. This feature is effectively unused.

----

Additionally, the `ListBox::removeItem` method calls `sort()` after removing an item. If the list is already sorted before removal, then it will still be sorted after removal. There is no need to re-sort the list.

DanRStevens:
To add to the above:

[...]

PR #489, opened 2020-07-09
Remove ListBox sort during add

DanRStevens:
Reference: #479

Remove feature where `ListBox` tries to resort the list when new items are added. This is a rather inefficient way to keep the list sorted. Instead, items should be pre-sorted before being added, or should all be added at once, and then sorted.

Curiously, removing most of this feature means the `FileIo` list box will now be sorted, where it wasn't sorted before. This is because it manually called `sort()`, which previously did nothing if the `sorted` property wasn't set to `true` first.


DanRStevens:
Ok, I'm going to merge this now. Pretty sure the code is good. Just it's one of those changes I wanted you to be aware of.

PR #495, opened 2020-07-09
Add communication range to CommTower structure inspector

Brett208:


OPHD/Things/Structures/CommTower.h
```diff
@@ -18,6 +19,28 @@ class CommTower : public Structure
...
+	{
+		StringTable stringTable(2, 1);
+
+		stringTable[{0, 0}].text = "Communication Range:";
+
+		auto communicationRange = operational() ? constants::COMM_TOWER_BASE_RANGE : 0;
+
+		std::stringstream stringStream;
+		stringStream.precision(0);
+		stringStream << communicationRange;
```
DanRStevens, line 32: Maybe we should change `constants::COMM_TOWER_BASE_RANGE` type to `int`, and use `std::to_string` here.

It seems `constants::COMM_TOWER_BASE_RANGE` is only used in one other place, where a `float` is expected, though honestly a `distanceSquared` method with an `int` might have been more efficient.

Brett208: I was surprised it was a float. Changing to int seems good to me.

PR #504, opened 2020-07-10
Pixel aligned Controls

### LINKED ISSUE #434: Reintroduce floating point truncation for control positioning
This was stripped out in a previous pull request. I understand why it was done but it ultimately creates graphical issues when drawing GUI controls with positional information that has a fractional component.

That stated, originally this was done by truncating the fractional component from windows, buttons, etc. This can instead be relegated to the base Control and handled in the position handling functions.

This should be done before the next release.

DanRStevens:
Closes #434

Convert `Control` `position` and `size` from `float` to `int`. This enforces pixel level alignment of controls, which should prevent fuzzing `Control` borders and text.


ldicker83:
Bah, you beat me to it! :D

Issue #383, opened 2020-04-14
Ogg music playback error


DanRStevens:
Reference: #219

Ogg music playback was broken by the change from NuGet to Vcpkg for the SDL2-Mixer dependency in https://github.com/lairworks/nas2d-core/pull/371.

Error from `ophd.log`:
> Music::load(): Unrecognized audio format

----

There is an update for the SDL2-Mixer install in https://github.com/lairworks/nas2d-core/pull/599, which enables the extra audio formats, and changes from static linking to dynamic linking to work around `vcpkg` config problems. This installs SDL2-Mixer with the options:
```
SDL2-mixer[dynamic-load,libflac,libmodplug,libvorbis,opusfile]
```

If SDL2-Mixer was already installed with the previous settings, it may need to be removed with:
```
vcpkg remove sdl2-mixer:x64-windows sdl2-mixer:x86-windows
```

Once removed, the build scripts should automatically try to re-install it. Assuming the submodule has been updated for this change, SDL2-Mixer will be installed with the new settings.

With the updated `vcpkg` install options, the error message from `ophd.log` changes to:
> Music::load(): Failed loading vorbisfile.dll: The specified module could not be found.

As suggested by https://github.com/lairworks/nas2d-core/pull/599, a possible workaround is to manually copy the missing DLL, along with it's dependency DLLs to the executable folder:
- `vorbisfile.dll`
- `vorbis.dll`
- `ogg.dll`

These files can be found in the folder: `$(VcpkgRoot)bin\`
They need to be copied to: `$(TargetDir)`

A PostBuildStep can be setup to do the file copying, much like how `vcpkg` normally does it.

----

More long term, it might be nice to investigate why `vcpkg` doesn't copy the needed DLLs for this dependency, but does copy them for other dependencies. This likely relates to the option settings, as those DLLs are only needed to support the additional options. That may end up being a `vcpkg` fix though.

DanRStevens:
I've been looking into how `vcpkg` determines what DLLs to copy for the dependencies. I figured it did something to the computer during `vcpkg integrate install`, and from there would intercept builds to add in additional steps, including copying DLLs for the dependencies. Below is a summary of my findings.

----

The help text for "integrate install", and associated code, can be found in the source file: `vcpkg/toolsrc/src/vcpkg/base/commands.integrate.cpp`

[...]

PR #505, opened 2020-07-10
Copy OGG related dlls on post build for Visual Studio

Brett208:
Related Issue #383

### DIFF
--- a/OPHD/ophd.vcxproj
+++ b/OPHD/ophd.vcxproj
@@ -105,6 +105,16 @@
       <TargetMachine>MachineX86</TargetMachine>
       <AdditionalLibraryDirectories>$(SolutionDir)$(Configuration)\;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
+    <PostBuildEvent>
+      <Command>SET sourceDir="$(VcpkgCurrentInstalledDir)bin\"
+if $(ConfigurationName) == Debug (
+  SET sourceDir="$(VcpkgCurrentInstalledDir)debug\bin\"
+)
+
+xcopy /y /d "%sourceDir%ogg.dll" "$(OutDir)"
+xcopy /y /d "%sourceDir%vorbis.dll" "$(OutDir)"
+xcopy /y /d "%sourceDir%vorbisfile.dll" "$(OutDir)"</Command>
+    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
@@ -128,6 +138,16 @@
       <SubSystem>Console</SubSystem>
       <AdditionalLibraryDirectories>$(SolutionDir)$(Platform)\$(Configuration)\;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
+    <PostBuildEvent>
+      <Command>SET sourceDir="$(VcpkgCurrentInstalledDir)bin\"
+if $(ConfigurationName) == Debug (
+  SET sourceDir="$(VcpkgCurrentInstalledDir)debug\bin\"
+)
+
+xcopy /y /d "%sourceDir%ogg.dll" "$(OutDir)"
+xcopy /y /d "%sourceDir%vorbis.dll" "$(OutDir)"
+xcopy /y /d "%sourceDir%vorbisfile.dll" "$(OutDir)"</Command>
+    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <ClCompile>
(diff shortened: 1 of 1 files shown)


OPHD/ophd.vcxproj
```diff
@@ -105,6 +105,16 @@
       <TargetMachine>MachineX86</TargetMachine>
       <AdditionalLibraryDirectories>$(SolutionDir)$(Configuration)\;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
+    <PostBuildEvent>
+      <Command>SET sourceDir="$(VcpkgCurrentInstalledDir)bin\"
+if $(ConfigurationName) == Debug (
+  SET sourceDir="$(VcpkgCurrentInstalledDir)debug\bin\"
+)
```
DanRStevens, line 112: I suspect this can be simplified by using `VcpkgConfigSubdir`, which is conditionally either empty `''`, or `'debug\'`. That would allow us to remove the conditional block of our own.

https://github.com/microsoft/vcpkg/blob/544f8e4593764f78faa94bac2adb81cca5232943/scripts/buildsystems/msbuild/vcpkg.targets#L65

Effectively we could simplify this to:
```
SET sourceDir="$(VcpkgCurrentInstalledDir)$(VcpkgConfigSubdir)bin\"
```

OPHD/ophd.vcxproj
```diff
@@ -105,6 +105,16 @@
...
     </Link>
+    <PostBuildEvent>
+      <Command>SET sourceDir="$(VcpkgCurrentInstalledDir)bin\"
+if $(ConfigurationName) == Debug (
+  SET sourceDir="$(VcpkgCurrentInstalledDir)debug\bin\"
+)
+
+xcopy /y /d "%sourceDir%ogg.dll" "$(OutDir)"
+xcopy /y /d "%sourceDir%vorbis.dll" "$(OutDir)"
+xcopy /y /d "%sourceDir%vorbisfile.dll" "$(OutDir)"</Command>
```
DanRStevens, line 116: Any idea what the difference is between `$(OutDir)` and `$(TargetDir)`?

Reference:
https://docs.microsoft.com/en-us/cpp/build/reference/common-macros-for-build-commands-and-properties?redirectedfrom=MSDN&view=vs-2019

It's not clear to me which one should be used.

OPHD/ophd.vcxproj
Brett208: I'm not sure that VcpkgConfigSubdir is working as advertised? You should be able to see the expansion below. Maybe the spelling is wrong or something else?

VcpkgConfigSubdir Not Working

OPHD/ophd.vcxproj
Brett208: No, I've been using OutDir successfully on other projects and as mentioned the documentation isn't too clear on what the difference is.

Brett208:
@DanRStevens  See updated script. I think it prevents from running when vcpkg is not installed now...

OPHD/ophd.vcxproj
DanRStevens: Huh, fascinating. I wonder if it might be a recently added variable that perhaps doesn't exist in the copy you're using.

Probably best not to rely on it if it's not likely to work for other people.

OPHD/ophd.vcxproj
DanRStevens: I wonder if they might differ if `$(TargetPath)` is defined to be outside of `$(OutDir)`.

I noticed the Vcpkg script uses `$(TargetPath)` in its script, which is assume is more closely related to `$(TargetDir)` than to `$(OutDir)`. This may be splitting hairs though.

OPHD/ophd.vcxproj
```diff
@@ -106,9 +106,16 @@
...
+      <Command>SET vcpkgDir="$(VcpkgCurrentInstalledDir)"
+
+IF "vcpkgDir" == "" (
+  EXIT
+)
+
+SET sourceDir="%vcpkgDir%bin\"
+
+IF $(ConfigurationName) == Debug (
+  SET sourceDir="%vcpkgDir%debug\bin\"
```
DanRStevens, line 118: I had to read this a couple times to understand the logic. And I'm maybe mildly concerned if `EXIT` works in this context.

What if we check the original variable first, and then set variables based on configuration.
```
IF NOT "$(VcpkgCurrentInstalledDir)" == "" (
  IF "$(ConfigurationName)" == "Debug" (
    SET configDir="debug\"
  ) ELSE (
    SET configDir=""
  )

  SET sourceDir="$(VcpkgCurrentInstalledDir)%configDir%bin\"

  xcopy /y /d "%sourceDir%ogg.dll" "$(TargetDir)"
  xcopy /y /d "%sourceDir%vorbis.dll" "$(TargetDir)"
  xcopy /y /d "%sourceDir%vorbisfile.dll" "$(TargetDir)"
)
```

The `ELSE` block might not be strictly necessary, though probably a good idea in case the variable was already defined in the environment.

Brett208: I thought you were asking me to write the script. :) No issues with the rewrite.

One thing I found concerning is when googling VcpkgCurrentInstalledDir, no documentation seems to come up. Maybe because it is a new feature? I would have expected mention in MSDN or maybe on the vcpkg github page.

I installed vcpkg I think 2-3 months ago, maybe a bit earlier. Agree that if VcpkgConfigSubdir is actually that new, probably better not to support for a while? If you want closure, I can reinstall and check. Otherwise I was going to move on for now.

ldicker83:
This shows approved but there's a question there -- are we ready to merge this or no?

Brett208:
@ldicker83, I'd recommend waiting to merge until a commit is made implementing Dan's request (should be quick to do). It wouldn't harm anything if merged as is though and Dan's request added in a different PR.

OPHD/ophd.vcxproj
DanRStevens: Haha, I know. I felt kind of funny after posting that. Thing is, I didn't really know what the form of the solution would be until you first posted something, and I'm not really able to test changes from Linux.

The `VcpkgCurrentInstalledDir` would be Vcpkg specific, so probably no documentation on MSDN. You'd need to check the Vcpkg documentation for that, assuming they even document it.

Vcpkg is under fairly active development, so probably quite a few changes in the last 2-3 months. At any rate, we don't need to be relying on bleeding edge features, so should be fine to just move on.

DanRStevens:
Yeah, for a small change like this, maybe just wait to implement it. Nothing wrong with the original code, just some small optional changes that were made or discussed after the initial review. Brett should be able to merge when ready.

OPHD/ophd.vcxproj
Brett208: I had to set delayed expansion to get to work. Otherwise, seems to be working on my machine. Please merge if CI builds are acceptable.

I thought vcpkg was actively developed by Microsoft.

PR #507, opened 2020-07-12
Planet type to scoped enum

Brett208:
Since PlanetType is a member of Planet, I thought about renaming it just Type, but thought that might be too generic. IE you would call `Planet::Type::Ganymede`.

There are a couple of variables with PLANET_TYPE still in their names, but I didn't change them out as it looked like the changeset would start growing and I wasn't sure it was appropriate at this refactor.


Brett208:
** Force Push to remove 3 lines that were accidently committed via rebase **

DanRStevens:
I would be fine with short generic names such as `Type`, provided they are scoped, as is done here. No objection to the longer name though.

Nice update to the code style.

Tagging #319.

Issue #506, opened 2020-07-11
Use PlanetCharacteristics struct to initialize Planet class


Brett208:
I'd like to consolidate the different planet settings such as hostility, tileset, name, available depth, etc in a data struct called PlanetCharacteristics (Or could be PlanetType if this name is preferred).

Then, I'd like to change the Planet constructor to take a PlanetCharacteristics struct to initialize all the values.

Long term, I think this would feed something like a Planets.xml file that would allow storing the settings outside of source code and just parsing the xml file.

DanRStevens:
That sounds like an excellent plan. I like it.

ldicker83:
I'm game. It makes sense anyway since planet characteristics will expand as environment parameters will start to affect the simulation.

Brett208:
I am trying to write the XML document. I wanted to include the values within the elements without encoding them within an attribute. I think this is standard practice, but cannot seem to figure out how to query the key and value of an element when used in this fashion. It looks like the saved game xml only encodes data in attributes. Is that a requirement of the underlying architecture or desired style?

Below is what I was thinking as a start.

```
<?xml version="1.0" encoding="utf-8"?>
<PlanetTypes xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:noNamespaceSchemaLocation="PlanetTypes.xsd">
  <PlanetType>
    <Type>0</Type>
    <ImagePath>planets/planet_d.png</ImagePath>
    <MaxDepth>1</MaxDepth>
    <MaxMines>10</MaxMines>
  </PlanetType>
</PlanetTypes>
```

I see an xmlElement has a value function but no key function. Maybe I don't understand how to use the NAS2D xml library?

-Brett

DanRStevens:
It's derived from the [TinyXml](http://www.grinninglizard.com/tinyxml/) library. There is also a [TinyXML2](https://github.com/leethomason/tinyxml2) library, which we may move towards, and might someday include as an external dependency. Their documentation might help.

In terms of NAS2D support for reading values from XML files, we mostly only support reading attributes. I've got some work in progress on the `Configuration` class in NAS2D to make it more generic.

Not too sure what to say stylistically about attributes versus tag values. I find attributes result in less redundancy, though don't support nesting.

Brett208:
I think moving the data into xml is waiting for ability to load data not contained in an xml attribute via NAS2D from my perspective. If someone wants to design it for loading through attributes, I would be okay though.

[...]

PR #508, opened 2020-07-12
Planet hostility scoped enum

Brett208:
After moving PlanetHostility into the Planet class, I thought the word Planet became redundant in the enum name. Happy to switch it back to PlanetHostility if this isn't proferred by others.

Prep-work before starting on issue #506


DanRStevens:
Looks good to me. I agree the "Planet" prefix becomes redundant when scoped inside the `Planet` class.

DanRStevens:
Tagging #319

PR #510, opened 2020-07-13
Add Planet Attributes

Brett208:
Next steps are either expanding PlanetAttributes to include more information are create the xml file for PlanetAttributes.

### DIFF
--- a/OPHD/States/PlanetSelectState.cpp
+++ b/OPHD/States/PlanetSelectState.cpp
@@ -13,6 +13,8 @@
 #include "NAS2D/Mixer/Mixer.h"
 #include "NAS2D/Renderer/Renderer.h"
 
+#include <vector>
+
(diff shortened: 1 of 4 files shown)


OPHD/States/PlanetSelectState.cpp
```diff
@@ -86,6 +88,13 @@ PlanetSelectState::~PlanetSelectState()
 	Utility<Mixer>::get().stopAllAudio();
 }
 
+namespace {
+	std::vector<Planet::Attributes> planetAttributes = {
```
DanRStevens, line 92: For hardcoded data, you could potentially use a `std::array` here, since it's a fixed size. Long term though, if data is eventually loaded from a file, the `std::vector` makes more sense. I assume that's what you were going for.

PR #516, opened 2020-07-14
Expand planet attributes

Brett208:
Continued work against #506

### DIFF
--- a/OPHD/States/PlanetSelectState.cpp
+++ b/OPHD/States/PlanetSelectState.cpp
@@ -13,8 +13,6 @@
 #include <NAS2D/Mixer/Mixer.h>
 #include <NAS2D/Renderer/Renderer.h>
 
-#include <vector>
-
 using namespace NAS2D;
 
 Planet::PlanetType PLANET_TYPE_SELECTION = Planet::PlanetType::None;
--- a/OPHD/States/PlanetSelectState.cpp
+++ b/OPHD/States/PlanetSelectState.cpp
@@ -13,9 +13,13 @@
 #include <NAS2D/Mixer/Mixer.h>
 #include <NAS2D/Renderer/Renderer.h>
 
+#include <cstddef>
+#include <limits>
+
 using namespace NAS2D;
 
-Planet::PlanetType PLANET_TYPE_SELECTION = Planet::PlanetType::None;
+std::size_t planetSelection;
+constexpr std::size_t planetSelectionInvalid = std::numeric_limits<int>::max();
 
 static Font* FONT = nullptr;
 static Font* FONT_BOLD = nullptr;
@@ -120,7 +124,7 @@ void PlanetSelectState::initialize()
 	mPlanets[2]->mouseEnter().connect(this, &PlanetSelectState::onMousePlanetEnter);
 	mPlanets[2]->mouseExit().connect(this, &PlanetSelectState::onMousePlanetExit);
 
-	PLANET_TYPE_SELECTION = Planet::PlanetType::None;
+	planetSelection = planetSelectionInvalid;
 
 	mQuit.size({100, 20});
 	mQuit.position({static_cast<int>(renderer.width()) - 105, 30});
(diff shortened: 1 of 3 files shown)


OPHD/States/PlanetSelectState.cpp
```diff
@@ -90,9 +92,9 @@ PlanetSelectState::~PlanetSelectState()
 
 namespace {
 	std::vector<Planet::Attributes> planetAttributes = {
-		{ Planet::Attributes{Planet::PlanetType::Mercury, "planets/planet_d.png", 1, 10} },
-		{ Planet::Attributes{Planet::PlanetType::Mars, "planets/planet_c.png", 4, 30} },
-		{ Planet::Attributes{Planet::PlanetType::Ganymede, "planets/planet_e.png", 2, 15} }
+		{ Planet::Attributes{Planet::PlanetType::Mercury, "planets/planet_d.png", Planet::Hostility::High, 1, 10, "maps/merc_01", "tsets/mercury.png"} },
+		{ Planet::Attributes{Planet::PlanetType::Mars, "planets/planet_c.png", Planet::Hostility::Low, 4, 30, "maps/mars_04", "tsets/mars.png"} },
+		{ Planet::Attributes{Planet::PlanetType::Ganymede, "planets/planet_e.png", Planet::Hostility::Medium, 2, 15, "maps/ganymede_01", "tsets/ganymede.png"} }
```
Brett208, line 97: Intention is to move all this into an xml file eventually. It sounds like NAS2D will have some xml updates occurring so trying to wait for that before moving to xml file.

OPHD/States/PlanetSelectState.cpp
```diff
@@ -192,44 +194,11 @@ State* PlanetSelectState::update()
...
-			hostility = Planet::Hostility::Medium;
-			break;
-
-		default:
-			return mReturnState;
-			break;
-		}
-
-		MapViewState* mapview = new MapViewState(map, tileset, dig_depth, max_mines, hostility);
+		MapViewState* mapview = new MapViewState(planet.mapImagePath(), planet.tilesetPath(), planet.digDepth(), planet.maxMines(), planet.hostility());
```
Brett208, line 201: I am thinking about passing planet into MapViewState instead of all the parameters.

I think Planet should be separated into a Planet and a PlanetView class. PlanetView would be responsible for drawing the planet instead of Planet handling its own draw logic.

OPHD/States/PlanetSelectState.cpp
```diff
@@ -13,9 +13,13 @@
...
 #include <NAS2D/Renderer/Renderer.h>
 
+#include <cstddef>
+#include <limits>
+
 using namespace NAS2D;
 
-Planet::PlanetType PLANET_TYPE_SELECTION = Planet::PlanetType::None;
+std::size_t planetSelection;
+constexpr std::size_t planetSelectionInvalid = std::numeric_limits<int>::max();
```
DanRStevens, line 22: This is mixing `std::size_t` and `int`.

Actually, if you use `auto` for the variable type here, the return type of `max()` should match the `numeric_limits` template type parameter.

OPHD/States/PlanetSelectState.cpp
DanRStevens: Makes sense. I think you're on to something here with this refactor.

When I was reading the changes, I was kind of thinking rather than store individual fields and copy fields one at a time, you could contain the struct and copy the struct as a whole. Could do similar with the accessors, replacing accessors for each attribute with a single accessor to get a `const &` reference to the struct. Struct fields are public, so the caller would be free to read whatever subfields from there, without have to write explicit accessors for all of them.

DanRStevens:
Seems there's now a merge conflict with one of the other recently merged branches.

OPHD/States/PlanetSelectState.cpp
Brett208: good catch.

Brett208:
Not sure why that caused a merge commit, but should be cleanup up now.

PR #523, opened 2020-07-18
AppVeyor auto release on tag

DanRStevens:
AppVeyor will now package EXE and DLL files, and store them as artifacts associated with the build. This happens for every build.

If the build is for a new Git tag, the artifacts will also be uploaded to GitHub as a new release. This happens only for tagged builds.

----

Would appreciate someone testing the release artifacts package to make sure it runs. It will need a copy of the `data/` folder to run with.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -43,3 +43,13 @@ after_build:
 artifacts:
   - path: "*.zip"
     name: BuildArtifacts
+deploy:
+  on:
+    branch: master
+    APPVEYOR_REPO_TAG: true
+    CONFIGURATION: Release
+  artifact: BuildArtifacts
+  description: ''
+  provider: GitHub
+  auth_token:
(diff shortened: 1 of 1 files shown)


DanRStevens:
I managed to get the artifact package + `data/` folder to run on Linux using Wine. That suggests a working package, so I'm going to merge this.

I noticed the audio was a bit garbled and distorted, though I suspect that's from performance problems from running under Wine.

Would appreciate if someone tried out the package plus a copy of the `data/` folder on Windows and report back how things go.

Brett208:
@DanRStevens , I tested the x64 artifacts on Windows 10. After adding the data directory it worked as advertised. Music was not garbled, so must either be a Wine problem or a performance issue on Wine. I wouldn't think playing a music track by itself would tax Wine?

Thanks for adding the artifacts.

PDB symbols seems to be saving nicely as well.

DanRStevens:
Ok, thanks for testing.

Music being garbled might be a combination of Wine performance issues and sound buffer size. I believe the code was pretty explicit about setting a buffer size, which can garble sound if it's too small.

Issue #530, opened 2020-07-18
Accessing Planet::Attributes from a derived structure


Brett208:
I would like to give the structure class access to the Planet::Attributes so various structures such as the solar power panel and solar power plant can understand planet properties. 

I was thinking about passing planentAttributes as a const reference into all Structure constructors, but wasn't sure if that was appropriate. 

-Brett

DanRStevens:
Hmm, I see how it makes sense, but also seems a bit funny to pass that struct to every `Structure` constructor. Maybe it can be limited to the relevant `Structure` subclasses?

Might be worth taking a look at `Structure* StructureCatalogue::get(StructureID type)`:
```cpp
case StructureID::SID_SOLAR_PANEL1:
	structure = new SolarPanelArray();
	break;

case StructureID::SID_SOLAR_PLANT:
	structure = new SolarPlant();
	break;
```

Brett208:
OK, sounds good. I propose moving `StructureCatalog::init()` from program startup to anytime a new game is loaded or saved. This will allow passing `planet::Attributes` into the `StructureCatalog` based on starting game conditions. I think it will also facilitate modifying the recycle and structure cost tables based on completed research when loading a new game in the future. 

I didn't want to move it without bringing it up in case someone had an issue with the architectural change.

DanRStevens:
Makes sense. Sounds like that could be a useful design change.

Brett208:
Currently, saved games do not actually save the `Planet::Type`.  I would need to introduce a breaking change. I would propose going from: 
 
``` XML
<OutpostHD_SaveGame version="0.30">
	<properties sitemap="maps/mars_04" tset="tsets/mars.png" diggingdepth="4" />
	<view_parameters currentdepth="0" viewlocation_x="151" viewlocation_y="36" />
```

to

``` XML
<OutpostHD_SaveGame version="0.31">
	<properties planetType="mars" />
	<view_parameters currentdepth="0" viewlocation_x="151" viewlocation_y="36" />
</OutpostHD_SaveGame>
```

The game would see that a mars type planet was selected and then load the mars attributes as opposed to storing all the attributes inside the saved file. This future proofs saves from changes to planet attributes, both addition of new attributes and changing values of existing attributes like tileset locations.

Any issue if I break saved games to do this shortly? Someone could easily update the standard saved games included in data by hand to keep them working. (Not sure if I have access, but might I think it is stored on the wiki).

DanRStevens:
[...]

PR #531, opened 2020-07-19
Solar panel refactor

Brett208:
This pull request meandered some and ended up completing some prep-work for #530.

### DIFF
--- a/OPHD/Things/Structures/SeedPower.h
+++ b/OPHD/Things/Structures/SeedPower.h
@@ -1,38 +1,20 @@
 #pragma once
 
 #include "Structure.h"
+#include "../../Constants.h"
 #include <string>
 
 const int SEED_POWER_PRODUCTION = 50;
 
-class SeedPower: public Structure
+class SeedPower: public PowerStructure
 {
 public:
-	SeedPower() :
-		Structure(constants::SEED_POWER, "structures/seed_1.sprite", StructureClass::CLASS_ENERGY_PRODUCTION)
+	SeedPower() : PowerStructure(constants::SEED_POWER, "structures/seed_1.sprite", StructureClass::CLASS_ENERGY_PRODUCTION)
 	{
 		sprite().play(constants::STRUCTURE_STATE_CONSTRUCTION);
 		maxAge(150);
 		turnsToBuild(5);
 		requiresCHAP(false);
-	}
-
-	StringTable createInspectorViewTable() override
-	{
-		StringTable stringTable(2, 1);
-
-		stringTable[{0, 0}].text = "Power Produced:";
-
-		auto energyProduced = calculateEnergyProduced();
-
-		stringTable[{1, 0}].text = std::to_string(energyProduced) + " / " + std::to_string(SEED_POWER_PRODUCTION);
-
-		if (energyProduced == 0)
-		{
-			stringTable[{1, 0}].textColor = constants::WARNING_TEXT_COLOR;
-		}
-
-		return stringTable;
 	}
 
 protected:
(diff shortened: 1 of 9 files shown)


OPHD/Things/Structures/Structure.h
```diff
@@ -63,6 +63,7 @@ class Structure: public Thing
 
 public:
 	Structure(const std::string& name, const std::string& spritePath, StructureClass structureClass);
+	virtual ~Structure() {}
```
Brett208, line 66: I think we want virtual destructors in both Structure and PowerStructure since they are base classes?

Brett208:
The solar satellite collector dish still needs updated to use PowerStructure assuming this PR looks good.

OPHD/Things/Structures/Structure.h
DanRStevens: Yes, but technically they already have a virtual destructor inherited from `Thing`:
```cpp
	virtual ~Thing()
	{
		#ifdef _DEBUG
		std::cout << mName << ": He's dead Jim!" << std::endl;
		#endif
	}
```

As such, you don't need to explicitly mention it in derived classes, unless you're creating a specific `override` for it.

OPHD/Things/Structures/SeedPower.h
```diff
@@ -1,49 +1,36 @@
...
-	int calculateEnergyProduced()
+	int calculateEnergyProduced() override
 	{
 		return operational() ? SEED_POWER_PRODUCTION : 0;
 	}
+
+	int calculateMaxEnergyProduction() override
+	{
+		return SEED_POWER_PRODUCTION;
+	}
```
DanRStevens, line 35: You could implement `calculateEnergyProduced` in a generic way in the base class:
```cpp
	int calculateEnergyProduced() override
	{
		return operational() ? calculateMaxEnergyProduction() : 0;
	}
```

OPHD/Things/Structures/Structure.h
Brett208: Ah right, I missed checking Thing. Thanks.

OPHD/Things/Structures/SeedPower.h
Brett208: I thought about this. I didn't know if power plants would eventually produce partial max power if they only had some of the needed resources available to consume or were damaged, etc. so left it as a pure virtual function.

DanRStevens: That could be neat.

Would the behaviour apply to all power plants, or only specific power plants?

DanRStevens:
Looks like there are now some merge conflicts.

Brett208:
I figured that might happen based on the changesets. I'll clean it up.

OPHD/Things/Structures/SeedPower.h
Brett208: In particular, the Fusion Power Plant takes resources, so that was the one I was thinking about. 

Now that I'm thinking about it more, we can provide the default implementation as you suggested and override if needed in the child class. If that was your original suggestion, yet good idea, lets go with that. :)

OPHD/Things/Structures/Structure.h
ldicker83: Still makes sense to declare it explicitly as a method of documentation.

DanRStevens: Mmm, I find it mildly confusing when an explicit override is mentioned, but does nothing. It makes me question what's going on, and I generally go digging to find an implementation, which might not actually exist in any meaningful way. I find it less confusing when all mention is simply dropped.

Plus, given things like `std::unique_ptr`, and it's ability to be used as a base for RAII type objects, very few of our own should have explicit destructors. There are a number of classes whose destructors I would like to optimize away.

PR #544, opened 2020-07-22
Mean solar distance

Brett208:
This was a tough refactor.

I used actual meanSolarDistance values for Mars, Mercury, and Jupiter.

I standardized at 1 AU (Earth) producing 50 solar power. This puts Mars around 35 power, Mercury around 125, and Ganymede at about 9.

This is a bit different value for Mars but I don't think it was really balanced to start with. 

Nothing scientific about the formula, but it seemed to produce a reasonable curve with Mercury being a lot better than Mars and Ganymede likely useless. Please feel free to adjust.

MapViewState is probably not the right place for a lot of the loading and game settings to exist, but didn't really address that with this branch.

### DIFF
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -147,17 +147,19 @@ void MapViewState::load(const std::string& filePath)
 	XmlElement* map = root->firstChildElement("properties");
 	int depth = 0;
 	std::string sitemap;
+	std::string tilesetPath;
 	XmlAttribute* attribute = map->firstAttribute();
 	while (attribute)
 	{
 		if (attribute->name() == "diggingdepth") { attribute->queryIntValue(depth); }
 		else if (attribute->name() == "sitemap") { sitemap = attribute->value(); }
+		else if (attribute->name() == "tset") { tilesetPath = attribute->value(); }
 		attribute = attribute->next();
 	}
 
 	mMapDisplay = Image(sitemap + MAP_DISPLAY_EXTENSION);
 	mHeightMap = Image(sitemap + MAP_TERRAIN_EXTENSION);
-	mTileMap = new TileMap(sitemap, map->attribute("tset"), depth, 0, Planet::Hostility::None, false);
+	mTileMap = new TileMap(sitemap, tilesetPath, depth, 0, Planet::Hostility::None, false);
 	mTileMap->deserialize(root);
 
 	delete pather;
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -157,6 +157,7 @@ void MapViewState::load(const std::string& filePath)
 		attribute = attribute->next();
 	}
 
+	StructureCatalogue::init();
 	mMapDisplay = Image(sitemap + MAP_DISPLAY_EXTENSION);
 	mHeightMap = Image(sitemap + MAP_TERRAIN_EXTENSION);
 	mTileMap = new TileMap(sitemap, tilesetPath, depth, 0, Planet::Hostility::None, false);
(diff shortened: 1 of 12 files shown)


Brett208:
Just added actually saving the mean solar distance. I'm getting the following error when saving a game. The error seems to happen when there are robots present. I didn't test enough to determine if it was caused by this changeset or an underlying problem. x64 debug build.

SaveDebugAssertion

It occurs on the following line of writeRobots

```C++
checkRobotDeployment(robot, robotMap, digger, RobotType::ROBOT_DIGGER);
```

OPHD/States/MapViewStateIO.cpp
```diff
@@ -145,22 +146,20 @@ void MapViewState::load(const std::string& filePath)
...
 	XmlAttribute* attribute = map->firstAttribute();
 	while (attribute)
 	{
-		if (attribute->name() == "diggingdepth") { attribute->queryIntValue(depth); }
-		else if (attribute->name() == "sitemap") { sitemap = attribute->value(); }
-		else if (attribute->name() == "tset") { tilesetPath = attribute->value(); }
+		if (attribute->name() == "diggingdepth") { attribute->queryIntValue(mPlanetAttributes.maxDepth); }
+		else if (attribute->name() == "sitemap") { mPlanetAttributes.mapImagePath = attribute->value(); }
+		else if (attribute->name() == "tset") { mPlanetAttributes.tilesetPath = attribute->value(); }
+		else if (attribute->name() == "meansolardistance") { mPlanetAttributes.meanSolarDistance = std::stof(attribute->value()); }
```
DanRStevens, line 155: This seems a bit inconsistent with how float values are read. Is there a [`queryFloatValue`](https://leethomason.github.io/tinyxml2/classtinyxml2_1_1_x_m_l_attribute.html#a049dea6449a6259b6cfed44a9427b607) in TinyXML 1?

Not really sure if one way should be preferred over the other.

OPHD/Things/Structures/SolarPlant.h
```diff
@@ -25,6 +26,9 @@ class SolarPlant : public PowerStructure
 
 	int calculateMaxEnergyProduction() override
 	{
-		return SOLAR_PLANT_BASE_PRODUCUCTION;
+		return static_cast<int>(SOLAR_PLANT_BASE_PRODUCUCTION / mMeanSolarDistance);
```
DanRStevens, line 29: I suppose if we were looking for mathematical accuracy, we should consider using the [inverse square law](https://en.wikipedia.org/wiki/Inverse-square_law).

----

Need to make sure you're not dividing by 0 here. I believe you set a default value of 0 for that field.

Brett208: I'll add the divide by 0 check.

I'm good with any answer concerning the inverse square law. I don't think it is worth too much thought beyond lets make it fun as there are probably dozens of other variables that could be considered? I was planning on leaving the actual balance for gameplay open and just add the feature. The solar distances are accurate in case we display those on planet selection the data would be correct.

OPHD/States/MapViewStateIO.cpp
Brett208: I appear to only have a query double or int function available via NAS2D. Opened an issue on NAS2D to request feature.

Brett208:
Just validated the debug assertion associated with 'checkRobotDeployment' exists on the master independent of this branch.

Brett208:
See PR #547 for fix to std::map debug assertion. This was caused by undefined behaviour of empty std::map being caught by Visual Studio when compiled in debug mode. I must of had at least one deployed robot when saving in debug mode in the past.

There are probably some warts that could be worked out in this code, but I am ready to merge if looks acceptable.

This will not break save compatibility. If Mean Solar Distance is not contained in the save file, it will leave the value as 0. Power calculation will catch the value of zero and use 1 AU by default. This behaviour is not well tested by me, but should prevent need to rebuild save files. Although inserting the value into the existing saves in the data folder would be simple if someone wishes to pursue.

DanRStevens:
Tagging #530 as related.

PR #547, opened 2020-07-26
Prevent undefined behaviour when serializing robots to Xml

Brett208:
Using an iterator within an empty std::map leads to undefined behaviour as caught by Visual Studio when running code in debug mode. Reference http://www.cplusplus.com/reference/map/map/begin/.

Resolves issue noted in PR #544 but unrelated to changes in that PR.

### DIFF
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -516,6 +516,10 @@ void checkRobotDeployment(XmlElement* _ti, RobotTileTable& _rm, Robot* _r, Robot
 	_ti->attribute("age", _r->fuelCellAge());
 	_ti->attribute("production", _r->turnsToCompleteTask());
 
+	if (_rm.empty()) {
+		return;
+	}
+
 	const auto it = _rm.find(_r);
 	if (it->first == _r)
 	{
@@ -525,7 +529,6 @@ void checkRobotDeployment(XmlElement* _ti, RobotTileTable& _rm, Robot* _r, Robot
 		_ti->attribute("y", position.y);
 		_ti->attribute("depth", tile.depth());
 	}
-
 }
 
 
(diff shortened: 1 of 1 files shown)


OPHD/States/MapViewStateHelper.cpp
```diff
@@ -516,6 +516,10 @@ void checkRobotDeployment(XmlElement* _ti, RobotTileTable& _rm, Robot* _r, Robot
 	_ti->attribute("age", _r->fuelCellAge());
 	_ti->attribute("production", _r->turnsToCompleteTask());
 
+	if (_rm.empty()) {
+		return;
+	}
+
 	const auto it = _rm.find(_r);
 	if (it->first == _r)
```
DanRStevens, line 524: We should probably check validity of `it` after the call to `find`. It's possible we could be looking up an item that doesn't not exist in the collection, even though the collection is non-empty.

According to [`std::map::find`](https://en.cppreference.com/w/cpp/container/map/find), we should compare the iterator against `_rm.end()` before  using it. If it's not equal to `.end()`, it will be valid, and the key will match `_r`. That would also make the check more consistent with other similar checks spread throughout the code.

```cpp
const auto it = _rm.find(_r);
if (it != _rm.end())
{
    // ...
```

With an improved check on the result of `find`, we shouldn't need to check if the collection was empty or not.

DanRStevens:
I made the change I suggested. I'm not too sure how you went about verifying the bug/fix though.

Brett208:
Visual Studio will catch it when the portion of code is called that has an empty map if compiled in debug mode. You would have to actually run the compilation and try to iterate over the empty collection. Unsure if the VS debugger also has to be attached to trigger. I'll test shortly.

Brett208:
Fix works on MSVC. Thanks for finding the better implementation.

DanRStevens:
Thank you for testing.

Issue #14, opened 2018-04-27
Have Mine search for nearest available Smelter


ldicker83:
When a Mine has finished building and activates, it needs to find the nearest Smelter that can accept additional resources (e.g., if a Smelter just has ores piling up it's overloaded).

This should probably be limited to a 30 tile radius around the mine for a few reasons:

1. Reduces search algorithm time to find all available Smelters and determine routes between them.
2. Forces the player to pick and choose between mines so that they can't just drop a miner on every mine on the map and expect to get resources from everything.

The radius limit is an arbitrary number. We can probably assume that a tile is roughly 500x500 feet. That being the case, a 30 tile radius is roughly a 2.88 mile distance. Searching within a 5.76 mile circular area seems reasonable for this kind of a search.

Issue #12, opened 2018-04-27
Implement Truck Routing


ldicker83:
jtremolo and I have bounced around a few ideas via PM about the trucking implementation for OutpostHD. Vomov on the Forums has also posted about it and the general idea behind trucking is this:

When a new mine is set up, it will do a search around itself to find the nearest Smelter. If the Smelter is already full (basically too many mines are already depositing ores into it), it will search for the next nearest smelter. Once it makes determines which Smelter it will send ores to, it will establish a route for trucks to follow. This route will likely use an A* pathing algorithm.

Once established, this route should be checked each turn to determine whether or not the path has become obstructed (due to construction crossing the original path). If that is the case, a new route needs to be determined.

PR #548, opened 2020-07-26
Routing from mines

ldicker83:
This is effectively finished. Next is truck management at mines but that's a different change set from this.

Implements #12 and #14

### DIFF
--- a/OPHD/Map/TileMap.cpp
+++ b/OPHD/Map/TileMap.cpp
@@ -608,10 +608,33 @@ void TileMap::AdjacentCost(void* state, std::vector<micropather::StateCost>* adj
 		Tile* adjacent_tile = getTile(tilePosition + offset, 0);
 		float cost = constants::ROUTE_BASE_COST;
 
-		if (!adjacent_tile || !adjacent_tile->empty() || adjacent_tile->index() == TerrainType::TERRAIN_IMPASSABLE) { cost = FLT_MAX; }
-		else { cost *= static_cast<float>(adjacent_tile->index()) + 1.0f; }
+		if (!adjacent_tile || adjacent_tile->index() == TerrainType::TERRAIN_IMPASSABLE)
+		{
+			cost = FLT_MAX;
+		}
+		else if (adjacent_tile && !adjacent_tile->empty())
+		{
+			if (adjacent_tile == mPathStartEndPair.first || adjacent_tile == mPathStartEndPair.second)
+			{
+				cost *= static_cast<float>(adjacent_tile->index()) + 1.0f;
+			}
+			else
+			{
+				cost = FLT_MAX;
+			}
+		}
+		else
+		{
+			cost *= static_cast<float>(adjacent_tile->index()) + 1.0f;
+		}
 
 		micropather::StateCost nodeCost = { adjacent_tile, cost };
 		adjacent->push_back(nodeCost);
 	}
+}
+
+
+void TileMap::pathStartAndEnd(void* start, void* end)
+{
+	mPathStartEndPair = std::make_pair(start, end);
 }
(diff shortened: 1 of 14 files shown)


OPHD/Map/TileMap.cpp
```diff
@@ -608,10 +608,33 @@ void TileMap::AdjacentCost(void* state, std::vector<micropather::StateCost>* adj
...
-		else { cost *= static_cast<float>(adjacent_tile->index()) + 1.0f; }
+		if (!adjacent_tile || adjacent_tile->index() == TerrainType::TERRAIN_IMPASSABLE)
+		{
+			cost = FLT_MAX;
+		}
+		else if (adjacent_tile && !adjacent_tile->empty())
+		{
+			if (adjacent_tile == mPathStartEndPair.first || adjacent_tile == mPathStartEndPair.second)
+			{
+				cost *= static_cast<float>(adjacent_tile->index()) + 1.0f;
```
DanRStevens, line 619: It's not super clear to me at first glance why `cost` is being multiplied, particularly in an accumulator style way. It would make more sense to me if this was written here as:
```cpp
cost = constants::ROUTE_BASE_COST * static_cast<float>(adjacent_tile->index()) + 1.0f;
```

The code becomes easier to understand when you don't have to go hunting down a previous assignment.

OPHD/Map/TileMap.cpp
```diff
@@ -608,10 +608,33 @@ void TileMap::AdjacentCost(void* state, std::vector<micropather::StateCost>* adj
...
+			}
+			else
+			{
+				cost = FLT_MAX;
+			}
+		}
+		else
+		{
+			cost *= static_cast<float>(adjacent_tile->index()) + 1.0f;
+		}
```
DanRStevens, line 629: Maybe there's a way to combine the conditions leading to these assignments with the previous identical assignments.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -185,14 +185,11 @@ static RouteList findRoutes(micropather::MicroPather* solver, TileMap* tilemap,
...
 
-static bool compareRoute(Route& a, Route& b) { return a.cost < b.cost; }
-
-
 static Route findLowestCostRoute(RouteList& routeList)
 {
 	if (routeList.empty()) { return Route(); }
 
-	std::sort(routeList.begin(), routeList.end(), compareRoute);
+	std::sort(routeList.begin(), routeList.end(), [](const Route& a, const Route& b) { return a.cost < b.cost; } );
```
DanRStevens, line 192: You could potentially find the lowest cost route without sorting. Sorting would have minimum O(n lg(n)) and more likely O(n^2) worst case efficiency, while searching should only require O(n). Assuming you only need the minimum cost, and don't care about the rest, that would improve efficiency.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -162,6 +162,34 @@ void MapViewState::updateMorale()
 }
 
 
+using namespace micropather;
+
+MicroPather* pather = nullptr;
+using Route = std::vector<void*>;
```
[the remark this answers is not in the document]
ldicker83: Seems you answered your own question -- but yeah, it's how MicroPather was designed so I stuck with it. Allows for us to define our own State or Node types

OPHD/States/MapViewStateTurn.cpp
ldicker83: Sure, but this seems like mostly a moot point as it happens during turn calculation.

That said, finding the lowest cost route would still require looking at each route and keeping a running counter. Routes are usually only going to be a handful at a time so it seems like negligible cost. We can refactor as needed if bottle necks are identified during profiling.

OPHD/Map/TileMap.cpp
ldicker83: I would agree, as the comment notes, this can be done better but for whatever reason I'm just not seeing it.

OPHD/Map/TileMap.cpp
ldicker83: See above comment

PR #550, opened 2020-07-27
Store Planet Descriptions in std::map

Brett208:
I think this will be easier to maintain in the long run than current form (which was commented as needing cleaned up). It would be nice to push into an xml file eventually.

### DIFF
--- a/OPHD/States/PlanetSelectState.cpp
+++ b/OPHD/States/PlanetSelectState.cpp
@@ -239,12 +239,14 @@ void PlanetSelectState::onMousePlanetEnter()
 
 	for (std::size_t i = 0; i < mPlanets.size(); ++i)
 	{
-		// FIXME: Ugly, will be difficult to maintain in the future.
 		if (mPlanets[i]->mouseHovering())
 		{
-			if (mPlanets[i]->attributes().type == Planet::PlanetType::Ganymede) { mPlanetDescription.text(constants::PLANET_DESCRIPTION_GANYMEDE); }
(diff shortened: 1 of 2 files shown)


OPHD/States/PlanetSelectState.cpp
```diff
@@ -239,12 +239,14 @@ void PlanetSelectState::onMousePlanetEnter()
...
 		{
-			if (mPlanets[i]->attributes().type == Planet::PlanetType::Ganymede) { mPlanetDescription.text(constants::PLANET_DESCRIPTION_GANYMEDE); }
-			if (mPlanets[i]->attributes().type == Planet::PlanetType::Mars) { mPlanetDescription.text(constants::PLANET_DESCRIPTION_MARS); }
-			if (mPlanets[i]->attributes().type == Planet::PlanetType::Mercury) { mPlanetDescription.text(constants::PLANET_DESCRIPTION_MERCURY); }
+			auto iterator = constants::PLANET_DESCRIPTIONS.find(mPlanets[i]->attributes().type);
+
+			if (iterator != constants::PLANET_DESCRIPTIONS.end())
+			{
+				mPlanetDescription.text(iterator->second);
+			}
```
DanRStevens, line 249: You can potentially shorten this to:
```cpp
mPlanetDescription.text(constants::PLANET_DESCRIPTIONS.at(mPlanets[i]->attributes().type));
```

If the item does not exist in the `std::map`, the `at` method will `throw` and exception. Assuming you want an exception here of course.

Brett208: Ah right, I forgot about the `.at` function. I like the idea of adding an exception message.

PR #565, opened 2020-08-03
StringTable position improvements

Brett208:
In preparation for listing resources consumed by a structure. Would like to list consumed resources to the right of the inspector view table.

### DIFF
--- a/OPHD/UI/StringTable.cpp
+++ b/OPHD/UI/StringTable.cpp
@@ -34,7 +34,7 @@ void StringTable::draw(NAS2D::Renderer& renderer) const
 
 void StringTable::position(NAS2D::Point<int> position)
 {
-	mScreenRect = NAS2D::Rectangle<int>::Create(position, mScreenRect.size());
+	mScreenRect.startPoint(position);
 }
 
 NAS2D::Point<int> StringTable::position() const
--- a/OPHD/UI/StringTable.cpp
+++ b/OPHD/UI/StringTable.cpp
@@ -62,12 +62,12 @@ void StringTable::setDefaultTextColor(NAS2D::Color color)
 	mDefaultTextColor = color;
 }
 
-void StringTable::setHorizontalPadding(float padding)
+void StringTable::setHorizontalPadding(int padding)
 {
 	mHorizontalPadding = padding;
 }
 
-void StringTable::setVerticalPadding(float padding)
+void StringTable::setVerticalPadding(int padding)
 {
 	mVerticalPadding = padding;
 }
@@ -105,10 +105,10 @@ void StringTable::computeRelativeCellPositions()
 	auto columnWidths = computeColumnWidths();
 	auto rowHeights = computeRowHeights();
 
-	float columnOffset = 0;
+	int columnOffset = 0;
 	for (std::size_t column = 0; column < mColumnCount; ++column)
 	{
-		float rowOffset = 0;
+		int rowOffset = 0;
 		for (std::size_t row = 0; row < mRowCount; ++row)
 		{
 			auto cellIndex = getCellIndex(CellCoordinate{column, row});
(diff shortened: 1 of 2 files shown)


OPHD/UI/StringTable.cpp
```diff
@@ -32,7 +32,7 @@ void StringTable::draw(NAS2D::Renderer& renderer) const
 	}
 }
 
-void StringTable::position(NAS2D::Point<int> position)
+void StringTable::position(const NAS2D::Point<int>& position)
```
DanRStevens, line 35: For structs as small as `Point`, it's typically faster to pass by-value. That's the convention we've been using in most places for `Point` and `Vector`. Admittedly, we're not entirely consistent about it. As an added bonus, passing by-value tends to create less syntactic noise.

For larger structs, such as `Rectangle`, we do still tend to pass by `const&`. The difference in speed is not typically very large for structs of these small sizes, so this rule doesn't need to be set in stone. Typically the speed switch over for by-value versus by-reference passing of structs is at about 2-3 `int` sized elements.

OPHD/UI/StringTable.h
```diff
@@ -68,7 +70,7 @@ class StringTable
 	std::vector<CellWithPosition> mCells;
 	const std::size_t mColumnCount;
 	const std::size_t mRowCount;
-	NAS2D::Point<int> mPosition;
+	NAS2D::Rectangle<int> mScreenRect = { };
```
DanRStevens, line 73: `Rectangle` structs default to zero initialization. The explicit initialization here is not needed.

OPHD/UI/StringTable.cpp
```diff
@@ -28,18 +28,23 @@ void StringTable::draw(NAS2D::Renderer& renderer) const
...
 
-		renderer.drawText(*getCellFont(i), cell.text, mPosition + cell.textRelativePosition, textColor);
+		renderer.drawText(*getCellFont(i), cell.text, position() + cell.textRelativePosition, textColor);
 	}
 }
 
 void StringTable::position(const NAS2D::Point<int>& position)
 {
-	this->mPosition = position;
+	mScreenRect = NAS2D::Rectangle<int>::Create(position, mScreenRect.size());
```
DanRStevens, line 37: This could be written as:
```cpp
mScreenRect.startPoint(position);
```

OPHD/UI/StringTable.cpp
```diff
@@ -115,6 +120,16 @@ void StringTable::computeRelativeCellPositions()
...
 	}
+
+	if (mCells.size() == 0)
+	{
+		mScreenRect.size({ 0, 0 });
+	}
+	else 
+	{
+		mScreenRect.width = static_cast<int>(mCells.back().textRelativePosition.x + columnWidths.back());
+		mScreenRect.height = static_cast<int>(mCells.back().textRelativePosition.y + rowHeights.back());
```
DanRStevens, line 131: This could potentially be written as:
```cpp
const auto lastCellSize = NAS2D::Vector{columnWidths.back(), rowHeights.back()}.to<int>();
mScreenRect.size(mCells.back().textRelativePosition.to<int>() + lastCellSize);
```

If cell position and sizes were converted to `int`, the above could have the type conversions removed. Recently most controls were converted to use `int` based coordinates, so that may make sense to do. That is particularly true for controls containing text, as off pixel values may cause the text to appear fuzzy.

Additionally, the name `textRelativePosition` could be simplified to `textOffset`. Normally a "position" is a `Point`, while an "offset" is a `Vector`. The data type here is `Vector`.

OPHD/UI/StringTable.cpp
Brett208: OK, note fix in PR #567 for Control.h. Will revert this change shortly.

Brett208:
*** Force Push to remove const ref passing of position argument commit ***

OPHD/UI/StringTable.cpp
Brett208: Name update looks good. I like the proposed shorter name of 'textOffset'. 

Not my fault someone switched all the code to int instead of float for positions after StringTable's creation. :)

I'll look at using int for `textOffset`. If it gets too big, we can move to a separate branch, but think it will be easier.

Brett208:
Thanks @DanRStevens for the quality code review. pushed changes should have addressed the issues, assuming the automated build passes.

PR #566, opened 2020-08-03
Scope enum StructureState

Brett208:
A couple of interesting changes were required to prevent some casting. I think it all improves readability.

This changeset will likely cause merge conflict in other uncommitted branches due to the prolific use of StructureState.

I wasn't sure it was best to move StructureState out of Structure.h, but I didn't think it was wise for Structure.h to reference Common.h and Common.h to reference Structure.h at the same time. And I'm not sure you are allowed to forward declare enum class? If this move isn't palatable, maybe there is an alternate solution.

### DIFF
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -50,21 +50,21 @@ Difficulty DifficultyFromString(std::string difficultyStr)
 
 std::map<StructureState, Color> STRUCTURE_COLOR_TABLE
 {
-	{ StructureState::UNDER_CONSTRUCTION, Color{150, 150, 150, 100} },
-	{ StructureState::OPERATIONAL, Color{0, 185, 0} },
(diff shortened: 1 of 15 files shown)


Brett208:
Ah right. I had mixed up which could be forward declared. I'll try to forward declare before merging.

Brett208:
Apparently you cannot forward declare an enum class that is the member of a class...

Brett208:
Tagging #319

DanRStevens:
Huh, Interesting. Seems the scope nesting does pose problems for enum forward declarations. I had to try it just to see what happens. And of course, check StackOverflow :grin: 
[Is in-class enum forward declaration possible?](https://stackoverflow.com/questions/27019292/is-in-class-enum-forward-declaration-possible)

Apparently it really isn't possible.

Anyway, looks like you found a pretty good solution.

PR #593, opened 2020-08-10
Structure inspector improvements

Brett208:
Couple of notes:

- I created a base class called OreRefining for the SeedSmelter and Smelter. I used that to combine the StringTable results in the StructureInspector. I didn't look at further code sharing opportunites. Not sure the name is the best for the base class, but something to start with.
- I was having trouble figuring out how to pull the current capacity of the Storage Tanks and Smelters, so I just displayed Max Capacity for now. It looks like resource pools are continuing to get overhauled, so maybe better to wait until that settles anyways?
- The smelters should include their conversion rates of the different minerals and planned conversion for the next turn. This may lead to a dedicated smelter inspector window instead of trying to use the generic StructureInspector similar to a mine or warehouse
- I was having trouble figuring out how to display if a structure is actually powered or not, so I left it as just power requirement. I didn't look too hard either though.

### DIFF
--- a/OPHD/UI/StructureInspector.cpp
+++ b/OPHD/UI/StructureInspector.cpp
@@ -142,6 +142,9 @@ void StructureInspector::update()
 	auto position = mRect.startPoint() + NAS2D::Vector{ 5, 25 };
 	drawLabelAndValue(position,"Type: ", mStructureClass);
 
+	position.y += 20;
+	drawLabelAndValue(position, "Power Required: ", std::to_string(mStructure->energyRequirement()));
+
 	position = mRect.startPoint() + NAS2D::Vector{190, 25};
 	drawLabelAndValue(position,"State: ", structureStateDescription(mStructure->state()));
 
--- a/OPHD/UI/StructureInspector.cpp
+++ b/OPHD/UI/StructureInspector.cpp
@@ -164,7 +164,4 @@ void StructureInspector::update()
 	stringTable.computeRelativeCellPositions();
 	stringTable.position(mRect.startPoint() + NAS2D::Vector<float>{10, 135});
 	stringTable.draw(renderer);
-
-	position = mRect.startPoint() + NAS2D::Vector{5, mRect.height - FONT->height() - 5};
-	renderer.drawText(*FONT, "This window is a work in progress", position, NAS2D::Color::White);
 }
--- a/OPHD/UI/StructureInspector.cpp
+++ b/OPHD/UI/StructureInspector.cpp
@@ -132,15 +132,14 @@ void StructureInspector::update()
 
 	auto& renderer = Utility<Renderer>::get();
 
-	auto position = mRect.startPoint() + NAS2D::Vector{5, 25};
 	if (mStructure == nullptr)
 	{
-		drawLabelAndValue(position, "NULLPTR!", "");
+		text("NULLPTR!");
 		return;
 	}
-	drawLabelAndValue(position, mStructure->name(), "");
+	text(mStructure->name());
 
-	position.y += 20;
+	auto position = mRect.startPoint() + NAS2D::Vector{ 5, 25 };
 	drawLabelAndValue(position,"Type: ", mStructureClass);
 
 	position = mRect.startPoint() + NAS2D::Vector{190, 25};
--- a/OPHD/Things/Structures/OreRefining.h
+++ b/OPHD/Things/Structures/OreRefining.h
@@ -1,0 +1,31 @@
+#pragma once
+
+
+#include "Structure.h"
+#include <string>
+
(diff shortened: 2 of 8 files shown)


ldicker83:
Apologies for the lengthy reply, there's a bunch to respond to here.

> I was having trouble figuring out how to pull the current capacity of the Storage Tanks and Smelters

The StorageTanks don't currently store anything. Instead, the game determines what it can store by counting up the number of StorageTanks, multiplying by 500 (the current default capacity) and adding 250 to it (CC capacity). This is done in MapViewState.cpp at [line 230](https://github.com/OutpostUniverse/OPHD/blob/master/OPHD/States/MapViewState.cpp#L230):

```cpp
int MapViewState::totalStorage(Structure::StructureClass structureClass, int capacity)
{
	int storageCapacity = 0;

	// Command Center has a limited amount of food storage for when colonists first land.
	if (ccLocation() != CcNotPlaced)
	{
		storageCapacity += constants::BASE_STORAGE_CAPACITY;
	}

	const auto& structures = Utility<StructureManager>::get().structureList(structureClass);
	for (auto structure : structures)
	{
		if (structure->operational() || structure->isIdle())
		{
			storageCapacity += capacity;
		}
	}

	return storageCapacity;
}
```

I do want to take advantage of the updated StorableResources introduced in #583 PR. I opened a new issue for this here #595.

Admittedly this is a naive way of doing this hence the above statement.

Smelters have two 'queues' or 'pools' of StorableResources: _storage_ and _production_. Mined ore is moved into _production_ from `MineFacility`'s, and refined ore is moved into _storage_. It's a little unintuitive at first because it's not well documented. My apologoes for that.

As for 'capacity', this is defined as a hard-coded constant in the individual structures. E.g., SeedSmelter has a hard limit of 500 units of ore, Smelter has a hard limit of 800. The idea is that each 'capacity' is divided into four equal storage 'buckets'. So the SeedSmelter has four buckets (one for each `StorableResources` type) of 125 each for a total of 500. Smelter has a limit of 200 per bucket for a total of 800. Getting the current level and remaining capacity is done using the new `StorableResources` class and its related operators:

```cpp

auto& stored = smelter->storage();
int currentLevel = 0;
for(size_t i = 0; i < stored.resources.size(); ++i) { currentLevel += stored.resources[i]; }
int remainingCapacity = smelter->maxCapacity() - currentLevel;
```

This is somewhat contrived code and shows how an OreRefining class can provide a few interface functions such as `int currentLevel() const;` and 'int maxCapacity() const;' to make things a bit clearer.

OreRefining works fine for me so no complaints here. I wanted to do something similar so you're saving me some time here. 👍 

> The smelters should include their conversion rates of the different minerals and planned conversion for the next turn. This may lead to a dedicated smelter inspector window instead of trying to use the generic StructureInspector similar to a mine or warehouse

Generally agreed here. I don't know if I want to have a specific SmelterInspector window -- not sure it's necessary though to be fair the StructureInspector window is very generic and was mostly intended for debug purposes.

> I was having trouble figuring out how to display if a structure is actually powered or not, so I left it as just power requirement. I didn't look too hard either though.

Structures are not explicitly flagged as powered. Instead, during the turn processing, the StructureManager will iterate through all of the structures and deduct the energy requirements for the structure from the 'energy pool'. If there is insufficient energy to operate a structure, the structure is marked as `DisabledReason::Energy`. This is done in `StructureManager::update()` and `StructureManager::updateStructures()` at [line 177](https://github.com/OutpostUniverse/OPHD/blob/master/OPHD/StructureManager.cpp#L177):

```cpp
if (structure->energyRequirement() > totalEnergyAvailable())
{
	structure->disable(DisabledReason::Energy);
	continue;
}
```

OPHD/UI/StructureInspector.cpp
```diff
@@ -132,15 +132,14 @@ void StructureInspector::update()
 
 	auto& renderer = Utility<Renderer>::get();
 
-	auto position = mRect.startPoint() + NAS2D::Vector{5, 25};
 	if (mStructure == nullptr)
 	{
-		drawLabelAndValue(position, "NULLPTR!", "");
+		text("NULLPTR!");
```
ldicker83, line 137: Wondering if this should be treated as an exception? Would be a different change set but something to note.

Brett208: I agree an exception would be better. I can make that change in near future.

Brett208:
I can take a look at the 'OreRefining' base class some more unless you are planning to work it. Thank you for the detailed explanation on the different questions I had.

ldicker83:
I was thinking about it but I have a few other things I want to work on first (like some crashing and work on the XML load/save code) before I go into it.

PR #605, opened 2020-08-10
Throw runtime exception if pointer to structure is null inside StructureInspector

Brett208:


ldicker83:
Was thinking of doing exactly this, thanks!

Issue #596, opened 2020-08-10
Food level is always at 0 in UI HUD


ldicker83:
This is due to a change in the way food storage is handled. This will need to be updated to account for the changes to resource storage introduced in #583.

PR #606, opened 2020-08-10
Smelter improvements

Brett208:
I'm finding a bit difficult to test because all my colonists are dying the turn they land on the colonist lander. I'm assuming this is because the cargo lander doesn't seem to be providing food anymore?

Implements changes discussed in PR #593 

Unsure if issue #596 means that there is no food to draw from or if it is just a display bug.

### DIFF
--- a/OPHD/Things/Structures/OreRefining.h
+++ b/OPHD/Things/Structures/OreRefining.h
@@ -18,12 +18,28 @@ public:
 
 	StringTable createInspectorViewTable() override
 	{
-		StringTable stringTable(2, 1);
+		StringTable stringTable(2, 4);
+		
+		stringTable.setColumnText(
+			0,
+			{
+				"Common Metals Storage:",
+				"Rare Metals Storage:",
+				"Common Minerals Storage:",
+				"Rare Minerals Storage:"
+			});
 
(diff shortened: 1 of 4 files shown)


ldicker83:
Right, the food thing I figured would be a problem after merging #583. It's something I'll be working on through Wednesday (some tonight I hope). In the mean time, it's not a great way to handle it, but could either modify the food values directly in... crud. Might need to add a food value to the CC now that I think about it.

Bleh, I knew this would be a bit of a tough change with far reaching effects. Still, I think it's for the better.

OPHD/Things/Structures/OreRefining.h
```diff
@@ -28,4 +28,6 @@ class OreRefining : public Structure
 
 protected:
 	virtual int calculateMaxStorage() = 0;
+
+	std::array<int, 4> OreConversionDivisor{ 2, 2, 3, 3 };
```
ldicker83, line 56: The idea here was to allow for tweaking in derived classes but this change still makes sense as it provides a base conversion rate that can still be overridden in classes inheriting from OreRefining. Commenting on this mostly for the sake of documenting design decisions.

PR #607, opened 2020-08-11
Update food level read write

### LINKED ISSUE #584: Adjust Agridome savegame read/write
The original way Agridome food was saved was in the ResourcePool class. Since that's been removed in PR #583, this will need to be saved/read a different way.

ldicker83:
Closes #584 and #596

### DIFF
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -205,6 +205,7 @@ void MapViewState::load(const std::string& filePath)
 
 	updateRobotControl(mRobotPool);
 	updateResidentialCapacity();
+	updateFood();
 
 	if (mTurnCount == 0)
 	{
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -403,6 +403,20 @@ void MapViewState::readStructures(Xml::XmlElement* element)
 			static_cast<SeedLander*>(st)->position({x, y});
 		}
 
+		if (type_id == StructureID::SID_AGRIDOME)
+		{
+			auto agridome = static_cast<Agridome*>(st);
+
+			auto foodStorage = structure->firstChildElement("food");
+			if (foodStorage == nullptr)
+			{
+				throw std::runtime_error("MapViewState::readStructures(): Agridome saved without a food level node.");
+			}
+
+			auto foodLevel = foodStorage->attribute("level");
+			agridome->foodLevel(std::stoi(foodLevel));
+		}
+
 		st->age(age);
 		st->forced_state_change(static_cast<StructureState>(state), static_cast<DisabledReason>(disabled_reason), static_cast<IdleReason>(idle_reason));
 		st->connectorDirection(static_cast<ConnectorDir>(direction));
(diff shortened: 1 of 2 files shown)


OPHD/States/MapViewStateIO.cpp
```diff
@@ -403,6 +403,20 @@ void MapViewState::readStructures(Xml::XmlElement* element)
...
+
+			auto foodStorage = structure->firstChildElement("food");
+			if (foodStorage == nullptr)
+			{
+				throw std::runtime_error("MapViewState::readStructures(): Agridome saved without a food level node.");
+			}
+
+			auto foodLevel = foodStorage->attribute("level");
+			agridome->foodLevel(std::stoi(foodLevel));
+		}
```
DanRStevens, line 424: I feel like this has more XML tag nesting than is required. This implies a parent context such as:
```
<agridome>
  <food level="0" />
</agridome>
```

It might be simpler to have it as:
```
<agridome food="0" />
```

Though I am aware the parent tag might not be named "agridome", and I saw elsewhere you were planning to use numeric structure type IDs.

OPHD/States/MapViewStateIO.cpp
```diff
@@ -194,7 +194,8 @@ void MapViewState::load(const std::string& filePath)
 	readPopulation(root->firstChildElement("population"));
 	readTurns(root->firstChildElement("turns"));
 
-	XmlElement* energy = root->firstChildElement("energy");
+	auto energy = root->firstChildElement("energy");
+	if (!energy) { throw std::runtime_error("MapViewState::load(): Savegame file is missing '<energy>' tag."); }
```
DanRStevens, line 202: Similarly, a flatter structure seems possible here.

OPHD/States/MapViewStateIO.cpp
ldicker83: I agree. I want to make that a separate PR and just include food level as an attribute of structures without making it an actual element with a single attribute. Kinda like what the factories do where they simply attach `production_completed` and `production_type` type attributes to a structure node:

```xml
<structure x="164" y="31" depth="1" age="31" state="1" forced_idle="0" disabled_reason="0" idle_reason="0" type="Residential Facility" direction="1" pop0="0" pop1="0" /> <!-- standard structure -->
<structure x="164" y="33" depth="0" age="58" state="1" forced_idle="0" disabled_reason="0" idle_reason="0" type="SEED Factory" direction="1" pop0="2" pop1="0" production_completed="3" production_type="2" /> <!-- factory -->
<structure x="168" y="31" depth="0" age="50" state="1" forced_idle="0" disabled_reason="0" idle_reason="0" type="Agricultural Dome" direction="1" pop0="1" pop1="0" food_level="0" /> <!-- simplified agridome -->
<structure x="164" y="35" depth="0" age="9" state="1" forced_idle="0" disabled_reason="0" idle_reason="0" type="Robot Command" direction="1" pop0="4" pop1="0" robots="4,5,6" /><!-- simplified RCC -->
```

I noticed looking at a savegame that we could also simplify RCC's in the same way. Structures that actually use a storage and production pool makes sense to have the nested elements for each 'pool' and attributes for the associated resources.

OPHD/States/MapViewStateIO.cpp
ldicker83: Probably. Currently save games look like this:

```xml
<turns count="58" />
<energy energy="100" />
```

Which seems very inefficient to me. This could be done with a single element and a few attributes attached.

PR #626, opened 2020-08-15
Add structure name table

### LINKED ISSUE #619: Add Structure String Table
From #597:

> A string table will need to be defined with a numeric identifier as a key (e.g., StructureID enum). Traditionally I've done this as a std::map but I think a std::array will do the trick here especially if the table lookup is done through a function that returns a std::string and requires a class enum.

ldicker83:
Closes #619

### DIFF
--- a/OPHD/Things/Structures/Structure.cpp
+++ b/OPHD/Things/Structures/Structure.cpp
@@ -60,6 +60,54 @@ const std::string& structureClassDescription(Structure::StructureClass _class)
 }
 
 
+
+static std::array<std::string, StructureID::SID_COUNT> StructureNameTable =
+{
+	constants::AGRIDOME,
+	constants::AIR_SHAFT,
+	constants::CARGO_LANDER,
+	constants::CHAP,
+	constants::COLONIST_LANDER,
+	constants::COMMAND_CENTER,
+	constants::COMMERCIAL,
+	constants::COMM_TOWER,
+	constants::FUSION_REACTOR,
+	constants::HOT_LABORATORY,
+	constants::LABORATORY,
+	constants::MEDICAL_CENTER,
+	constants::MINE_FACILITY,
+	constants::MINE_SHAFT,
+	constants::NURSERY,
+	constants::PARK,
+	constants::RECREATION_CENTER,
+	constants::RED_LIGHT_DISTRICT,
+	constants::RESIDENCE,
+	constants::ROBOT_COMMAND,
+	constants::SEED_FACTORY,
+	constants::SEED_LANDER,
+	constants::SEED_POWER,
+	constants::SEED_SMELTER,
+	constants::SMELTER,
+	constants::SOLAR_PANEL1,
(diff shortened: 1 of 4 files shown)


OPHD/Things/Structures/Structure.cpp
```diff
@@ -60,6 +60,54 @@ const std::string& structureClassDescription(Structure::StructureClass _class)
 }
 
 
+
+static std::array<std::string, StructureID::SID_COUNT> StructureNameTable =
```
Brett208, line 64: If StructureNameTable was made a std::map<StructureID, std::string> instead of a std::array, it would facilitate making StructureID an enum class.

I assume std::map is slightly slower than an array lookup, but I'm also guessing the performance here isn't critical enough to matter.

-Brett

Brett208:
@ldicker83, I would recommend adding the following headers. I suspect they are getting included from other headers, but maybe better to self document their use.

**Structure.h Add:**
* std::string

**Structure.cpp Add:**
* std::array (maybe not needed if my suggestion is taken on other comment)
* std::map

OPHD/Things/Structures/Structure.cpp
ldicker83: I'd considered that -- it would also make it easy to add new structure names without having to ensure order is correct. @DanRStevens seemed to suggest otherwise -- I personally prefer the std::map version but std::array is technically faster.

I'm okay with either implementation, really. It's why I opted for an accessor function so that internal implementation can be changed without needing to make changes elsewhere in the code.

DanRStevens: You're both quite correct in your observations. Using a `std::map` would allow the conversion to be done without using `static_cast`. The performance difference between `std::map` and `std::array` is expected to be quite small, and of no consequence here. The accessor function does hide the `static_cast` in a single method, preventing spread to other code. Assuming the `StructureName` method is used for conversion, the `StructureID` can be upgraded to an `enum class` without adding additional `static_cast` conversions. Effectively higher efficiency can be achieved with low source code cost. I also expect many current lookups can be removed with future refactoring (general code observation, not specific to this enum).

```cpp
std::string StructureName(StructureID id)
{
	return StructureNameTable[static_cast<size_t>(id)];
}
```

----

Btw, why not lowercase for the initial letter in `StructureName`? Other global functions seem to use lowerCameCase.

----

The current version of `StructureName` is not safe for out of range `id` values. Using `.at(index)` instead of `operator[index]` would ensure safety, causing an exception to be raised if an out of bounds `id` value was passed in. Additionally, a manual bounds check can be inserted if a more detailed error message is desired.

```cpp
std::string StructureName(StructureID id)
{
	return StructureNameTable.at(static_cast<size_t>(id));
}
```

DanRStevens:
Brett appears to be correct about header inclusion. We should add direct includes for `<array>`, `<map>`, and `<string>`, rather than rely on indirect inclusion from other included headers.

PR #627, opened 2020-08-15
Further StructureInspector improvements

Brett208:


OPHD/Things/Structures/Structure.h
```diff
@@ -101,6 +101,7 @@ class Structure: public Thing
 
 	// ATTRIBUTES
 	StructureClass structureClass() const { return mStructureClass; }
+	const std::string& classDescription();
```
ldicker83, line 156: Should this maybe be a static function? Something to consider for future improvements?

DanRStevens: The old method could have been made a `static` member:
```cpp
const std::string& structureClassDescription(Structure::StructureClass _class)
```

However, the new method removes the parameter, and instead uses the class member variable `mStructureClass`. As such the new member function could not be marked `static`.

```cpp
const std::string& Structure::classDescription()
{
	return STRUCTURE_CLASS_TRANSLATION[mStructureClass];
}
```

DanRStevens: Though it could be marked as `const`.

Brett208: It looks the member function `ClassDescription' was reverted to a freestanding static function during the merge conflict resolution. Based on comments above, I think that was accidental.

I will open a new branch, make the class instance function const as recommended by @DanRStevens. I'll also move the static function into the class but leave in place as requested by @ldicker83. This will allow the shorthand calling if you already have an instance and flexibility to call if you do not have an instance of a Structure. 

The missing const on the function was an oversight on my part, thanks for catching.

PR #625, opened 2020-08-15
StringTable and StructureInspector improvements

Brett208:
Closes #621 

Spent a lot of time trying to keep commits clean, but the one where StringTable is switched to in StructureTable didn't turn out clean.

PR #631, opened 2020-08-17
Structure&structure inspector improvements reapply

Brett208:
Rework from #627 

Style notes:
* Do we want to have multiple calls to public, protected, private in headers to distinguish different categories of class members
   * I removed multiple calls to private from `StructureInspector`, but that doesn't seem in line with the rest of the code base

```C++
private:
virtual functions

private:
instance functions

private:
class variables
```

* Some function arguments include an _ prefix
   * I removed this prefix in a couple of places. Is that ok, or do we want to standardize one way or the other.

Thanks,
Brett

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -427,8 +427,6 @@ void MapViewState::nextTurn()
 
 	checkColonyShip();
 
-	mStructureInspector.check();
-
 	// Check for Game Over conditions
 	if (mPopulation.size() < 1 && mLandersColonist == 0)
 	{
--- a/OPHD/Things/Structures/OreRefining.h
+++ b/OPHD/Things/Structures/OreRefining.h
@@ -18,20 +18,34 @@ public:
 
 	StringTable createInspectorViewTable() override
 	{
-		StringTable stringTable(2, 4);
-		
+		StringTable stringTable(3, 5);
+
+		stringTable.setColumnFont(0, stringTable.GetDefaultFont());
+		stringTable.setRowFont(0, stringTable.GetDefaultTitleFont());
+		stringTable.setHorizontalPadding(20);
+
 		stringTable.setColumnText(
 			0,
 			{
-				"Common Metals Storage:",
-				"Rare Metals Storage:",
-				"Common Minerals Storage:",
-				"Rare Minerals Storage:"
+				"",
+				"Common Metal",
+				"Rare Metal",
+				"Common Minerals",
+				"Rare Minerals"
+			});
+
+		stringTable.setRowText(
+			0,
+			{
+				"Material",
+				"Storage",
+				"Ore Conversion Rate"
 			});
 
 		for (std::size_t i = 0; i < storage().resources.size(); ++i)
 		{
-			stringTable[{1, i}].text = writeStorageAmount(storage().resources[i]);
+			stringTable[{1, i + 1}].text = writeStorageAmount(storage().resources[i]);
+			stringTable[{2, i + 1}].text = std::to_string(OreConversionDivisor[i]) + " : 1";
 		}
 
 		return stringTable;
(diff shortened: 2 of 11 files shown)


Brett208:
I didn't have time to re-remove StructureInspector::check and the TextArea, but should be able to circle back around this afternoon.

Wanted to push so far to get input on some of the style items above.

-Brett

ldicker83:
It's a stylistic choice I used tp help me distinguish between different types of declarations in classes -- e.g., variables, functions, using directives, etc.

I have no particular preference mostly because I don't want it to become a religious code debate (we all know how that goes). I know it's a bit contrary to standard form but ultimately I don't really care which style is used.

Brett208:
I appreciate avoiding too rigorous of coding standards. However, I'll change it back to use multiple private calls for consistency here with the rest of the code base.

Brett208:
It looks like changes within StructureInspector made in PR #625 were also wiped out by the merge. Damage is more extensive than I first thought. 

I plan to combine those commits into 1 or 2 changes and maybe try to reference the original commits? Fortunately, it is all fresh in my head...

Brett208:
Still more work to go but taking a break

Brett208:
OK, at the end of this PR's work.

Turned out the bulk of PR #625 & #627 were erased by the merge conflict resolution. I didn't realize the full extent of the removal when starting out or maybe would have looked for a better solution. Maybe there is a way to undelete branches and reapply them. 

I made a couple of small improvements along the way though so wanted to keep them in. Mostly style or const keywords that I had missed, etc.

I'm going to go ahead and merge since the code had been partially reviewed in the past. I was practicing rebase as appropriate and since I was pushing to the network it caused a lot of reporting noise on this page. Sorry about that.

-Brett

DanRStevens:
Hmm, concerning that we lost changes during a merge conflict resolution. Might be good to understand how that came about. Git can be a little hard to understand in corner cases like this.

Looking at history, I did note a few anomalies.
1f11cd3d Appears to be the base commit on `master` that is ancestor to both #625 and #627.
d2455e7f Appears right after, which appears to have been committed directly to `master`. There is no associated PR for this commit.
8b66386d Seems to be the problem merge, the first ancestor of which is d2455e7f.

Curiously the problem merge has a commit message that is formatted differently than usual:
> 8b66386d Merge branch 'master' of https://github.com/OutpostUniverse/OPHD

It looks like this commit was done locally, rather than as a branch merge on GitHub. Contrast it with the commit messages for the two PRs that seemed to have been undone (which were branch merges on GitHub):

> 79a63a1c Merge pull request #627 from OutpostUniverse/FurtherStructureInspectorImprovements
> a637f6e3 Merge pull request #625 from OutpostUniverse/StringTableAndStructureInspectorImprovements


Based on the above, we may also want to take a close look at #626, which was also merged to the same history as #625 and #627:
> c1c9de0e Merge pull request #626 from OutpostUniverse/AddStructureNameTable

----

Admitted, I don't know how this sequence of events lead to the observed outcome. I would assume Git should provide reasonable warnings in such a case, though it can sometimes be a bit hard to understand Git.

In terms of recommendations, maybe we should strive to always use GitHub branch merges, and avoid committing directly to `master`. I know it can be easy to forget to create a new branch, and commits sometimes accidentally get added to `master`. In that case, you can create a new branch after the fact, which will then hold the new commits. The `master` branch can then be unwound to the GitHub's version of master `origin/master`:
```
git checkout master
git reset --hard origin/master
```

If you pull, you should always expect to get a **fast-forward** update. If for some reason you get a merge commit, or worse, a merge conflict, that's probably a sign of divergent `master` histories. In that case, abort any merge conflicts, copy the changes to a branch, and reset the local `master` branch to `origin/master`. You are then free to fixup any divergent history on the branch you created earlier, perhaps using `git rebase master`.

----

@ldicker83, did you notice any strange Git messages that you can remember? Maybe mention of merge conflicts? Was there anything unusual about doing the merge conflict?

PR #634, opened 2020-08-18
Tube age

### LINKED ISSUE #628: Tube age remains at 0 of 400
As viewed from the StructureInspector, tube age always remains at 0. 

Are tubes supposed to age like other structures, or are they supposed to remain forever? I can N/A the age in the inspector view if they are not supposed to age.

-Brett

Brett208:
Closes #628

### DIFF
--- a/OPHD/Things/Structures/Structure.h
+++ b/OPHD/Things/Structures/Structure.h
@@ -110,6 +110,7 @@ public:
 	int turnsToBuild() const { return mTurnsToBuild; }
 	int age() const { return mAge; }
 	int maxAge() const { return mMaxAge; }
+	bool ages() const { return maxAge() > 0; }
 	int energyRequirement() const { return mEnergyRequirement; }
 
 	// FLAGS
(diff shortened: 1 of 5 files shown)


OPHD/Things/Structures/Structure.h
```diff
@@ -110,6 +110,7 @@ class Structure: public Thing
 	int turnsToBuild() const { return mTurnsToBuild; }
 	int age() const { return mAge; }
 	int maxAge() const { return mMaxAge; }
+	bool ages() const { return maxAge() > 0; }
```
Brett208, line 113: I couldn't think of a better name. Oftentimes good to include is or has prefix for function names that return a bool.

DanRStevens: Indeed. Can't think of a suggestion for this one.

Brett208:
Dan, thanks for reviewing the pull requests. Always appreciated.

Brett

PR #666, opened 2020-08-26
Refactor ListBoxItem

DanRStevens:
Reference : #479

A bit of prep-work to consolidate how `ListBoxItem` is treated.

### DIFF
--- a/OPHD/UI/StructureListBox.cpp
+++ b/OPHD/UI/StructureListBox.cpp
@@ -34,7 +34,7 @@ static void drawItem(Renderer& renderer, StructureListBox::StructureListBoxItem&
 
 	renderer.drawBox(NAS2D::Rectangle{x + 2, y + 2 - offset, w - 4, LIST_ITEM_HEIGHT - 4}, structureColor);
 
-	renderer.drawText(*MAIN_FONT_BOLD, item.Text, NAS2D::Point{x + 5, ((y + 15) - MAIN_FONT_BOLD->height() / 2) - offset}, structureTextColor);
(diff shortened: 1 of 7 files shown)


OPHD/UI/StructureListBox.cpp
```diff
@@ -41,10 +41,10 @@ static void drawItem(Renderer& renderer, StructureListBox::StructureListBoxItem&
 
 
 StructureListBox::StructureListBoxItem::StructureListBoxItem(Structure* s) :
-	ListBoxItem(s->name()),
-	structure(s),
-	structureState(""),
-	colorIndex(s->state())
+	ListBoxItem{s->name()},
```
ldicker83, line 44: What's the benefit of this change?

DanRStevens: Ahh, so I've been using the `{}` style in most other parts of the project, so partly that was for consistency, though there are some benefits to the newer syntax.

The brace init list `{}` works with either a constructor, or with a `struct` which has no constructor, where the values would then be taken to initialize the fields of the struct, in order. The function call init style `()` will only work with a constructor. If all the constructor does is set struct fields, in order, then changing the init style to `{}` allows you to potentially move the explicit constructor.

I believe I was planning to convert the `ListBoxItem` structs to be plain `struct` without a constructor. I don't see such a change in this particular change set, though I believe that was something I was moving towards when I made that change.

ldicker83: Works for me, thanks!

PR #671, opened 2020-08-26
Refactor Label Control

Brett208:
Preliminary work before actually filling in TextControl fields


OPHD/UI/Core/Label.cpp
```diff
@@ -50,11 +50,6 @@ void Label::color(const NAS2D::Color& color)
 }
 
 
-int Label::textWidth() const
-{
-	return TXT_FONT->width(text());
-}
```
DanRStevens, line 56: I think this is confusing the `Control` size (`Rectangle`), to which clipping should be done (though probably isn't), and the natural size of the assigned text, which may be larger (which might cause clipping) or smaller (with extra padded space). This value could be used to resize a `Label` `Control` so that it fits snugly around the text that it contains.

OPHD/UI/Core/Label.cpp
```diff
@@ -22,7 +22,7 @@ Label::Label(std::string newText)
 
 void Label::autoSize()
 {
-	size(textSize() + NAS2D::Vector{FIELD_PADDING * 2, FIELD_PADDING * 2});
+	size(TXT_FONT->size(text()) + NAS2D::Vector{FIELD_PADDING * 2, FIELD_PADDING * 2});
```
DanRStevens: Here `autoSize` is meant to set the size of the `Control` based on the assigned text, with space for the internal padding. This can be used to ensure text is drawn without any clipping, and so surrounding controls can be placed in accordance with the natural text size.

OPHD/UI/Core/Label.h
```diff
@@ -39,6 +39,5 @@ class Label : public TextControl
 
 private:
 	NAS2D::Color textColor{NAS2D::Color::White};
-	const NAS2D::Font* mFont;
```
DanRStevens: This looks to be an accidental change. The commit message doesn't match up with the change.

OPHD/UI/Core/Label.cpp
Brett208: Label currently always resizes the control based on the size of the text. I wasn't looking to change that behavior here but also wasn't commenting on a preference on clipping or auto resizing. 

I'm unsure what action to take on this comment. Are you looking to have the function textWidth restored?

OPHD/UI/Core/Label.cpp
Brett208: Would you prefer the function name be switched back to autoSize? I had assumed Label's intention was to always resize based on provided text.

OPHD/UI/Core/Label.h
Brett208: Yeah, that was an accidental delete. Will fix shortly.

OPHD/UI/Core/Label.cpp
DanRStevens: Yes, I think the original code was quite useful, and there is no equivalent for it elsewhere.

OPHD/UI/Core/Label.cpp
DanRStevens: Ahh, I see. I think the intention I was going for, was to auto size in the constructor when a string is set, but otherwise not override if the size is changed. Leave it to the user to call the `autoSize` method if they want to adjust size after changing the text.

There may be cases where labels should have a fixed size. Perhaps in a grid layout for instance, where cells have a certain maximum width. In that case, the user can call `size` to set a consistent size on a set of `Label` instances.

Brett208: OK, that logic sounds not too difficult to setup.

DanRStevens: Rough logic anyway.

After writing my last comment, I was wondering about the behaviour when changing the text. Does it call `autoSize` then? Should it? I'm open to discussion on that.

I suppose the main point is you can set `size` after an `autoSize`, so the actual `Label` size might not snugly fit the text. Though unless we clip rendering, perhaps none of this actually matters.

Anyway, hopefully that at least explains why it was the way it was.

Brett208: Yeah, I understand what you are going for. Correct that clipping isn't actually supported. 

I was planning to solve by adding an enum to control resizing.

Really I just wanted to do some general refactor before creating TextControl not a full redesign of Label.

OPHD/UI/Core/UIContainer.h
```diff
@@ -11,7 +11,7 @@
  * 
  * Generally not intended to be used by itself.
  */
-class UIContainer: public Control
+class UIContainer: public TextControl
```
DanRStevens, line 14: This one caught me by surprise since a generic `UIContainer` doesn't inherently support text operations. Though the `Window` class does derives from `UIContainer`, and the `Window` class has text for the window title.

My first thought was to add the text support only to `Window`, but there doesn't seem to be a way to do that without introducing multiple inheritance. That's something I'd like to avoid for the time being.

I suppose this is a reasonable way to organize things for now. No better solution comes to mind.

ldicker83:
Are we still working on this or is it good for merging?

Brett208:
@DanRStevens,

Can you check if the capability you were looking for in Label has been restored? If so, recommend merging

-Brett

DanRStevens (commented):
Looks like it should be fine.

Clang seems to be complaining about an inconsistency where the `override` keyword was omitted for a `Structure` destructor (but was used on other virtual method overrides).

DanRStevens:
Looks like the `override` issue had been introduced on `master` recently. It would have been brought in by your recent merge. Fixed now.

Brett208:
Thanks for fixing the override issue. Yeah, the merge must have brought it in...

PR #687, opened 2020-09-02
Add food storage to command center

### LINKED ISSUE #630: Introduce food storage to Command Center
With the recent changes to resource storage, food is now handled directly by the agridomes. This becomes a problem with the CC not being counted as part of food storage which it should have 250 units of storage available.

ldicker83:
Closes #630

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -41,6 +41,19 @@ static int pullFood(int& food, int amount)
 }
 
 
+static inline void pullFoodFromStructure(FoodProduction* producer, int& remainder)
+{
+	if (remainder <= 0) { return; }
+
+	int foodLevel = producer->foodLevel();
+	int pulled = pullFood(foodLevel, remainder);
+
+	producer->foodLevel(foodLevel);
+	remainder -= pulled;
+}
+
+
+
 /**
  * 
  */
@@ -65,27 +78,12 @@ void MapViewState::updatePopulation()
 
 	for (auto structure : foodproducers)
 	{
-		if (remainder <= 0) { break; }
-
 		FoodProduction* foodProducer = static_cast<FoodProduction*>(structure);
-
-		int foodLevel = foodProducer->foodLevel();
-		int pulled = pullFood(foodLevel, remainder);
-
-		foodProducer->foodLevel(foodLevel);
-		remainder -= pulled;
+		pullFoodFromStructure(foodProducer, remainder);
 	}
 
-	/// \fixme There's some code repetition here that can be done better
-	if (remainder > 0)
-	{
-		auto cc = static_cast<FoodProduction*>(mTileMap->getTile(ccLocation()).structure());
-		int foodLevel = cc->foodLevel();
-		int pulled = pullFood(foodLevel, remainder);
-
-		cc->foodLevel(foodLevel);
-		remainder -= pulled;
-	}
+	auto cc = static_cast<FoodProduction*>(mTileMap->getTile(ccLocation()).structure());
+	pullFoodFromStructure(cc, remainder);
 }
 
 
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -50,6 +50,53 @@ static inline void pullFoodFromStructure(FoodProduction* producer, int& remainde
 }
 
 
+static RouteList findRoutes(micropather::MicroPather* solver, TileMap* tilemap, Structure* mine, const StructureList& smelters)
+{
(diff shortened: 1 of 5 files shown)


OPHD/Things/Structures/CommandCenter.h
```diff
@@ -1,14 +1,14 @@
 #pragma once
 
-#include "Structure.h"
+#include "FoodProduction.h"
 
 /**
  * Implements the Command Center structure.
  */
-class CommandCenter: public Structure
+class CommandCenter: public FoodProduction
```
DanRStevens, line 8: I see what you mean by counter-intuitive. I think this is partly why I have a sneaking suspicion that we might want a flatter structure hierarchy, and use a more data driven approach to control structure characteristics. Perhaps every structure can potentially have any feature, which is enabled/disabled by structure specific data. Think all structures having a power production value, but it being 0 for everything but power plants, and every structure having a food production capacity, but it being 0 for everything but agridomes. There are a lot of orthogonal concerns, which don't map well to inheritance (without introducing multiple inheritance), so a data driven approach might avoid some of the structure classification problems. Admitted though, I haven't thought deeply enough about the idea to know how well it would pan out.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -363,9 +347,7 @@ void MapViewState::checkColonyShip()
 }
 
 
-/**
- * 
- */
+
```
DanRStevens, line 342: Nitpick: Looks like an extra blank line was added here. (3 blank lines between methods)

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -50,6 +50,53 @@ static inline void pullFoodFromStructure(FoodProduction* producer, int& remainde
 }
 
 
+static RouteList findRoutes(micropather::MicroPather* solver, TileMap* tilemap, Structure* mine, const StructureList& smelters)
```
DanRStevens, line 53: I agree with re-organizing and grouping `static` freestanding methods.

I think slightly better is to wrap them in an unnamed namespace, which allows the `static` keyword to be removed. This particular use of the `static` keyword is perhaps less well known, and requires more careful reading to distinguish it from static class member methods.

Using unnamed namespaces or `static` on freestanding methods basically does the same thing, in that they prevent access from other translation units. This also protects against potential link errors if the method names are re-used (for different purposes) by separate translation units.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -106,15 +106,8 @@ void MapViewState::updatePopulation()
...
-	// FOOD CONSUMPTION
-	int food_consumed = mPopulation.update(mCurrentMorale, mFood, residences, universities, nurseries, hospitals);
 	StructureList &foodproducers = structureManager.structureList(Structure::StructureClass::FoodProduction);
-	int remainder = food_consumed;
-
-	if (mFood > 0)
-	{
-		remainder -= pullFood(mFood, remainder);
-	}
+	int remainder = mPopulation.update(mCurrentMorale, mFood, residences, universities, nurseries, hospitals);
```
DanRStevens, line 110: The removal of the `if` block doesn't coincide with the commit message. Maybe double check that the removal was intentional.

OPHD/States/MapViewStateTurn.cpp
ldicker83: I find unnamed namespaces to add more clutter than they're worth. Using `static` in this sense is very simple and anybody who's used C or C++ for more than a year or so should know immediately what it means. On the other hand, link errors due to name collisions is potentially worth it. I guess since the function naming I'm using is very specific I'm not worried about name collisions.

OPHD/Things/Structures/CommandCenter.h
ldicker83: Agreed on many points though I hesitate to use a data driven approach for the simple fact that it winds up producing massive classes with a lot of flaggy behavior. It would reduce the need for casting downstream though so that would be a significant benefit.

The 'if' block was intended. I accidentally squashed too many commits into one when trying to clean up the history. It's a relatively new thing for me and a detail I missed.

To note it here, the 'if' block was actually wrong as the mFood variable is a cached value used for display purposes.

OPHD/States/MapViewStateTurn.cpp
ldicker83: Fixed

OPHD/States/MapViewStateTurn.cpp
ldicker83: It was, as per a previous comment, I accidentally squashed too many commits into one. The if block is incorrect in this context as `MapViewState::mFood` is a cached value, not a working pool. The actual food storage is in the structures themselves.

PR #688, opened 2020-09-03
Fix an exception thrown due to use of std::map::at()

ldicker83:
Fixes an exception being thrown by the use of `std::map::at()` due to PRODUCT_NONE being set. Previous use was `std::map::operator[]` with the intended behavior of creating default product costs if an item wasn't found.

### DIFF
--- a/OPHD/Things/Structures/Factory.cpp
+++ b/OPHD/Things/Structures/Factory.cpp
@@ -20,12 +20,14 @@ using ProductionTypeTable = std::map<ProductType, ProductionCost>;
  */
 const ProductionTypeTable PRODUCTION_TYPE_TABLE =
 {
+	{ ProductType::PRODUCT_NONE, ProductionCost() },
+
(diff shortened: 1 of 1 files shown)


DanRStevens:
Ahh, curious, what is causing a production type lookup with no product type?

ldicker83:
It's actually by design, it happens when the fullscreen UI and the factory production window when a factory is idle. It's production value is set to `PRODUCT_NONE == -1` which was never added to the table. Using `map::at()` throws an exception if an entry is not in the `std::map`. The previous implementation used `operator[]` which would automatically add a `PRODUCT_NONE` entry into the product cost table. It relied on this behavior to avoid throwing since some entries were not explicitly added.

DanRStevens:
I'm wondering if that section of code should be updated to handle the none case specially. Maybe avoid the table lookup. Though it sounds like other products may be missing from the table?

ldicker83:
There are other products but I'd really rather not bother with special case code. The lookup is an integer based lookup so it's not exactly slow and it's not in performance critical code.

Issue #677, opened 2020-08-27
Remove low value comments - Constructor/Destructor


DanRStevens:
Comments should in general not simply restate what the code already says. Better to have no comment than a comment that just restates the code. When code is updated, comments are often missed, which can lead to stale comments that are either meaningless, or wrong, and possibly misleading. If comments don't add any value, they should be removed.

A common case of this are comments that mark constructors and destructors. These simply restate what the code is saying, and so should be removed.

Example:
```cpp
/*
 * C'tor.
 */
```

Example:
```cpp
/*
 * D'tor.
 */
```

There appear to be 29 affected files. (Containing `'tor`)

----

Longer example:
```cpp
/**
 * C'tor
 */
StateManager::StateManager() :
	mActiveState(nullptr),
	mActive(true)
{
	// Ensure that all quit messages are handled in some way even if a State object doesn't.
	Utility<EventHandler>::get().quit().connect(this, &StateManager::handleQuit);
}


/**
 * D'tor.
 *
 * \note	The active State is destroyed when the StateManager is destroyed.
 */
StateManager::~StateManager()
{
	if (mActiveState)
	{
		Utility<EventHandler>::get().disconnectAll();
	}
}
```

Side note: You may notice the destructor comment is outdated here, and mentions destroying the active `State`, even though there is no such code in the destructor. Effectively this is one of those stale comments. This is due to a change a while back where the active `State` was changed from using a raw pointer, to using `std::unique_ptr`. When using `std::unique_ptr`, the pointed to object is automatically released when the `std::unique_ptr` goes out of scope, which happens when the owning `StateManager` object is destroyed at the end of the destructor. There does not need to be explicit code for this, as it all happens automatically.

The relevant definition for the active state from the header file is:
```cpp
std::unique_ptr<State> mActiveState;
```

Out of date comments like this should be removed whenever possible.

----

If the destructor comment was still valid, or when uncertain if such a comment is invalid, that one line could be preserved, while the rest of the empty comment could be removed.

Original comment (with uncertain, possible still valid information):
```cpp
/**
 * D'tor.
 *
 * \note	The active State is destroyed when the StateManager is destroyed.
 */
```

Option 1 (multi-line Doxygen comment, low value lines removed, uncertain lines preserved):
```cpp
/**
 * \note	The active State is destroyed when the StateManager is destroyed.
 */
```

Option 2 (single like Doxygen comment, uncertain line preserved):
```cpp
[...]

Issue #676, opened 2020-08-27
Remove empty Doxygen placeholder comments


DanRStevens:
There are a lot of empty Doxygen placeholder comments. We should probably just remove them. If/when we have a note to add, the comment can be added in then. For now though, we should drop all empty comments.

Example:
```cpp
/**
 *
 */
```

These generally appear before function definitions. We can probably use a global find/replace to remove them, though that may leave extra blank lines in their place. Generally we want 2 blank lines separating function definitions. We should find a way to remove the comments, while keeping inter-function spacing normalized at 2 blank lines.

----

Longer example of empty comments:
```cpp
/**
 * 
 */
MineFacility::MineFacility(Mine* mine) : Structure(constants::MINE_FACILITY,
	"structures/mine_facility.sprite",
	StructureClass::Mine,
	StructureID::SID_MINE_FACILITY),
	mMine(mine)
{
	sprite().play(constants::STRUCTURE_STATE_CONSTRUCTION);
	maxAge(1200);
	turnsToBuild(2);

	requiresCHAP(false);
	selfSustained(true);
}


/**
 * 
 */
void MineFacility::activated()
{
	mMine->increaseDepth();
	mMine->active(true);
}


/**
 * 
 */
void MineFacility::think()
```

Brett208:
PR #700 pulls all the empty comment blocks (that I could find, I think it is pretty thorough though).

I left function vertical spacing alone. I don't know if we want to close this issue or leave open until they are standardized as there are quite a few with single vertical spacing.

DanRStevens:
I'm fine with closing if all the known empty comments have been removed. The vertical spacing comment was more about removing empty comments and leaving copious amounts of empty space where the removed lines were. I didn't intend to require all spacing be normalized, just that we don't introduce more spacing inconsistencies while cleaning this up, which I think was accomplished.

Brett208:
Ok, makes sense.

PR #700, opened 2020-09-13
Improve code comments

### LINKED ISSUE #644: Remove PVS-Studio comments
A number of files start with a comment:
```cpp
// This is an open source non-commercial project. Dear PVS-Studio, please check it.
// PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com
```

These comments appear to be obsolete now, and should be removed.

----

PVS-Studio was a local plugin tool that would scan the source code for errors. It required that comment to be present to allow for free scanning. As Visual Studio has built in scanning, which is apparently as good or better, the plugin is no longer being used.

Brett208:
Closes #644 
Partially address #676
Partially address #677

### DIFF
--- a/OPHD/UI/Core/Slider.cpp
+++ b/OPHD/UI/Core/Slider.cpp
@@ -182,9 +182,6 @@ void Slider::_buttonCheck(bool& buttonFlag, Rectangle<int>& rect, float value)
 }
 
 
-/**
- *
- */
 void Slider::onMouseDown(EventHandler::MouseButton button, int x, int y)
 {
 	if (!enabled() || !visible() || !hasFocus()) { return; }
@@ -203,9 +200,6 @@ void Slider::onMouseDown(EventHandler::MouseButton button, int x, int y)
 }
 
 
-/**
- *
- */
 void Slider::onMouseUp(EventHandler::MouseButton button, int x, int y)
 {
 	if (button != EventHandler::MouseButton::BUTTON_LEFT) { return; }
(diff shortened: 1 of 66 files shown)


Brett208:
I know the issues this PR addresses are marked for V1.0.0, but I thought it would be good to generally clean the code now. That way people won't use the older style as an example within the code base.

OPHD/UI/Core/Slider.cpp
```diff
@@ -1,5 +1,5 @@
 /**
- * \file	Slider.c++
+ * \file	Slider.cpp
  * \brief	Functions for sliding controls.
  * \author	Goof
```
DanRStevens, line 4: Bit of a side note, though I've read that when author's are listed in the source file of open source projects, it tends to discourage other people form maintaining the file. Perhaps we should omit author comments from source files. The information is still part of the commit meta data. Similar for comments that simply state the date a change was made.

DanRStevens (approved):
Thank you for taking care of that.

OPHD/UI/Core/Slider.cpp
DanRStevens: ... of course, I've never felt comfortable removing someone else's author comments. But then, that's maybe getting to the heart of the above.

ldicker83:
It's marked for 1.0.0 but that doesn't mean it needs to wait for that. For me it's a way of prioritizing tasks mostly for myself and just getting them assigned so that by 1.0.0 all of the tasks are completed. I have no objections to taking on tasks that are assigned for other milestones earlier especially if they're not related to core gameplay. :D

OPHD/UI/Core/Slider.cpp
Brett208: I had mixed thoughts about this as well. I think you are right about removing attribution unless it is a large module like the Pathfinder or something.

ldicker83: Author notes are an artifact of the past when source control wasn't a thing. Author information is in the repository commit history, I see very low value in keeping authorship comments in place. Including the file name also makes no real sense.

PR #701, opened 2020-09-13
Fix energy value display

### LINKED ISSUE #691: Energy value is always set to 0
Energy value is always set to 0 regardless of how much energy is being produced. This appears to be a regression defect.

ldicker83:
Closes #691

### DIFF
--- a/OPHD/States/MapViewStateDraw.cpp
+++ b/OPHD/States/MapViewStateDraw.cpp
@@ -176,7 +176,7 @@ void MapViewState::drawResourceInfo()
 	{
 		std::tuple{NAS2D::Rectangle{96, 32, iconSize, iconSize}, refinedResourcesInStorage(), totalStorage(Structure::StructureClass::Storage, 1000), totalStorage(Structure::StructureClass::Storage, 1000) - refinedResourcesInStorage() <= 100},
 		std::tuple{NAS2D::Rectangle{64, 32, iconSize, iconSize}, mFood, totalStorage(Structure::StructureClass::FoodProduction, 1000), mFood <= 10},
-		std::tuple{NAS2D::Rectangle{80, 32, iconSize, iconSize}, mEnergy, NAS2D::Utility<StructureManager>::get().totalEnergyProduction(), mEnergy <= 5}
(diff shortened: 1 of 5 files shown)


OPHD/States/MapViewStateDraw.cpp
```diff
@@ -172,11 +172,12 @@ void MapViewState::drawResourceInfo()
 	}
 
 	// Capacity (Storage, Food, Energy)
+	const auto& sm = NAS2D::Utility<StructureManager>::get();
```
DanRStevens, line 172: Not sure how other people feel, though I generally prefer variable names to be spelled out in full. For the definition, the abbreviation may be reasonable clear, but at the point of use that context is lost.

DanRStevens (commented):
Ahh, nice to see this update.

Curious how the submodule was updated in this branch. I believe I made some small breaking changes recently to NAS2D, though the corresponding OPHD update was already merged to master. I assume this relates to the merge of master, and possibly the locally checked out submodule being the earlier out of date version. I suspect a `git submodule update` (checkout the expected submodule version) would have fixed the issue, with needing to commit a new update. Git does have a slightly confusing behaviour, in that when you pull or checkout a new version of OPHD, it won't automatically update the NAS2D submodule to a matching version. That's generally a separate command.

ldicker83:
Yeah, for whatever reason TortoiseGit doesn't do a recursive checkout/merge/update/whatever on its own and I forgot that there were breaking changes to nas2d-core that also needed to be merged in. It's what I get for not compiling first which normally I do before committing merge changes like that.

OPHD/States/MapViewStateDraw.cpp
ldicker83: This one is iffy for me -- I don't generally object but the scope here is so limited it almost seems like overkill to spell it out. The intention of the variable in the first place was to 1) avoid multiple lookups and 2) help shorten the verbosity of the code. I'm not sure if fully spelling it out here would help more than it would hinder considering the limited scope?

DanRStevens: True, the scope is quite limited here, though refactoring does tend to move variables around, so scope can sometimes expand. It's nice to be able to expand scope without having to rename things at the same time. Besides, if you use auto complete, it doesn't make much of a typing difference. And if you don't use autocomplete, the speed of programming is never limited by the speed of typing. More likely, the speed of understanding already written code will be more relevant. In that case, seeing the use "sm.someMethod" will be less immediately clear than "structureManager.someMethod". The definition is mercifully close, but it still requires cross referencing the use with the defining line.

PR #707, opened 2020-09-20
Add road structure

### LINKED ISSUE #117: Add Road Structure
Title basically says it all.

I want to include the Road in the path solver's logic so that'll be the second part of this implementation though in a separate issue.

ldicker83:
Closes #117


OPHD/States/MapViewStateTurn.cpp
```diff
@@ -276,9 +277,9 @@ void MapViewState::updateResources()
 			std::clamp(stored.resources[3], 0, 25)
 		};
 
-		stored -= moved;
-		addRefinedResources(moved);
-		stored += moved;
+stored -= moved;
+addRefinedResources(moved);
+stored += moved;
 	}
```
DanRStevens, line 283: What happened to the indentation here?

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -278,11 +278,9 @@ bool selfSustained(StructureID id)
 	case StructureID::SID_COMM_TOWER:
 	case StructureID::SID_ROAD:
 		return true;
-		break;
 
 	default:
 		return false;
-		break;
 	}
```
DanRStevens: Another way this could be written, and easily extended, is to use a `std::array` to hold the IDs of self sustained structures, then use `std::find` to check if the ID exists in the self sustained array. The array could be `constexpr` defined in the cpp file, and the method could become a one-liner:
```cpp
return std::find(selfSustainedIds.begin(), selfSustainedIds.end(), id) != selfSustainedIds.end();
```

You could even mark the function as `constexpr`, allowing it to be evaluated at compile time (when passed a literal constant, rather than a variable).

DanRStevens (commented):
Looks good. Had a couple of small comments to add.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -350,6 +351,34 @@ void MapViewState::updateFood()
...
+
+		std::array<bool, 4> surroundingTiles{ false, false, false, false };
+		for (size_t i = 0; i < 4; ++i)
+		{
+			const auto tileToInspect = tileLocation + DirectionClockwise4[i];
+			if (!mTileMap->isValidPosition(tileToInspect)) { continue; }
+			if (!mTileMap->getTile(tileToInspect).thingIsStructure()) { continue; }
+
+			surroundingTiles[i] = mTileMap->getTile(tileToInspect).structure()->structureId() == StructureID::SID_ROAD;
+		}
```
DanRStevens, line 375: One more comment I forgot to add:

In other projects, I've seen this sort of thing tackled where the bit flags are binary encoded into an index. Effectively there are 16 possible graphics for 4 possibly adjacent tiles, including T-intersections.

```cpp
int index = 0;
int field = 1;
for (const auto direction : DirectionClockswise4)
{
  if (check(tileLocation + direction))
  {
    index |= field;
    field <<= 1;
  }
}
```

If you exclude T-intersections, then there are 4 possible graphics.

```cpp
const int index =
  (check(tilePosition + DirectionNorth) | check(tilePosition + DirectionSouth)) |
  (check(tilePosition + DirectionEast) | check(tilePosition + DirectionWest)) * 2;
```

An index of 0 means no adjacent tiles. Assuming no unconnected point graphics, you can probably default index 0 to some default direction. A max index (either 15 or 3), means all adjacent tiles (cross intersection).

The index can then potentially be used directly to select the tile index to draw from an array of images.

ldicker83: That would probably be a faster method that uses less memory but it's also a little less understandable when quickly reading through it. I opted for pattern matching in this case because it's pretty simple and there are a very finite number of options.

Something to consider for the future though.

OPHD/States/MapViewStateHelper.cpp
ldicker83: True.

If this case statement starts to get larger (before it was a simple if structureID == Comm Tower), that would be the solution I opt for.

OPHD/States/MapViewStateTurn.cpp
ldicker83: Definitely accidental change. Not sure how/why I missed this when committing. Gonna push a change direct to master over this (since I messed up and accidentally committed a bunch of changes direct to master instead of a new branch, whoops).

PR #713, opened 2020-09-26
Implement truck assignment

### LINKED ISSUE #690: Implement truck assignment to mine facility
Trucks are treated as a sort of commodity unlike robots. When trucks are assigned to a MineFacility, they should be pulled out of Warehouse storage and added to an internal count of trucks assigned to a mine facility.

ldicker83:
Implements and closes #690

### DIFF
--- a/OPHD/Things/Structures/MineFacility.h
+++ b/OPHD/Things/Structures/MineFacility.h
@@ -23,6 +23,10 @@ public:
 
 	int digTimeRemaining() const;
 
+	int assignedTrucks()  const { return mAssignedTrucks; }
+	void addTruck() { mAssignedTrucks = std::clamp(++mAssignedTrucks, 1, 10); }
+	void removeTruck() { mAssignedTrucks = std::clamp(--mAssignedTrucks, 1, 10); }
+
 	/**
 	 * Gets a pointer to the mine the MineFacility manages.
 	 */
@@ -44,6 +48,7 @@ private:
 private:
 	int mMaxDepth = 0; /**< Maximum digging depth. */
 	int mDigTurnsRemaining = 0; /**< Turns remaining before extension is complete. */
+	int mAssignedTrucks = 1; /**< All mine facilities are built with at least one truck. */
 
 	Mine* mMine = nullptr; /**< Mine that this facility manages. */
 
--- a/OPHD/Things/Structures/MineFacility.h
+++ b/OPHD/Things/Structures/MineFacility.h
@@ -24,8 +24,8 @@ public:
 	int digTimeRemaining() const;
 
 	int assignedTrucks()  const { return mAssignedTrucks; }
-	void addTruck() { mAssignedTrucks = std::clamp(++mAssignedTrucks, 1, 10); }
-	void removeTruck() { mAssignedTrucks = std::clamp(--mAssignedTrucks, 1, 10); }
+	void addTruck() { mAssignedTrucks = std::clamp(mAssignedTrucks + 1, 1, 10); }
+	void removeTruck() { mAssignedTrucks = std::clamp(mAssignedTrucks - 1, 1, 10); }
 
 	/**
 	 * Gets a pointer to the mine the MineFacility manages.
--- a/OPHD/UI/MineOperationsWindow.cpp
+++ b/OPHD/UI/MineOperationsWindow.cpp
@@ -36,7 +36,9 @@ MineOperationsWindow::MineOperationsWindow() :
 	chkRareMinerals{"Rare Minerals"},
 	btnIdle{"Idle"},
 	btnExtendShaft{"Dig New Level"},
-	btnOkay{"Close"}
+	btnOkay{"Close"},
+	btnAssignTruck{"Add Truck"},
+	btnUnassignTruck{"Remove Truck"}
 {
 	text(constants::WINDOW_MINE_OPERATIONS);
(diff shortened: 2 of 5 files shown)


ldicker83:
So... yeah, I see the errors that are being issued but I don't understand them. How would the operation work on an uninitialized pointer in this case? Is it because I defined these in the header / inline? If that's the case, I hate GCC more than I ever have but what the hell ever, I can move them to the source file and not have them inlined.

DanRStevens:
Ahh yes, it's complaining about the line:
```cpp
void addTruck() { mAssignedTrucks = std::clamp(++mAssignedTrucks, 1, 10); }
```

The error comes from modifying the `mAssignedTrucks` variable twice, without a sequence point between them. Once is from the `++mAssignedTrucks`, and the other is from the assignment `mAssignedTrucks = ...`. Since these two operations are unsequenced, you end up with undefined results, hence the error message `... may be undefined [-Werror=sequence-point]`.

OPHD/Things/Structures/MineFacility.h
```diff
@@ -23,6 +23,10 @@ class MineFacility: public Structure
 
 	int digTimeRemaining() const;
 
+	int assignedTrucks()  const { return mAssignedTrucks; }
+	void addTruck() { mAssignedTrucks = std::clamp(++mAssignedTrucks, 1, 10); }
+	void removeTruck() { mAssignedTrucks = std::clamp(--mAssignedTrucks, 1, 10); }
```
DanRStevens, line 28: Based on context here, it seems you only want to modify the variable when it falls within a specified bounds. As such, I would avoid using `++mAssignedTrucks` and `--mAssignedTrucks`, which would unconditionally modify the variables. Instead, pass `mAssignedTrucks + 1`, and `mAssignedTrucks - 1`.

OPHD/UI/MineOperationsWindow.cpp
```diff
@@ -137,13 +141,44 @@ void MineOperationsWindow::btnIdleClicked()
...
 void MineOperationsWindow::btnAssignTruckClicked()
 {
+	updateTruckAvailability();
+
+	if (mAvailableTrucks == 0) { return; }
+	
+	auto& warehouseList = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::Warehouse);
+	for (auto warehouse : warehouseList)
+	{
+		Warehouse* w = static_cast<Warehouse*>(warehouse);
```
DanRStevens, line 151: When using the Atom editor, short variables with only 1 letter will not highlight when selected. This makes it harder to spot other references to the variable.

You also need to be careful to do whole word matching when renaming such variables, since "w" is likely to exist as a substring in other names.

ldicker83: This is fair, though I'm not sure what name to use besides something like `wh` or `warehousePtr` or something else equally obnoxious. I'm open to suggestion.

DanRStevens: I would normally go with `warehouse`. I see you already have a `warehouse` variable from before the cast though. I might be tempted to rename the original to `structure`, and the casted one to `warehouse`.
```cpp
auto& warehouses = structureManager.strutcureList(Structure::StructureClass::Warehouse);
for (auto structure : warehouses)
{
  auto* warehouse = static_cast<Warehouse*>(structure);
  warehouse->warehouseSpecificMethod();
}
```

----

Or perhaps the list itself could be casted. That would make iteration over the expected type much more natural.
```cpp
auto& warehouses = static_cast<std::vector<Warehouse*>>(structureManager.strutcureList(Structure::StructureClass::Warehouse));
for (auto warehouse : warehouses)
{
  warehouse->warehouseSpecificMethod();
}
```

----

This gets me thinking, maybe we could have a templated `StructureManager` method that allows getting structures of a certain type, and would use that exact type for the return value. Then you could write something like:
```cpp
const auto warehouses = structureManager.structureList<Warehouse>();
for (auto warehouse : warehouses)
{
  warehouse->warehouseSpecificMethod();
}
```

ldicker83: I like this idea the best. Not sure how I'd go about implementing it... my knowledge of template programming is woefully lacking. Definitely worth of its own task.

OPHD/Things/Structures/MineFacility.h
ldicker83: Yup, that's precisely what I was going for. This is why so-called 'clever' programming is almost always a bad idea. While the code behaved as expected on the MSVC++ compiler, it can be ambiguous and could have unintended side effects.

In this case I'd rather use the much more clear `variable + 1` option.

OPHD/UI/MineOperationsWindow.cpp
DanRStevens: Indeed, that last option is the least obvious to implement.

If you want immediate results without having to think too much, I'd go with one of the first two options for now. Can always update things later.

----

If you want the benefits of the third method, but without having to deal with templates, you can create specialized methods, such as `getWarehouses()` and such. Internally a `getWarehouses` method could make use of `structureList` internally, and wrap the internal call with a `static_cast`, as per the second option.

Alternatively, if you want to go direct to the template approach, you could make use of `dynamic_cast`. If a cast to the template type succeeds, add the pointer to the output list.

PR #729, opened 2020-12-10
Refactor ListBox interface

DanRStevens:
Pre-work making `ListBox` interface changes to help address #479. This tries to make the interfaces a little more similar between `ListBox` and `ListBoxBase`, and perhaps hopefully easier to use.

The big difference between `ListBox` and `ListBoxBase` is the different fields supported by the `ListBoxItem` struct. Methods in the `ListBox` class that directly touch those fields are less portable between the two classes. To work around this, direct read-only access to `ListBoxItem` structs is provided, pushing those differences out of the `ListBox` classes. In particular, the caller will know which type of `ListBox` it is using, and so it will know which fields should be present in a `ListBoxItem` struct. Eventually that `ListBoxItem` struct could become a template parameter.

One area I haven't yet touched is selecting an item by name. This of course touches the `ListBoxItem` fields directly, and so will not be portable between class types. An update to make that method generic will likely involve lambdas. A change I figure I'll make in a separate pull request.

### DIFF
--- a/OPHD/UI/Core/ListBox.cpp
+++ b/OPHD/UI/Core/ListBox.cpp
@@ -160,6 +160,12 @@ void ListBox::setSelectionByName(const std::string& item)
 }
 
 
+bool ListBox::isItemSelected() const
+{
+	return mCurrentSelection != constants::NO_SELECTION;
+}
+
+
 /**
  * Drops all items from the list.
  */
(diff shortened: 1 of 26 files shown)


OPHD/UI/Core/ListBox.cpp
```diff
@@ -169,7 +169,7 @@ const ListBox::ListBoxItem& ListBox::selected() const
 void ListBox::clear()
 {
 	mItems.clear();
-	mSelectedIndex = 0;
+	mSelectedIndex = constants::NO_SELECTION;
```
ldicker83, line 172: Definitely a potential bug here. I think other parts of the code sort of protect against this but this is definitely a defect.

PR #730, opened 2020-12-10
Add ListBox::selectIf template method

DanRStevens:
Add `ListBox::selectIf` template method, taking a unary predicate used to check items in the `ListBox`.

Use new method to improve efficiency, and to remove old public API method `ListBox::setSelectedByName`. The old method made assumptions about the contents of `ListBoxItem`, and so was not portable between the `ListBox` and `ListBoxBase`.

Prep-work for #479.

### DIFF
--- a/OPHD/UI/Core/ComboBox.cpp
+++ b/OPHD/UI/Core/ComboBox.cpp
@@ -4,6 +4,7 @@
 
 #include <NAS2D/Utility.h>
 #include <NAS2D/MathUtils.h>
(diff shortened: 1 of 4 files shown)


DanRStevens:
@ldicker83: It might make sense to unselect any selected items in case no match was found. What do you think? The original didn't do that, and it may be desirable to only update the selection if something was found, so I left it as is for now.

If we want to make the change, we can add this as the last line after the `for` loop:
```cpp
mSelectedIndex = constants::NO_SELECTION;
```

PR #735, opened 2020-12-14
Convert ListBox to template

DanRStevens:
Support work for: #479

Convert `ListBox` to template class. This allows alternate `ListBoxItem` structs to be used. Memory is fully owned by the vector, with no extra indirection or need for `new`/`delete`. Additional item fields can be added, with more complex rendering, and no loss of type information that would otherwise require casting to the derived type.


OPHD/UI/Core/ListBox.h
```diff
@@ -60,7 +60,10 @@ class ListBox : public Control
 	~ListBox() override;
 
 	bool isEmpty() const;
-	std::size_t count() const { return mItems.size(); }
+
+	std::size_t count() const {
```
ldicker83, line 88: So... not sure why we're expanding these. Also, the brace placement doesn't match the rest of the project.

DanRStevens: I'm kind of forgetting the context on this one now, though I believe it was an attempt to improve consistency. Templates need to be fully available at the point they are used, so template function implementations must be in the header. Some of those implementations are a bit long, so a lot of multi-line functions. I figured it'd be more consistent if every function was expanded to the multi-line format.

Good point on the brace placement. I suppose that style has become my new habit. I don't really think about it much anymore. Particularly for header functions and lambdas. I find the line trailing opening brace is more compact and spills nicer for deeply nested constructs when you need to split long lines. Though the argument of being more compact is a bit silly when I'm expanding single line methods to a multi-line format. Mostly it was just habit.

ldicker83: Makes sense on expanding them, makes it easier moving forward to expand the functions as needed for templatizing it.

I don't mean to be a brace placement nazi. As much as I don't care for that style I'm more concerned with the rest of the project using the 'older' style of starting brace placement on a new line. I won't split hairs about it though. :D

PR #740, opened 2020-12-30
Add MacOS build on CircleCI

DanRStevens:
This attempt caught defects fixed in #738 and #739:
https://app.circleci.com/pipelines/github/OutpostUniverse/OPHD/1160/workflows/c46deafd-1c3f-4194-a0a0-6e578acb7948/jobs/4968

Still need to figure out how to support the OpenGL dependency on MacOS.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,22 @@ commands:
       - run: make package
       - store_artifacts:
           path: .build/package/
+  brew-install:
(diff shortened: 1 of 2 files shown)


DanRStevens:
Ahh yes, that's why the problem seemed familiar. The NAS2D `makefile` already had a solution for linking to OpenGL on MacOS. Was able to just copy that to OPHD.

PR #749, opened 2021-01-31
Add biowaste processing

### LINKED ISSUE #96: Add bio-waste update during turn processing
After #95 is completed, the turn processing check will need to check each residential structure for bio-waste accumulation.

ldicker83:
Closes #96

### DIFF
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -191,6 +191,7 @@ void MapViewState::load(const std::string& filePath)
 
 	Utility<StructureManager>::get().updateEnergyProduction();
 	Utility<StructureManager>::get().updateEnergyConsumed();
+	Utility<StructureManager>::get().assignColonistsToResidences(mPopulationPool);
 
 	updateRobotControl(mRobotPool);
 	updateResidentialCapacity();
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -410,6 +410,21 @@ void MapViewState::readStructures(Xml::XmlElement* element)
 		loadResorucesFromXmlElement(structureNode->firstChildElement("production"), structure.production());
 		loadResorucesFromXmlElement(structureNode->firstChildElement("storage"), structure.storage());
 
+		if (structure.structureClass() == Structure::StructureClass::Residence)
+		{
+			auto waste = structureNode->firstChildElement("waste");
+			if (waste)
+			{
+				Residence* residence = static_cast<Residence*>(&structure);
+				auto accumulated = waste->attribute("accumulated");
+				residence->wasteAccumulated(std::stoi(accumulated));
+
+				auto overflow = waste->attribute("overflow");
+				residence->wasteOverflow(std::stoi(overflow));
+			}
+
+		}
+
 		if (structure.isWarehouse())
 		{
 			auto& warehouse = *static_cast<Warehouse*>(&structure);
--- a/OPHD/StructureManager.cpp
+++ b/OPHD/StructureManager.cpp
@@ -452,6 +452,15 @@ void StructureManager::serialize(NAS2D::Xml::XmlElement* element)
 			structureElement->linkEndChild(food);
 		}
 
+		if (structure->structureClass() == Structure::StructureClass::Residence)
+		{
+			Residence* residence = static_cast<Residence*>(structure);
+			auto* waste = new NAS2D::Xml::XmlElement("waste");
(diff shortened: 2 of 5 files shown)


OPHD/StructureManager.cpp
```diff
@@ -452,6 +452,15 @@ void StructureManager::serialize(NAS2D::Xml::XmlElement* element)
...
 		}
 
+		if (structure->structureClass() == Structure::StructureClass::Residence)
+		{
+			Residence* residence = static_cast<Residence*>(structure);
+			auto* waste = new NAS2D::Xml::XmlElement("waste");
+			waste->attribute("accumulated", residence->wasteAccumulated());
+			waste->attribute("overflow", residence->wasteOverflow());
+			structureElement->linkEndChild(waste);
+		}
```
DanRStevens, line 462: Looks like there are a lot of special case situations being handled in `StructureManager`. Maybe a `virtual` method should be added to `Structure` that returns a `Dictionary` of key/value pairs that need to be saved when writing a saved game file. The `StructureManager` can then translate the `Dictionary` to appropriate fields on an `XmlElement`.

At least that's assuming all such data that might be saved can be represented as key/value pairs.

You could also pass an `XmlElement` into a `virtual` method on `Structure` to have it write extra values. That would be more general, though it couples more of your code to the specifics of the XML library you're using. That may be worth avoiding since there is a plan to upgrade or replace that library.

OPHD/States/MapViewStateIO.cpp
```diff
@@ -410,6 +410,21 @@ void MapViewState::readStructures(Xml::XmlElement* element)
...
+			{
+				Residence* residence = static_cast<Residence*>(&structure);
+				auto accumulated = waste->attribute("accumulated");
+				residence->wasteAccumulated(std::stoi(accumulated));
+
+				auto overflow = waste->attribute("overflow");
+				residence->wasteOverflow(std::stoi(overflow));
+			}
+
+		}
```
DanRStevens, line 426: This looks to be the reverse of the previous. I suppose this could also be handled by a `virtual` method, though it might be more appropriate to use a factory method to construct `Structure` objects while loading a file. This might be worth thinking about a bit more.

It should be possible to handle this without needing to cast.

----

Curiously the writing and reading code for extra `Structure` fields is split into different classes, `StructureManager` and `MapViewState`. That should give pause. Ideally `Structure` should be responsible for its own fields.

OPHD/StructureManager.cpp
ldicker83: I like this idea though I don't have a complete idea on how to implement. If you get a chance opening a ticket detailing this change would be very helpful!

OPHD/States/MapViewStateIO.cpp
ldicker83: Yeah, I really hate how I managed to come up with that. I started with the idea of a serialize/deserialize interface (which you can sort of see with member function names) but then it sort of got away from me and devolved into the nonsense that it is now.

A serialize/deserialize interface would still make sense for this though before I go about cleaning it up I'd want a good solid idea on design and implementation. What we see here is a case of three different paradigms, none of which are well thought out or implemented and then just a bunch of unrelated code in all different parts of the program. Does the structure read/write itself? Does the StructureManager do it? Does MapViewState do it? Should none of them do it and instead have a special Serializer type of class handle it with each object that needs to be read/written inheriting an interface that is then called by the Serializer class?

Any of these options could work. I'm not sure which is "the best" but anything that's _consistent_. Fixing this would make adding new structures a lot easier because then there aren't at least two different places to look for read/write code.

ldicker83: > Curiously the writing and reading code for extra `Structure` fields is split into different classes, `StructureManager` and `MapViewState`. That should give pause. Ideally `Structure` should be responsible for its own fields.

That was my original intent but I wanted to avoid XML code in the `Structure` classes themselves. As I think about it, though, I realize that most structures don't need special case code so they can inherit from an interface class (say, `Serializable`), implement the basic stuff and then let the `Structure`'s that need to add additional information override the default implementation from `Structure`.

Just a thought at this point.

That same interface could be used by `MapViewState` or `GameState` (probably the latter) in the same fashion since it needs to read/write values as well.

`TileMap` probably too though that's debatable if `TileMap` should do it or if `MapViewState`/`GameState` should do it.

ldicker83:
Last post about this I swear.

I think I referenced this at some point but I came across a C++ header-only serialization library: http://uscilab.github.io/cereal/index.html

Might be worth using this to implement saving/loading since it can do binary and xml. Use XML during development and switch to binary for release modes?

PR #752, opened 2021-02-01
Fix an out of bounds issue in the StructureInspector window that...

ldicker83:
... attempts to access a row in StringTable that isn't allocated.

Not sure how this one was missed? Realized it when it came to structures that employ Scientists.

### DIFF
--- a/OPHD/UI/StructureInspector.cpp
+++ b/OPHD/UI/StructureInspector.cpp
@@ -52,7 +52,7 @@ void StructureInspector::update()
 	}
 	text(mStructure->name());
 
-	StringTable stringTable(4, 3);
+	StringTable stringTable(4, 4);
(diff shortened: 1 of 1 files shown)


Brett208:
Sorry, that was probably my fault. Thanks for fixing.

ldicker83:
No worries, I figured it was an oversight. I do that sort of thing all the time -_-

DanRStevens:
I wonder if there is a way to auto size the `StringTable`, based on array data. Might be complicated by the 2D nature of the data though.

Brett208:
Yeah, since I implemented it in a 1D vector under the hood it might be a bit complex to auto size. I think it would be a nice feature to add though if someone had the time on their hands.

PR #753, opened 2021-02-03
Add biowaste recycling logic

### LINKED ISSUE #750: Implement Recycling Facility logic
After #97 is complete, the logic involved in dealing with bio-waste will need to be added.

ldicker83:
Closes #750

Adds biowaste processing logic when player builds recycling structures and adds a morale hit when waste is overflowing at residences.

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -336,7 +336,27 @@ void MapViewState::updateResidentialCapacity()
 
 void MapViewState::updateBiowasteRecycling()
 {
+	auto& residences = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::Residence);
+	auto& recyclingFacilities = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::Recycling);
 
+	if (residences.empty() || recyclingFacilities.empty()) { return; }
+
+	auto residenceIterator = residences.begin();
+	for (auto recyclingFacility : recyclingFacilities)
+	{
+		int count = 0;
+
+		// Consider a different control path
+		if (!recyclingFacility->operational()) { continue; }
+
+		for (; residenceIterator != residences.end(); ++residenceIterator)
+		{
+			Residence* residence = static_cast<Residence*>(*residenceIterator);
+			residence->pullWaste(30);
+			++count;
+			if (count >= 10) { break; }
+		}
+	}
 }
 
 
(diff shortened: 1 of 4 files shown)


OPHD/Things/Structures/Recycling.h
```diff
@@ -18,6 +22,26 @@ class Recycling : public Structure
...
+	
+	
+	/**
+	 * Number of residential units the facility can support
+	 * each turn.
+	 */
+	virtual int residentialSupportCount() const
+	{
+	return ResidentialSupportCount;
+	}
```
DanRStevens, line 42: Indentation

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -334,6 +341,42 @@ void MapViewState::updateResidentialCapacity()
...
+			if (count >= recycling->residentialSupportCount())
+			{
+				break;
+			}
+		}
+
+		// We got to the end of the residence list, don't waste time iterating over
+		// any remaining recycling facilities.
+		if (residenceIterator == residences.end()) { break; }
+	}
```
DanRStevens, line 376: Maybe the loop should be on the `count` variable, with an early exit via `return` if the end of the residences list is reached. That would avoid the need for an `if` test after the `for` loop.

```cpp
for (auto recyclingFacility : recyclingFacilities) {
  for (int count = 0; count < recycling->residentialSupportCount(); ++count) {
    if (residenceIterator == residences.end()) {
      return; // No more residences, so don't waste time iterating over remaining recycling facilities
    }
    Residence* residence = static_cast<Residence*>(*residenceIterator);
    residence->pullWaste(recycling->wasteProcessingCapacity());
  }
}
```

DanRStevens (commented):
Ahh, nice to see progress on the game logic. :slightly_smiling_face:

OPHD/States/MapViewStateTurn.cpp
DanRStevens: I've noticed those `operational()` checks in a few places. Maybe it would be helpful to have a method that returns a list of operational structures of a given type.

Perhaps that can tie in to the earlier idea from maybe a few months back about having a template method to return a list of structures of a given type, so the casting isn't needed.

ldicker83: Blah, managed to miss this comment before merging.

I like the suggestion. Reminds me to get a second set of eyes on code.

Candidate for refactoring later on :D

PR #754, opened 2021-02-03
Mark StructureManager::structureList return value as const

DanRStevens:
To prevent accidental modification of the original structure lists, the return value should have `const` added, as below:
```cpp
const StructureList& structureList(Structure::StructureClass structureClass);
```


ldicker83:
You know, I was looking at this and thinking "These lists should probably be const since we don't want to modify the list even if we're modifying the objects pointed at by the lists".

Thanks for fixing!

PR #757, opened 2021-02-06
Adjust mine operations window text

### LINKED ISSUE #743: Turns remaining to dig new mine level causes text to display incorrectly
**Version:** 0.8.0
**OS:** Windows 10 64 Bit
**Description:** After clicking the button _Dig New Level_, the text field to show turns remaining overlays on top of the Trucks Assigned field.

Dig new level text overlap

ldicker83:
Closes #743

### DIFF
--- a/OPHD/UI/MineOperationsWindow.cpp
+++ b/OPHD/UI/MineOperationsWindow.cpp
@@ -1,5 +1,6 @@
 #include "MineOperationsWindow.h"
 
+#include "StringTable.h"
 #include "TextRender.h"
 
 #include "../Cache.h"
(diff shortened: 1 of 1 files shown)


ldicker83:
As a note for commit 2482129, a lot of the padding spacing is formatted automatically by the IDE according to set formatting rules. The code that I commit conforms to this type of spacing. In some PR's and I'll have commits like this one that will automatically format code and mark that as a 'grooming' post.

I'm less concerned about 'proper spacing' as I am about brace placement. And brace placement only to stick with conventions used in the rest of the code.

DanRStevens:
Hmm, for most of the code base, I haven't been following the convention of adding spaces inside of `{}` when initializing data. I generally add spaces inside `{}` for code blocks, such as lambdas. Similarly I noticed it used for single line `if` statements. For data though, I tend to keep the braces right with no spaces.

I remember we had a style file at one point for one of the projects. Though we don't seem to be using that here. I'm guessing this is default formatting suggested by MSVC.

I like that it warned you about the missing `&` potentially affecting performance.

ldicker83:
It didn't mention the performance overhead, I deduced that from it's warning of "Hey, so auto doesn't deduce references and you're making a copy here when you may actually want a reference."

The Visual Studio IDE is really getting very good at suggestions and finding potential mistakes as you're working. The tools are really getting good.

PR #762, opened 2021-02-08
Add guard to avoid potentially dereferencing a nullptr in savegame load code

### LINKED ISSUE #761: MapViewState::load() throws exception if a RobotCommandCenter has an empty robot ID string
The exception is thrown from `NAS2D::split()` if a Robot Command Center is saved without any robots attached to it.

I haven't decided if `NAS2D::split()` should threat this as incorrect usage and throw a runtime error exception (something like "Empty string passed") or if it should silently ignore this and just return an empty string. I've opened a separate ticket for that in the NAS2D project to consider.

As for OPHD's usage case, this should be addressed and prevented from being an issue since an empty robot ID list string doesn't need to be processed.

ldicker83:
As noted by @DanRStevens in the comments, there is the potential for a nullptr dereference. This PR corrects that.

### DIFF
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -450,15 +450,18 @@ void MapViewState::readStructures(Xml::XmlElement* element)
 			auto& rcc = *static_cast<RobotCommand*>(&structure);
 			XmlAttribute* robotsAttribute = structureNode->firstChildElement("robots")->firstAttribute();
 
-			for (const auto& string : NAS2D::split(robotsAttribute->value(), ','))
+			if (!robotsAttribute->value().empty())
 			{
-				const auto robotId = NAS2D::stringTo<int>(string);
-				for (auto* robot : mRobotPool.robots())
+				for (const auto& id_string : NAS2D::split(robotsAttribute->value(), ','))
 				{
-					if (robot->id() == robotId)
+					const auto robotId = NAS2D::stringTo<int>(id_string);
+					for (auto* robot : mRobotPool.robots())
 					{
-						rcc.addRobot(robot);
-						break;
+						if (robot->id() == robotId)
+						{
+							rcc.addRobot(robot);
+							break;
+						}
 					}
 				}
 			}


OPHD/States/MapViewStateIO.cpp
```diff
@@ -450,15 +450,18 @@ void MapViewState::readStructures(Xml::XmlElement* element)
 			auto& rcc = *static_cast<RobotCommand*>(&structure);
 			XmlAttribute* robotsAttribute = structureNode->firstChildElement("robots")->firstAttribute();
 
-			for (const auto& string : NAS2D::split(robotsAttribute->value(), ','))
+			if (!robotsAttribute->value().empty())
```
ldicker83, line 453: Not digging the deeper nesting here. Would rather have the code block that handles robot id lists just be skipped but I'm failing to come up with a better way to do this that doesn't use `goto` to skip it. `break` and `continue` aren't good options either.

Could just have this spit out a log message if empty? I dunno, @DanRStevens if you can think of a better way to do this let me know before merging this PR. Thanks!

ldicker83:
As a note: This PR may be unnecessary if we opt to specifically handle empty strings passed to `NAS2D::split()`. See https://github.com/lairworks/nas2d-core/issues/825 for details.

OPHD/States/MapViewStateIO.cpp
DanRStevens: Might want to check the `firstChildElement` call is actually able to find an appropriate node before attempting to call `firstAttribute`.

From the docs, it appears `firstChildElement` can return `nullptr` if no such node is found:
http://www.grinninglizard.com/tinyxmldocs/classTiXmlHandle.html#f0aea751320f5e430fac6f8fff3b8dd4

ldicker83: You're generally right, but since the save file is generated by the game, it should always be there.

Maybe for the sake of less fragile code I should put it in there, but then there's the question of silently ignore or throw?

DanRStevens: Well, it's an external text file. Someone can always edit it with a text editor and mess with the data. Generally best to never trust external data files that people can mess with.

ldicker83: That is fair. I'll make that change here and undo the nested if since the latest change to NAS2D fixes the issue. Thanks!

ldicker83:
Going to use a different change set to handle the above mentioned issue.

PR #763, opened 2021-02-08
Refactor StorableResources

DanRStevens:
I was investigating using C++20, and the only problem seemed to be in `StorableResources`, so I refactored it a bit.

As we don't have unit tests covering that class, the changes may bear a little extra scrutiny.

For some background on re-writing the `operator` delegation (`<=`, `<`, `>=`, `>`), see:
https://stackoverflow.com/questions/38579088/c-why-implement-op-in-terms-of-op-and-not-the-other-way-around

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -269,12 +269,12 @@ void MapViewState::updatePlayerResources()
 	 */
 	auto& playerResources = const_cast<StorableResources&>(mPlayerResources);
 
-	playerResources.clear();
-
+	StorableResources resources;
 	for (auto structure : storage)
 	{
-		playerResources += structure->storage();
+		resources += structure->storage();
 	}
+	playerResources = resources;
 }
 
 
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -249,9 +249,9 @@ int MapViewState::totalStorage(Structure::StructureClass structureClass, int cap
 int MapViewState::refinedResourcesInStorage()
 {
 	int total = 0;
-	for (size_t i = 0; i < mPlayerResources.resources.size(); ++i)
+	for (size_t i = 0; i < mResourcesCount.resources.size(); ++i)
 	{
-		total += mPlayerResources.resources[i];
+		total += mResourcesCount.resources[i];
 	}
 	return total;
 }
@@ -267,7 +267,7 @@ void MapViewState::updatePlayerResources()
 	{
 		resources += structure->storage();
 	}
-	mPlayerResources = resources;
+	mResourcesCount = resources;
 }
 
 
@@ -1115,9 +1115,9 @@ void MapViewState::placeStructure()
 		}
 
 		// Check build cost
-		if (!StructureCatalogue::canBuild(mPlayerResources, mCurrentStructure))
+		if (!StructureCatalogue::canBuild(mResourcesCount, mCurrentStructure))
 		{
-			resourceShortageMessage(mPlayerResources, mCurrentStructure);
+			resourceShortageMessage(mResourcesCount, mCurrentStructure);
 			return;
 		}
 
@@ -1130,7 +1130,7 @@ void MapViewState::placeStructure()
 		if (structure->isFactory())
 		{
 			static_cast<Factory*>(structure)->productionComplete().connect(this, &MapViewState::factoryProductionComplete);
-			static_cast<Factory*>(structure)->resourcePool(&mPlayerResources);
+			static_cast<Factory*>(structure)->resourcePool(&mResourcesCount);
 		}
 
 		auto cost = StructureCatalogue::costToBuild(mCurrentStructure);
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -262,19 +262,12 @@ void MapViewState::updatePlayerResources()
 	StructureList storage = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::Storage);
 	storage.insert(storage.begin(), mTileMap->getTile(ccLocation(), 0).structure());
 
-	/**
-	 * This at first looks heinous. However, mPlayerResources is declared const
-	 * to ensure that it is treated as a read-only object. This function is the
-	 * only one allowed to modify it.
-	 */
-	auto& playerResources = const_cast<StorableResources&>(mPlayerResources);
-
 	StorableResources resources;
 	for (auto structure : storage)
 	{
 		resources += structure->storage();
 	}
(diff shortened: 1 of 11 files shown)


ldicker83 (commented):
I'm not opposed to this change, though I don't think it's as ambiguous as one might expect since STL containers use `empty()`. It is easier to read when looking over code using it since it follows natural language patterns (hence not opposing the change).

We'll want to consider doing this with all classes in NAS2D and OPHD as a case for refactor going forward. Low priority but it makes sense to have everything consistent.

OPHD/States/MapViewState.cpp
```diff
@@ -269,12 +269,12 @@ void MapViewState::updatePlayerResources()
...
 
-	playerResources.clear();
-
+	StorableResources resources;
 	for (auto structure : storage)
 	{
-		playerResources += structure->storage();
+		resources += structure->storage();
 	}
+	playerResources = resources;
```
ldicker83: Makes sense to me.

OPHD/States/MapViewState.h
```diff
@@ -230,7 +230,7 @@ class MapViewState : public Wrapper
 	int mFood{ 0 };
 
 	// POOLS
-	const StorableResources mPlayerResources; /**< Player's current refined resources. */
+	StorableResources mPlayerResources; /**< Player's current refined resources. */
```
ldicker83, line 233: This should be const. mPlayerResources is used to provide feedback to the player of what they have in storage, it is not a pool to be drawn from. By making it const it prevents accidentally drawing from that pool.

This could be worked around by tallying up the players's current resources in a different way, I just use the StorableResources class as a sort of shortcut.

ldicker83:
So aside from my last comment about `mPlayerResources` being `const`, this should be good to go. There's probably a better way to do what I was doing with `mPlayerResources` but it really should be `const` because it's never supposed to be drawn from. It's purely a tally of the players current resources and the `const` works for me as a debug aid if any code attempts to change it besides that one specific place.

OPHD/States/MapViewState.h
DanRStevens: Do you mean it should be a const reference to data stored elsewhere? Because that's not how it's used here. It is actually being written to.

Perhaps we could change how resources are recalculated and revisit this. Currently though, there is no feasible way to make that field `const` unless we move responsibility to recalculating that value outside of the class and make it a `const StorableResources&` field.

DanRStevens:
I feel like that's a bit of an abuse of `const`. It would make more sense as a debug aid to temporarily mark the field as `const` and see if there are any unintended writes to it. I don't think it should be marked as `const` and then casted just to write to it. Casting away `const` and then writing to it is technically undefined behaviour.

The main purpose of `const_cast` is to pass `const` values to functions where parameters were not marked as `const` (perhaps in legacy code) but ideally should have been marked as `const`. In that case, the parameters are not being written to, so no undefined behaviour is invoked when you cast away `const` to pass a value to such a function.

----

Perhaps there is a deeper architectural issue that needs to be discussed here over why you wanted to mark that value as `const`. I'm afraid I don't quite have the full high level context to discuss that though.

DanRStevens:
Well, I can't find it now, but I did once come across a Stack Overflow question where someone had mistakenly used `empty()` instead of `clear()`, so it can happen. Plus I've come across naming recommendations that have called out early C++ standard methods for bad practice, and I believe `empty()` was one of their examples. I'm inclined to agree. Especially since a quick glance over the method names can and does cause people to confuse the two.

DanRStevens:
Was just thinking, maybe you want to call the field `mPlayerResourcesCache` if that's all it really is. The name would help indicate that it's not the primary source of resources and should not be drawn from.

A cache value though is necessarily not `const`. If it was `const`, it couldn't be updated without invoking undefined behaviour.

OPHD/States/MapViewState.h
ldicker83: I know it's being written to. That was the intent, to write only in that part of the code so that we're not counting resources stored in each structure every frame to get a tally.

This is literally a shortcut. This could be done better by just having a separate array in `MapViewState`, something like `std::array<int, 4> mResourceCount` and then just writing to that each update. That's probably the 'correct' way to do it but I'm lazy.

ldicker83:
> Was just thinking, maybe you want to call the field `mPlayerResourcesCache` if that's all it really is. The name would help indicate that it's not the primary source of resources and should not be drawn from.
> 
> A cache value though is necessarily not `const`. If it was `const`, it couldn't be updated without invoking undefined behaviour.

I wouldn't be opposed to that. Really marking it as const is a hold over from changing the way resources were stored. Originally all of the players resources were stored in that value. Each turn that value would be updated. Before the v0.8.0 tag I moved the actual storage of resources into individual structures. This meant that the mPlayerResources value became useless and didn't actually pull resources from where they were being stored. So I instead marked it const and used it as a tally for the display in the HUD.

This really could easily be changed to anything else. At the time it was to ensure that I'd gotten all the areas mPlayerResources was being used for actual storage. It was part of an iterative change set that I never got around to replacing with 'correct' code (see above comment on a much better way to do it).

DanRStevens:
I'm not sure there's much to gain by using a raw `std::array<int, 4>` instead of a `StorableResources` in `MapViewState`. The class was meant to represent a collection of resources, which that field is. It makes total sense to use that class, even if you don't intend to use all of the operations it provides.

This seems more like a scope and accessibility concern, rather than a `const` concern. I think your real intent here is to limit the scope of when that value is accessible for writing, rather than preventing all writes to it.

----

This might be an area to revisit for refactoring as a separate issue. Sounds like there have been changing requirements that may require an updated design. Not too sure what that would be at the moment though.

ldicker83:
> I'm not sure there's much to gain by using a raw `std::array<int, 4>` instead of a `StorableResources` in `MapViewState`. The class was meant to represent a collection of resources, which that field is. It makes total sense to use that class, even if you don't intend to use all of the operations it provides.
> 
> This seems more like a scope and accessibility concern, rather than a `const` concern. I think your real intent here is to limit the scope of when that value is accessible for writing, rather than preventing all writes to it.
> 
> This might be an area to revisit for refactoring as a separate issue. Sounds like there have been changing requirements that may require an updated design. Not too sure what that would be at the moment though.

You're _way_ over thinking this.

I don't care if it's a `StorableResource` or a `std::array<int, 4>` or four separate `int` variables. Implementation details are irrelevant. The point of it being `const` at all was a tool to help me find where I needed to make changes to the code when I was changing how the game dealt with moving around resources and that was it. It should have been changed after that to a non-`const` member and renamed to something more specific but I forgot to do that (months ago).

Let's just call it `mResourcesCount`, leave it as a non-`const` `StorableResources`, merge the changes and move on.

PR #764, opened 2021-02-08
Fix exception with uncontrolled robot

### LINKED ISSUE #704: Exception thrown in RobotCommand::removeRobot()
Discovered a logic error during testing -- when a Miner is placed onto the map before a Robot Command Center is built and its task completes _after_ the RCC is built, removal of the robot will throw an exception.

This is caused by an assumption that if there's a Robot Command Center, that any robot being removed from the game will always be under RCC control which in the beginning of the game may not yet be true.

This occurs with the Miner early on because the robot is consumed to build the mine. This issue is extremely unlikely to happen with Dozer and Digger robots as they're generally lost when they 'die' out in the field though it will be possible to lose them before an RCC is built when disasters are implemented. Resolving this issue should prevent problems like this going forward.

ldicker83:
Fixes #704

### DIFF
--- a/OPHD/Things/Structures/RobotCommand.cpp
+++ b/OPHD/Things/Structures/RobotCommand.cpp
@@ -53,22 +53,15 @@ void RobotCommand::addRobot(Robot* robot)
 
 
 /**
- * Removes a robot to the management pool of the Robot Command structure.
- *
- * \param	robot	Pointer to a Robot to remove from command list.
+ * Removes a robot from the management pool of the Robot Command structure.
  */
 void RobotCommand::removeRobot(Robot* robot)
 {
-	if (mRobotList.empty())
-	{
-		throw std::runtime_error("RobotCommand::removeRobot(): Robot list empty.");
-	}
+	if (!robot) { throw std::runtime_error("RobotCommand::removeRobot() called with nullptr"); }
+	if (mRobotList.empty()) { return; }
 
 	auto iter = find(mRobotList.begin(), mRobotList.end(), robot);
-	if (iter == mRobotList.end())
-	{
-		throw std::runtime_error("RobotCommand::removeRobot(): Removing a robot that is not under the command of this Robot Command Facility.");
-	}
+	if (iter == mRobotList.end()) { return; }
 
 	mRobotList.erase(iter);
 }
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -1213,7 +1213,6 @@ void MapViewState::updateRobots()
 				tile->removeThing();
 			}
 
-			/// \fixme	Brute force.
 			for (auto rcc : Utility<StructureManager>::get().structureList(Structure::StructureClass::RobotCommand))
 			{
 				static_cast<RobotCommand*>(rcc)->removeRobot(robot);


OPHD/Things/Structures/RobotCommand.cpp
```diff
@@ -53,22 +53,15 @@ void RobotCommand::addRobot(Robot* robot)
...
+ * Removes a robot from the management pool of the Robot Command structure.
  */
 void RobotCommand::removeRobot(Robot* robot)
 {
-	if (mRobotList.empty())
-	{
-		throw std::runtime_error("RobotCommand::removeRobot(): Robot list empty.");
-	}
+	if (!robot) { throw std::runtime_error("RobotCommand::removeRobot() called with nullptr"); }
+	if (mRobotList.empty()) { return; }
```
DanRStevens, line 61: If the parameter was `Robot&` rather than `Robot*`, it would be impossible to pass `nullptr` without really sick pointer casts and use of undefined behaviour. References are non-nullable. I'd recommend changing the function signature and getting rid of the `!robot` check.

----

Most methods can be written to work correctly for empty lists without explicit checks for them. I believe that is the case here. We should strive to get rid of the `mRobotList.empty()` check.

OPHD/Things/Structures/RobotCommand.cpp
```diff
@@ -53,22 +53,15 @@ void RobotCommand::addRobot(Robot* robot)
...
 
 	auto iter = find(mRobotList.begin(), mRobotList.end(), robot);
-	if (iter == mRobotList.end())
-	{
-		throw std::runtime_error("RobotCommand::removeRobot(): Removing a robot that is not under the command of this Robot Command Facility.");
-	}
+	if (iter == mRobotList.end()) { return; }
 
 	mRobotList.erase(iter);
```
DanRStevens, line 66: Perhaps we should make use of [`std::remove_if`](https://en.cppreference.com/w/cpp/algorithm/remove). One method call should be able to replace this block of code.

```cpp
remove_if(mRobotList.begin(), mRobotList.end(), robot);
```

DanRStevens (commented):
I think we can simplify the code further.

OPHD/Things/Structures/RobotCommand.cpp
ldicker83: I forgot remove_if is a thing. Not sure which change set to make this a part of or if it should just be a refactor thing.

OPHD/Things/Structures/RobotCommand.cpp
ldicker83: We could do it as that but robots are all stored as pointers. I could just dereference it and pass it along that way but if using remove_if wouldn't it work fine with a nullptr anyway?

DanRStevens: Yes, you could add a `*` before the parameter at the call point.

You may have a point though, that `remove_if` may work with `nullptr`, so this becomes a non-issue.

OPHD/Things/Structures/RobotCommand.cpp
DanRStevens: I would handle with an independent followup PR to do the refactoring.

PR #765, opened 2021-02-09
Null derefernce guard in rcc read

ldicker83:
Small refactor of the RCC xml read code that adds in some basic guards so things aren't crashy if a user decides to make changes in a save game file.


DanRStevens:
I like the organization on this one.

PR #766, opened 2021-02-09
Improve remove robot function

ldicker83:
>Perhaps we should make use of [`std::remove_if`](https://en.cppreference.com/w/cpp/algorithm/remove). One method call should be able to replace this block of code.

```cpp
remove_if(mRobotList.begin(), mRobotList.end(), robot);
```

>Edit: It also runs in a way that enables us to remove the `empty()` check.

_Originally posted by @DanRStevens in https://github.com/OutpostUniverse/OPHD/pull/764#discussion_r572463784_

Implements this suggestion.

### DIFF
--- a/OPHD/Things/Structures/RobotCommand.cpp
+++ b/OPHD/Things/Structures/RobotCommand.cpp
@@ -28,11 +28,6 @@ bool RobotCommand::isControlling(Robot* robot) const
 
 /**
  * Adds a robot to the management pool of the Robot Command structure.
- * 
- * \param	robot	Pointer to a Robot to add to command list.
- * 
- * \note	Performs some basic sanity checking. Will throw if list is full or
- *			the given pointer to a Robot is already in the list.
  */
 void RobotCommand::addRobot(Robot* robot)
 {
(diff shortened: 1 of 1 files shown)


OPHD/Things/Structures/RobotCommand.cpp
```diff
@@ -57,11 +57,5 @@ void RobotCommand::addRobot(Robot* robot)
...
 void RobotCommand::removeRobot(Robot* robot)
 {
-	if (!robot) { throw std::runtime_error("RobotCommand::removeRobot() called with nullptr"); }
-	if (mRobotList.empty()) { return; }
-
-	auto iter = find(mRobotList.begin(), mRobotList.end(), robot);
-	if (iter == mRobotList.end()) { return; }
-
-	mRobotList.erase(iter);
+	mRobotList.erase(std::remove_if(mRobotList.begin(), mRobotList.end(), [robot](Robot* r1) { return r1 == robot; }), mRobotList.end());
```
DanRStevens, line 53: Oh yeah, I forgot the `std::erase`. I find that's an area where the STL is not super obvious.

----

If the lambda is just testing for equality, I believe the value overload could have been used instead.
```cpp
mRobotList.erase(std::remove_if(mRobotList.begin(), mRobotList.end(), robot), mRobotList.end());
```
See overload (1) instead of overload (3): [`std::remove`](https://en.cppreference.com/w/cpp/algorithm/remove)

Just noticing now I should have said `std::remove` earlier instead of `std::remove_if`.

DanRStevens (approved):
Looks good. Makes the code much more compact.

Noticed one small change that could shorten it a tiny bit more. Seems I used the wrong function name in my earlier suggestion.

OPHD/Things/Structures/RobotCommand.cpp
ldicker83: Yeah, I thought that `std::remove_if` was overkill? But I wanted to practice my lambda expressions since I'm still a little unfamiliar with them. Going to commit the change with `std::remove` and call this one done. :D

DanRStevens (approved):
Looks good to me

PR #769, opened 2021-02-14
Prioritize resource movement to command

### LINKED ISSUE #758: Have resources moved/stored in Command Center first if CC isn't at capacity
Something I noticed is that if the Command Center's stores are used during a time of crisis, the stores are never replenished afterward. This would lead to rapid colony collapse should a similar crisis occur and could result in an instant game-over for the player when these are basically reserve resources that should always be replenished.

Changes should be simple: Prioritize refined resources and food storage at the CC before storing at storage tanks or agridomes.

ldicker83:
Closes #758

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -372,7 +372,10 @@ void MapViewState::countFood()
 {
 	mFood = 0;
 
-	const auto& structures = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::FoodProduction);
+	auto structures = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::FoodProduction);
+	auto& command = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::Command);
+
+	structures.insert(structures.begin(), command.begin(), command.end());
 
 	for (auto structure : structures)
 	{
@@ -381,8 +384,6 @@ void MapViewState::countFood()
 			mFood += static_cast<FoodProduction*>(structure)->foodLevel();
 		}
 	}
-
-	mFood += static_cast<CommandCenter*>(mTileMap->getTile(ccLocation(), 0).structure())->foodLevel();
 }
 
 
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -463,14 +463,14 @@ void MapViewState::nextTurn()
 	checkConnectedness();
 	NAS2D::Utility<StructureManager>::get().update(mResourcesCount, mPopulationPool);
 
-	transferFoodToCommandCenter();
-	countFood();
-
 	mPreviousMorale = mCurrentMorale;
 
+	transferFoodToCommandCenter();
 	updateResidentialCapacity();
 
 	updatePopulation();
+	countFood();
+
 	updateCommercial();
 	updateBiowasteRecycling();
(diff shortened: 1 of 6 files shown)


OPHD/States/MapViewStateTurn.cpp
```diff
@@ -388,7 +388,32 @@ void MapViewState::countFood()
...
+		while (foodProducerIterator != foodProducers.end())
+		{
+			auto foodProducer = static_cast<FoodProduction*>(*foodProducerIterator);
+			const int foodMoved = std::clamp(foodToMove, 0, foodProducer->foodLevel());
+			foodProducer->foodLevel(foodProducer->foodLevel() - foodMoved);
+			commandCenter->foodLevel(commandCenter->foodLevel() + foodMoved);
+
+			foodToMove -= foodMoved;
+
+			if (foodToMove == 0) { return; }
```
DanRStevens, line 412: This should probably break out of the `for` loop, rather than `return`. There may be other command centers that can accept food.

Unless of course you want to limit mechanics to filling at most one CC per turn.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -372,7 +372,10 @@ void MapViewState::countFood()
 {
 	mFood = 0;
 
-	const auto& structures = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::FoodProduction);
+	auto structures = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::FoodProduction);
+	auto& command = NAS2D::Utility<StructureManager>::get().structureList(Structure::StructureClass::Command);
```
DanRStevens, line 376: Nothing wrong here, just wanted to comment since it wasn't obvious and I had to look it up. The `auto` keyword in `auto& command = ...` will deduce `const`. Technically the `const` in the removed line `const auto& structures = ...` could have been removed. If the reference was not there, the situation would be different as a copy would be made, and a copy is not required to retain the const-ness of the original.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -88,17 +88,17 @@ void MapViewState::updatePopulation()
 	int nurseries = structureManager.getCountInState(Structure::StructureClass::Nursery, StructureState::Operational);
 	int hospitals = structureManager.getCountInState(Structure::StructureClass::MedicalCenter, StructureState::Operational);
 
-	const auto& foodproducers = structureManager.structureList(Structure::StructureClass::FoodProduction);
+	auto foodproducers = structureManager.structureList(Structure::StructureClass::FoodProduction);
+	auto& command = structureManager.structureList(Structure::StructureClass::Command);
```
DanRStevens, line 92: Ditto.

DanRStevens (commented):
Noticed what looks like a small oversight.

Also commented on a couple of lines that are actually correct, just that were not obvious at first glance. Figured I'd leave a comment for posterity.

OPHD/States/MapViewStateTurn.cpp
ldicker83: That's a good point that I hadn't considered. I'll make it a task for another version (since atm there's only ever the one command center).

OPHD/States/MapViewStateTurn.cpp
ldicker83: This is a good point. It's easy enough for me to know because Visual Studio does show how it evaluates so I know it's const. I may change that later for the sake of being obvious.

PR #770, opened 2021-02-14
Renames updatePlayerResources() to the less ambiguous countPlayerReso…

ldicker83:
…urces()

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -257,7 +257,7 @@ int MapViewState::refinedResourcesInStorage()
 }
 
 
-void MapViewState::updatePlayerResources()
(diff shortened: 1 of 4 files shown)


DanRStevens:
Looks like this branch was merged and then deleted before CI could complete. CI shows a build failure, though it appears to be caused by the underlying branch disappearing before the x86 configuration could start, and not because of a code problem.

ldicker83:
Yeah, that was my mistake. It was the Windows build which I already know to work which is why I'm not worried about it. I didn't realize appveyor was still running (this happens when I'm too tired to be doing this).

That and it was a simple enough change to not matter. I remember my thought process being "I already know it builds on windows so merge/delete"

PR #756, opened 2021-02-04
Biowaste code refactor

DanRStevens:
Suggestions from #753

PR #771, opened 2021-02-14
Add residenceIterator increment operator so that all residences are t…

### LINKED ISSUE #768: Biowaste is not removed from all residences during a turn
Noticed that after recent changes biowaste still accumulated in all but one residence. Quick glance at the code and the problem appears to be that the residence iterator is never incremented. Whoops.

Noting it here for posterity so I don't forget since I seem to not be able to sleep lately.

ldicker83:
…raversed instead of only the first residence in the list

Closes #768

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -363,6 +363,7 @@ void MapViewState::updateBiowasteRecycling()
 
 			Residence* residence = static_cast<Residence*>(*residenceIterator);
 			residence->pullWaste(recycling->wasteProcessingCapacity());
+			++residenceIterator;
 		}
 	}
 }


OPHD/States/MapViewStateTurn.cpp
```diff
@@ -363,6 +363,7 @@ void MapViewState::updateBiowasteRecycling()
 
 			Residence* residence = static_cast<Residence*>(*residenceIterator);
 			residence->pullWaste(recycling->wasteProcessingCapacity());
+			++residenceIterator;
```
DanRStevens, line 366: Whoops, that was me. I introduced this bug during the attempt at refactoring in dff1b5e25286c8461e96cbfc8aac0b80d5ca3230 (PR #756). Thank you for catching and fixing it.

ldicker83: I know, I was going to use git blame and then thought, nah, it's an easy mistake to make and I didn't catch it when reading the code, only when I noticed that biowaste wasn't being removed from residences even though I had a recycling facility. Looked at the code and noticed we never traversed the container.

PR #776, opened 2021-02-16
Update population panel

### LINKED ISSUE #104: Add a morale change breakdown to the morale report window.
Forums user jballou suggested adding a breakdown of morale changes: https://forum.outpost2.net/index.php/topic,5718.msg88332.html#msg88332

This should be a fairly straight-forward change, I'll just need to find a good place to stick this in (update the population panel? maybe make it wider and make the population breakdown spread side to side in the panel instead of list straight down?) Will have to play with that a bit.

ldicker83:
Closes #104

### DIFF
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -116,6 +116,22 @@ const std::map<IdleReason, std::string> IDLE_REASON_TABLE =
 };
 
 
+const std::array<std::string, 5> MoraleStringTable =
+{
+	constants::MoraleTerrible,
+	constants::MoralePoor,
+	constants::MoraleFair,
+	constants::MoraleGood,
+	constants::MoraleExcellent
+};
+
+
+const std::string& moraleDescription(int index)
+{
+	return MoraleStringTable[index];
+}
+
+
 /**
  * Description table for products.
  */
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -116,19 +116,43 @@ const std::map<IdleReason, std::string> IDLE_REASON_TABLE =
 };
 
 
-const std::array<std::string, 5> MoraleStringTable =
+const std::array MoraleStringTable =
 {
-	constants::MoraleTerrible,
-	constants::MoralePoor,
-	constants::MoraleFair,
-	constants::MoraleGood,
-	constants::MoraleExcellent
+	std::string("Terrible"),
+	std::string("Poor"),
+	std::string("Fair"),
+	std::string("Good"),
+	std::string("Excellent"),
+
+	std::string("Morale is "),
+	std::string("Births"),
+	std::string("Deaths"),
+	std::string("No active Food Production"),
+	std::string("Parks & Arboretums"),
+	std::string("Recreational Facilities"),
+	std::string("Luxury Availability"),
+	std::string("Residential Over Capacity"),
+	std::string("Biowaste Overflowing"),
+	std::string("Structures Disabled"),
+	std::string("Structures Destroyed")
 };
 
 
-const std::string& moraleDescription(int index)
+const std::string& moraleString(int index)
 {
 	return MoraleStringTable[index];
+}
+
+
+const std::string& moraleString(Morale m)
+{
+	return MoraleStringTable[static_cast<int>(m)];
+}
+
+
+int moraleStringTableCount()
+{
+	return MoraleStringTable.size();
 }
 
 
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -152,7 +152,7 @@ const std::string& moraleString(Morale m)
 
 int moraleStringTableCount()
 {
-	return MoraleStringTable.size();
+	return static_cast<int>(MoraleStringTable.size());
 }
 
 
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -367,3 +367,9 @@ int pullTruckFromInventory();
  * \return 1 on success, 0 otherwise.
  */
 int pushTruckIntoInventory();
+
+
+const auto formatDiff = [](int diff)
+{
+	return ((diff > 0) ? "+" : "") + std::to_string(diff);
+};
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -272,6 +272,28 @@ enum ProductType
 };
 
 
+enum class Morale
+{
+	Terrible,
+	Poor,
+	Fair,
+	Good,
+	Excellent,
+
+	Description,
+	Births,
+	Deaths,
+	NoFoodProduction,
+	Parks,
+	Recreation,
+	Commercial,
+	ResidentialOverflow,
+	BiowasteOverflow,
+	StructuresDisabled,
+	StructuresDestroyed
+};
+
+
 /**
  * Contains population requirements for a given Structure.
  * 
@@ -320,7 +342,9 @@ const std::string& idleReason(IdleReason);
 /**
  * Gets a morale string based on value;
  */
-const std::string& moraleDescription(int);
+const std::string& moraleString(int);
+const std::string& moraleString(Morale);
+int moraleStringTableCount();
 
 /**
  * Super basic progress bar.
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -318,6 +318,11 @@ const std::string& disabledReason(DisabledReason);
 const std::string& idleReason(IdleReason);
 
 /**
+ * Gets a morale string based on value;
+ */
+const std::string& moraleDescription(int);
+
+/**
  * Super basic progress bar.
  */
 void drawBasicProgressBar(int x, int y, int width, int height, float percent, int padding = 4);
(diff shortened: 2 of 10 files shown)


OPHD/States/MapViewStateTurn.cpp
```diff
@@ -165,36 +165,57 @@ void MapViewState::updateMorale()
...
+		if (unit->wasteOverflow() > 0) { ++bioWasteAccumulation; }
 	}
 
-	mCurrentMorale -= residentialMoraleHit;
+	// positive
+	mCurrentMorale += birthCount;
+	mCurrentMorale += parkCount;
+	mCurrentMorale += recreationCount;
+	mCurrentMorale += commercialCount;
+	mCurrentMorale += foodProducingStructures > 0 ? 0 : -5;
```
DanRStevens, line 196: The `foodProducingStructures` line is actually a negative. Might be more clearly a negative if it was written as:
```cpp
mCurrentMorale -= foodProducingStructures > 0 ? 0 : 5;
```

Btw, any thoughts on if it's more clear or not to include parenthesis around ternary conditions?
```cpp
mCurrentMorale -= (foodProducingStructures > 0) ? 0 : 5;
```

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -165,36 +165,57 @@ void MapViewState::updateMorale()
...
+	mCurrentMorale -= bioWasteAccumulation * 2;
+	mCurrentMorale -= structuresDisabled;
+	mCurrentMorale -= structuresDestroyed;
+
 
 	mCurrentMorale = std::clamp(mCurrentMorale, 0, 1000);
+
+	// Provide reason values for PopulationPanel
+	if (birthCount > 0) { mPopulationPanel.addMoraleReason("Births", birthCount); }
+	if (deathCount > 0) { mPopulationPanel.addMoraleReason("Deaths", -deathCount); }
```
DanRStevens: There is effectively some code duplication here. The morale benefits and penalties from the section above are being duplicated in this section in the second parameter to `addMoraleReason`. If instead the `addMoraleReason` function was responsible for updating the running tally, that could ensure consistency between the two, without code duplication. As it is now, it would be very easy to update only one section with a new morale modifier, or make a typo between the two sections, and so the morale reasons might not match up with actual morale modifiers.

----

Edit: Looks like you figured this out in a later commit.

Additionally, the `if` checks are another area of duplicate code. Effectively the `if` checks are checking for non-zero morale modifiers, and if it's non-zero, to list the text reason along with the modifier value. Instead, the `if` check can be moved into `addMoraleReason`. This of course means a slight update to the `if` condition, since the original code infers a non-zero modifier from a base value, rather than from the morale modifier calculated from it.

To be more concrete, here's an example where the base value `residentialOverCapacityHit` is different from the morale modifier `bioWasteAccumulation * -2`  (wait, why are these different variables?):
```cpp
if (residentialOverCapacityHit > 0) { mPopulationPanel.addMoraleReason("Biowaste Overflowing", bioWasteAccumulation * -2); }
```

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -202,18 +201,17 @@ void MapViewState::updateMorale()
 	mCurrentMorale -= structuresDisabled;
 	mCurrentMorale -= structuresDestroyed;
 
-
 	mCurrentMorale = std::clamp(mCurrentMorale, 0, 1000);
 
-	// Provide reason values for PopulationPanel
+	mPopulationPanel.clearMoraleReasons();
 	if (birthCount > 0) { mPopulationPanel.addMoraleReason("Births", birthCount); }
 	if (deathCount > 0) { mPopulationPanel.addMoraleReason("Deaths", -deathCount); }
```
DanRStevens: You could potentially even initialize a new morale reasons list, and then assign it to the population panel at the end. That would have the effect of simultaneously clearing the text list of morale reasons, and any running tally of morale modifiers it might accumulate.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -14,6 +14,39 @@
...
+
+
+static const std::array moraleStringColor
+{
+	Color::Red,
+	Color::Orange,
+	Color::Yellow,
+	Color::White,
+	Color{ 0, 185, 0 }
+};
```
DanRStevens: These could be moved into local variables in the function where they are used.

OPHD/Common.cpp
```diff
@@ -116,22 +116,46 @@ const std::map<IdleReason, std::string> IDLE_REASON_TABLE =
...
-const std::array<std::string, 5> MoraleStringTable =
-{
-	constants::MoraleTerrible,
-	constants::MoralePoor,
-	constants::MoraleFair,
-	constants::MoraleGood,
-	constants::MoraleExcellent
+const std::array MoraleStringTable =
+{
+	std::string("Terrible"),
```
DanRStevens: I approve of cutting out the intermediate variables.

OPHD/Common.h
```diff
@@ -367,3 +367,9 @@ int pullTruckFromInventory();
  * \return 1 on success, 0 otherwise.
  */
 int pushTruckIntoInventory();
+
+
+const auto formatDiff = [](int diff)
+{
+	return ((diff > 0) ? "+" : "") + std::to_string(diff);
+};
```
DanRStevens, line 375: ... or make them global functions.

In which case, these should probably be functions, rather than lambdas.

DanRStevens (commented):
Some thoughts for future consideration.

OPHD/States/MapViewStateTurn.cpp
ldicker83: I would agree that parenthesis around ternary operators is generally more clear and is usually the way I try to write them. I would consider this a stylistic choice that should be encouraged.

And yes, it would make sense to write this in the negative. I think I was considering having a positive effect on growing food stores but that was unrelated to the number of actual food production structures versus actually seeing a positive trend, it just didn't translate over.

I think I'm going to open a separate change set / pr for code audits like these to clean up the things you notice that could be done better.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -500,5 +500,8 @@ void MapViewState::nextTurn()
 		mGameOverDialog.show();
 	}
 
+	mPopulationPanel.morale(mCurrentMorale);
+	mPopulationPanel.old_morale(mPreviousMorale);
```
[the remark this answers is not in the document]
ldicker83: Didn't I split it into smaller functions? Which is why it calls a long list of them? :D 

Your suspicions are correct. `mCurrentMorale` is actually updated in the function `updateMorale()`. The lines you noted could and probably should all go into the `updateMorale()` function. They don't need to be separated by as much logic as they are and that mostly comes from thinking about the larger picture and not at the finer details (which is why I really like having a second set of eyes on the code).

OPHD/States/MapViewStateTurn.cpp
ldicker83: This would be a mistake that I'm pretty sure I already ironed out in a later commit? Gonna have to double check that. (double checked, it is ironed out in a later commit).

OPHD/UI/PopulationPanel.cpp
ldicker83: I really hate doing that. It unnecessarily bloats the function when they could just be statically declared in the module in which they're used.

OPHD/Common.h
ldicker83: Common is 'global'. I figured it's such a basic thing that it could easily be in-lined into wherever they need to be. I suppose the compiler can figure that out on its own with an `inline` hint but eh. Again, more practice with lambda syntax.

OPHD/Common.cpp
ldicker83: Generally same here, I'd ultimately like to get rid of the Strings.h file and maybe use some sort of a global 'dictionary' that can load these values from a file at run-time.

I'm not a big fan of string literals in the code and prefer to either use named constants or a similar method to get them. It brings up the concern of non-latin language localization. Something that will be _really_ hard if not impossible with the current way things are handled (std::string really has no method of dealing with utf8 encoding and NAS2D's Font and text rendering don't handle it at all instead sticking with the ASCII definition of character sets... which could be fudged to localize but eh).

Anyway, yeah, the ultimate goal here for me is to move all of this from out of the `constants` namespace and into an object that does better translation using `class enum` values instead.

DanRStevens: A while back I was looking at the source code for [A Dark Room](https://github.com/doublespeakgames/adarkroom). I noticed they had an interesting translation mechanism. They had English strings in the code, but being looked up in a translation table. If nothing was loaded, the English would feed through to the output. Otherwise, a dictionary lookup would effectively translate each string and use that for the output. Looked like it made it pretty obvious to write the game for English speaking developers, while still allowing all game test to be translated with external translation files. The translation files paired up English phrases with the translated phrase.

Of course this sort of lookup is not context aware, so if you have the same English word used in two different ways in two different places, there would be no ability to translate it differently in those two places in other languages. Mind you, most strings were quite unique, so there was little risk of that happening in most places.

----

Luckily UTF-8 was designed to work well with most non-UTF-8 aware code. Aside from the `Font` rendering, it really shouldn't matter. It would also be possible to upgrade the `Font` rendering without disturbing any external code. The `Font` rendering interface is pretty clean in that regard. There would be no project wide updates needed to all calls to string rendering to make such an upgrade work. It would be very self contained.

ldicker83: That's a neat idea but I think ultimately I'd rather go with a basic approach of using some sort of enumerated value that's used to lookup a string through a table. Simple, does the job without making too many assumptions, etc.

I'm more worried about the font rendering code than anything else? I'm not completely sure I know how to handle it but it would be a great feature moving forward.

PR #777, opened 2021-02-17
Add tool tip

### LINKED ISSUE #747: Add timed hover tooltip to GUI Control's
So far none of the GUI elements in the game provide any sort of hover tool tip. It's time to add this.

Much like any modern OS, the idea here is to allow for a tooltip that's displayed a set number of seconds after the pointer enters the bounds of a control.

I'm thinking of doing this in a similar fashion to .Net's method using [code like this](https://docs.microsoft.com/en-us/dotnet/desktop/winforms/controls/how-to-set-tooltips-for-controls-on-a-windows-form-at-design-time?view=netframeworkdesktop-4.8):

```
toolTip1.SetToolTip(button1, "Save changes");
```

My first thought was to just add a property to Control and then have the control implement a Timer. I'm now thinking to do it as it's implied in the above code with a ToolTip GUI element, adding Control's to it with an associated tooltip text message, and then let the ToolTip control handle the timing, drawing, and positioning. Seems like a very clean way to handle this without modifying the base Control.

ldicker83:
Closes #747

### DIFF
--- a/OPHD/States/MapViewState.h
+++ b/OPHD/States/MapViewState.h
@@ -249,6 +249,15 @@ private:
 	Button mBtnToggleHeightmap;
 	Button mBtnToggleConnectedness;
 
+	// Bare Control's use for ToolTips
+	Control mTooltipSystemButton;
+	Control mTooltipCurrentTurns;
+	Control mTooltipPopulation;
+	Control mTooltipEnergy;
+	Control mTooltipFoodStorage;
+	Control mTooltipResourceStorage;
+	Control mTooltipResourceBreakdown;
+
 	ToolTip mToolTip;
 
 	IconGrid mStructures{"ui/structures.png", 46, constants::MARGIN_TIGHT};
--- a/OPHD/States/MapViewState.h
+++ b/OPHD/States/MapViewState.h
@@ -249,6 +249,8 @@ private:
 	Button mBtnToggleHeightmap;
 	Button mBtnToggleConnectedness;
 
+	ToolTip mToolTip;
+
 	IconGrid mStructures{"ui/structures.png", 46, constants::MARGIN_TIGHT};
 	IconGrid mRobots{"ui/robots.png", 46, constants::MARGIN_TIGHT};
 	IconGrid mConnections{"ui/structures.png", 46, constants::MARGIN_TIGHT};
--- a/OPHD/UI/Core/ToolTip.cpp
+++ b/OPHD/UI/Core/ToolTip.cpp
@@ -39,6 +39,34 @@ void ToolTip::add(Control& c, const std::string& str)
 }
 
 
+void ToolTip::buildDrawParams(std::pair<Control*, std::string>& item)
+{
+	constexpr int padding = constants::MARGIN_TIGHT * 2;
+
+	const int tooltipWidth = mFont.width(item.second) + padding;
+	const int tooltipHeight = mFont.height() + padding;
+	
+	auto tooltipPosition = item.first->position();
+
+	auto offset = Vector{ 0, -tooltipHeight - constants::MARGIN };
+
+	if (tooltipPosition.y + offset.y < 0)
+	{
+		offset.y = tooltipHeight + constants::MARGIN;
+	}
+
+
+	auto& renderer = Utility<Renderer>::get();
+	if (tooltipPosition.x + tooltipWidth > renderer.size().x)
+	{
+		offset.x -= (tooltipPosition.x + tooltipWidth) - (renderer.size().x - constants::MARGIN);
+	}
+
+	position(tooltipPosition + offset);
+	size({ tooltipWidth, tooltipHeight });
+}
+
+
 void ToolTip::onMouseMove(int x, int y, int dX, int dY)
 {
 	if (dX != 0 || dY != 0)
@@ -54,9 +82,11 @@ void ToolTip::onMouseMove(int x, int y, int dX, int dY)
 
 	for (auto& item : mControls)
 	{
+		if (mFocusedControl) { break; }
 		if (item.first->rect().contains({ x, y }))
 		{
 			mFocusedControl = &item;
+			buildDrawParams(item);
 			return;
 		}
 	}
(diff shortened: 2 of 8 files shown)


OPHD/UI/Core/ToolTip.h
```diff
@@ -0,0 +1,39 @@
...
+
+private:
+	const NAS2D::Font& mFont;
+	NAS2D::Timer mTimer;
+
+	NAS2D::Point<int> mMouseCoords;
+
+	std::pair<Control*, std::string>* mFocusedControl;
+
+	std::vector<std::pair<Control*, std::string>> mControls;
```
DanRStevens, line 38: The resulting code would probably be a little more readable if you defined a `struct` to hold `Control` and `std::string` values as named fields, rather than as `std::pair` with values `first` and `second`.

OPHD/UI/Core/ToolTip.cpp
```diff
@@ -0,0 +1,80 @@
...
+void ToolTip::add(Control& c, const std::string& str)
+{
+	for (auto& item : mControls)
+	{
+		if (item.first == &c)
+		{
+			item.second = str;
+			return;
+		}
+	}
```
DanRStevens, line 36: Much more readable when not using `first` and `second` (nor abbreviations like `c`):
```cpp
	for (auto& item : mControls)
	{
		if (item.control == &targetControl)
		{
			item.toolTipText = str;
			return;
		}
	}
```

OPHD/UI/Core/ToolTip.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+	mControls.push_back(std::make_pair(&c, str));
+}
```
DanRStevens, line 39: You could potentially use a `std::map` with the `Control&` (or `Control*`) as the key. That would eliminate the need to scan the list for updates.

OPHD/States/MapViewState.h
```diff
@@ -249,6 +249,15 @@ class MapViewState : public Wrapper
...
 	Button mBtnToggleConnectedness;
 
+	// Bare Control's use for ToolTips
+	Control mTooltipSystemButton;
+	Control mTooltipCurrentTurns;
+	Control mTooltipPopulation;
+	Control mTooltipEnergy;
+	Control mTooltipFoodStorage;
+	Control mTooltipResourceStorage;
+	Control mTooltipResourceBreakdown;
```
DanRStevens, line 259: I feel like these should belong to corresponding Panel classes, rather than all lumped together here.

DanRStevens (commented):
A few thoughts for future consideration.

OPHD/UI/Core/ToolTip.h
ldicker83: Why build anew when I can instead use what's available in the standard library?

I was using `typedef`'s way back when and eventually `using` directives but those mostly got stripped out because of this kind of reasoning. I dunno, I find that defining a `struct` or `class` just for this sort of case when I could use a `std::pair` or `std::tuple` instead to be overkill.

OPHD/UI/Core/ToolTip.cpp
ldicker83: /me shrugs

I dunno, it's a standard object so to me it's just as readable either way. Though granted, knowing what the pair actually is supposed to be helps. 

For me it's a trivial matter of mousing over a variable and I get this:

image

I'm not sure what your environment is like but if you're using something like vim as a text editor instead of a full-fledged IDE with code completion this could be a bit of a drag to work with.

OPHD/UI/Core/ToolTip.cpp
ldicker83: True, though I don't see too many cases of duplicate insertion. This just seemed like the lighter-weight way to go versus using a binary tree.

OPHD/States/MapViewState.h
ldicker83: These are 'hidden' or 'dummy' controls. Panels have no place in this context.

There is also no such beast in the current GUI implementation.

OPHD/UI/Core/ToolTip.h
DanRStevens: See the second answer, which has the most up votes:
https://stackoverflow.com/questions/3607352/struct-with-2-cells-vs-stdpair

This was also address in The Clean Code Collection, which also suggested using a custom `struct` to improve readability:
https://www.amazon.com/Robert-Martin-Clean-Code-Collection-ebook/dp/B00666M59G

OPHD/UI/Core/ToolTip.cpp
DanRStevens: I use Atom, which looks like most IDE's, though doesn't have mouse over popups, or any real special C++ code support other than syntax highlighting. The only way I can know what `mControls` is, is to go look for the definition. And there's no jump to definition support, so I actually need to search for it.

OPHD/States/MapViewState.h
DanRStevens: Yeah, I think you're right. Sounds like we need an actual Control for these sorts of things.

OPHD/UI/Core/ToolTip.h
ldicker83: Honestly that's a pretty good answer and a strong argument against std::pair/std::tuple. Could be something to look at going forward. Plus, to be fair, if needs change for elements later on it's a trivial matter to add an additional member type.

OPHD/UI/Core/ToolTip.cpp
ldicker83: Which is a valid point. I still think it makes sense to use an IDE that at least allows for this level of code peeking. VS Code may be a better alternative to Atom for this reason.

That stated, your previous answer has me thinking that your suggestion of using `struct`'s over `std::pair` or `std::tuple` is worth taking the time to address.

OPHD/States/MapViewState.h
ldicker83: See #781 for thoughts on an implementation.

PR #780, opened 2021-02-17
Add tooltips to hud

### LINKED ISSUE #778: Add tooltips to main HUD elements

ldicker83:
Closes #778

### DIFF
--- a/OPHD/States/MapViewStateUi.cpp
+++ b/OPHD/States/MapViewStateUi.cpp
@@ -378,7 +378,7 @@ void MapViewState::drawUI()
 	mGameOptionsDialog.update();
 	mWindowStack.update();
 
-	mToolTip.update();
+	if (!modalUiElementDisplayed()) { mToolTip.update(); }
 }
 
 
--- a/OPHD/States/MapViewStateUi.cpp
+++ b/OPHD/States/MapViewStateUi.cpp
@@ -115,10 +115,38 @@ void MapViewState::initUi()
 	// Initial Structures
 	mStructures.addItem(constants::SEED_LANDER, 0, StructureID::SID_SEED_LANDER);
 
+
+	// tooltip control sizes
+	constexpr auto hudHeight = constants::RESOURCE_ICON_SIZE + constants::MARGIN_TIGHT * 2;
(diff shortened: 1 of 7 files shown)


OPHD/States/MapViewStateUi.cpp
```diff
@@ -115,10 +115,38 @@ void MapViewState::initUi()
...
+	mTooltipFoodStorage.size({ 75, hudHeight });
+
+	mTooltipEnergy.position({ 560, 0 });
+	mTooltipEnergy.size({ 55, hudHeight });
+
+	mTooltipPopulation.position({ 670, 0 });
+	mTooltipPopulation.size({ 75, hudHeight });
+
+	mTooltipCurrentTurns.size({ 45, hudHeight });
+	mTooltipSystemButton.size({ hudHeight, hudHeight });
```
DanRStevens, line 136: I'm somewhat concerned by this. It seems odd to be adding dummy controls, though I suspect this is because a lot of GUI rendering was done with raw draw calls, rather than creating controls to draw the UI.

I suppose this might be a sign to move some of that rendering away from raw draw calls and over to controls to draw those items.

DanRStevens (commented):
The "dummy control" part caught my attention.

OPHD/States/MapViewStateUi.cpp
ldicker83: That's exactly what I was thinking. Since the ToolTip interface is implemented in terms of a Control, I needed to do this to add tooltips to the HUD.

Which is a huge, _ugly_ bit of code that could be done _way_ better using some derived Control type that would act in the intended way, perhaps organized into a tool-strip of sorts? Might be overthinking the design there but the basic idea is a 'bar' with a bunch of 'panels' in it with customized content. Then it would simply be a matter of having a ToolBar or whatever with a single call to update() it which would then update all of its 'children' and each of those 'children' could easily be added to the ToolTip. Each 'panel' object would override it's 'draw' function in some way for the customized content. Could maybe sub-in a lambda for it or just derive new types from 'panel' that provide their customized content.

I guess you could call this a placeholder until that's done. It's a semi-complex idea that's out of the context of this PR but definitely something I want to look into going forward.

PR #782, opened 2021-02-17
Fix mine facility no route

### LINKED ISSUE #774: Mine Report shows 'no route' until a turn is processed
Fullscreen UI for mines shows 'no route' until a turn is processed. This should be a fairly straight-forward change in the loading code that calls route processing to find routes before the player is given back control.

image

ldicker83:
Closes #774

### DIFF
--- a/OPHD/UI/Reports/MineReport.cpp
+++ b/OPHD/UI/Reports/MineReport.cpp
@@ -319,6 +319,13 @@ void MineReport::updateManagementButtonsVisiblity()
 	btnAddTruck.visible(mSelectedFacility);
 	btnRemoveTruck.visible(mSelectedFacility);
 
+	if (mSelectedFacility->destroyed() ||
+		mSelectedFacility->underConstruction())
+	{
+		btnAddTruck.visible(false);
+		btnRemoveTruck.visible(false);
+	}
+
 	chkCommonMetals.visible(mSelectedFacility);
 	chkCommonMinerals.visible(mSelectedFacility);
 	chkRareMetals.visible(mSelectedFacility);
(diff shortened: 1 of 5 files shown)


OPHD/UI/Reports/MineReport.cpp
```diff
@@ -335,11 +342,11 @@ void MineReport::lstMineFacilitySelectionChanged()
...
 
 	auto facility = static_cast<MineFacility*>(mSelectedFacility);
-	btnIdle.toggle(facility->state() == StructureState::Idle);
-	btnIdle.enabled(facility->state() == StructureState::Operational || facility->state() == StructureState::Idle);
+	btnIdle.toggle(mSelectedFacility->isIdle());
+	btnIdle.enabled(mSelectedFacility->operational() || mSelectedFacility->isIdle());
 
 	btnDigNewLevel.toggle(facility->extending());
-	btnDigNewLevel.enabled(facility->canExtend() && mSelectedFacility->operational() || mSelectedFacility->isIdle());
+	btnDigNewLevel.enabled(facility->canExtend() && (mSelectedFacility->operational() || mSelectedFacility->isIdle()));
```
DanRStevens: This looks like a bug fix logical change. Probably better to do in it's own commit so the change stands out more.

ldicker83: It was unintentionally omitted from another change set and I noticed it and opted to include it. There's something about single line changes being entire PR's that bother me to some degree. This is still a relatively small project with only a very small set of developers. Much larger projects with dozens or hundreds of developers sure, that makes more sense.

DanRStevens: I understand. When I first started using Git and branches, I worried creating a PR for small change sets was a bit silly or overkill. Though over time, I've come to find such changes to be quite normal. Not only does it not bother me anymore, but in fact I tend to like small PRs like that because they're easy and quick to review.

In the past I've reviewed and approved a number of PRs which were just a single character change for a typo in a comment. I rather like these sorts of changes. If you have similar fixes, sure, batch them into a PR. Though if it's the only typo you see, why not just fix it. And if the change doesn't relate to anything else, why not make it it's own PR. If you batch it with something unrelated, and the other changes don't pass review, it holds up a fix that didn't need to be held up. I've also experienced that quite a few times now. And if the change set is a mess, breaking out unrelated stuff helps untangle balls of messy code and makes rework easier.

In this case though, I wasn't suggesting a whole new PR. Just a separate commit for that one change. Looking back now, I'd already forgotten that change had been batched into commit f9e3458f0fa448b5194beed6c4155efb4c42db3a "Hide truck management buttons if a facility isn't in a state where trucks should be manageable". From the commit message, and with the noise of the other changes in the diff for that commit, it's not super obvious that it contains a bug fix. It also not obvious how that one line change relates to the other lines changed in that commit. Having a separate commits provides clearer indication of intent. In a way, it helps communicate with any reviewers, or anyone investigating history later on who is trying to figure out why something is written a certain way.

ldicker83: To be honest, I'm not sure how that line got changed the way it did. I had it set to that, then it got unset somehow??? So I pushed it back. Keep in mind a lot of this I'm doing quickly and don't care to solve problems with git and its weird merging system. Ran into a lot of issues with it this week and by this point got sick of the constant stream of "Git didn't exit cleanly" messages. So eh. /me shrugs

PR #783, opened 2021-02-17
Fix stability issues

ldicker83:
Noticed that the game was suddenly crashy if a new game was started. There's a case in MineReport where a nullptr dereference can occur if there are no mines in the game yet.

I also discovered that the mine route table wasn't getting cleared when the MapViewState was destroyed so fixed that while I was at it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Introduces biowaste produced by residential units and some quality of life impro
 - Fix crash when right-clicking a Robot Command Center
 - Fix a mistake that tallied food totals before population consumption which resulted in incorrect value displayed in the HUD
 - Fix Mine full-screen UI showing "no routes" when loading a game
+- Fix for phantom mine routes being drawn on the Mini Map if you quit to the main menu from an active game and start a new game
 
 
 ## [0.8.0] - 2021-01-01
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Introduces biowaste produced by residential units and some quality of life impro
 - Fix crash when right-clicking a Robot Command Center
 - Fix a mistake that tallied food totals before population consumption which resulted in incorrect value displayed in the HUD
 - Fix Mine full-screen UI showing "no routes" when loading a game
+- Fix for phantom mine routes being drawn on the Mini Map if you quit to the main menu from an active game and start a new game
```
DanRStevens, line 26: I wonder if this is related to the debug assertion memory error you were seeing the other day when leaving a level.

ldicker83: Could be? Since I've fixed a few of these mistakes I'm not seeing the same weirdness. Same with the doubleClick handler -- dangling reference bug among other things.

At the same time I noticed that when I was getting the crashing and heap assertions Visual Studio was using over 6GB of RAM and was starting to page in and out of the swap file, behavior I've never seen before ever. After a restart I haven't seen a reproduction of that.

Still though, finding the little things I did by glancing over the code still makes me think that a proper code audit is in store. A lot of the development process here has been "Get it done, make it elegant later" so that functionality can be pushed forward into an actual game versus having perfect code. Many of my previous projects failed because I was trying for "the best code evar" instead of just making a game and worrying about making it pretty later.

PR #785, opened 2021-02-17
Fix insta death when landing colonists

### LINKED ISSUE #784: Impossible to land colonists without everybody dying
I noticed when testing with a new game (versus previously saved game) that no matter what options are set all colonists die whenever you land a colonist lander. Doesn't matter what structures you have in place to support them, they all just die.

Probably has to do with the way food storage is handled when the command center is disabled.

May need to reconsider the population requirement for the CC and have it be a structure that can operate without any colonists.

ldicker83:
Closes #784

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Introduces biowaste produced by residential units and some quality of life impro
 - Depleted food stores in the Command Center will be refilled each turn from the production of food producing facilities
 - Morale will take a take a hit whenever biowaste in residences overflows
 - Improve behavior of UI elements in full-screen Mine Operations UI
+- Command Center no longer requires any population to operate
 
 ### Fixed
 - Fix text overlap in Mine Operations Window
@@ -24,6 +25,7 @@ Introduces biowaste produced by residential units and some quality of life impro
 - Fix a mistake that tallied food totals before population consumption which resulted in incorrect value displayed in the HUD
 - Fix Mine full-screen UI showing "no routes" when loading a game
 - Fix for phantom mine routes being drawn on the Mini Map if you quit to the main menu from an active game and start a new game
+- Fix for immediate colony death when landing colonists
 
 
 ## [0.8.0] - 2021-01-01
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -24,6 +25,7 @@ Introduces biowaste produced by residential units and some quality of life impro
 - Fix a mistake that tallied food totals before population consumption which resulted in incorrect value displayed in the HUD
 - Fix Mine full-screen UI showing "no routes" when loading a game
 - Fix for phantom mine routes being drawn on the Mini Map if you quit to the main menu from an active game and start a new game
+- Fix for immediate colony death when landing colonists
```
DanRStevens, line 28: If this only happens in non-tagged releases, there's probably no real need to add a Change Log entry for it.

ldicker83: I'm pretty sure this is a problem that affected v0.8.0 - population requirement was made in that changeset.

I'm noticing that my testing of new versions of the game is inadequate as I assume that the beginning parts of the game are already done and tested and so I focus on running from save games. A lot of the fixes I implemented that aren't tied to specific issues are because I opted to try running the game from the very beginning. Something I'm going to have to change moving forward.

DanRStevens: Sounds like we could use some automated tests to help ensure quality.

Admittedly though, I'm not too sure how to go about writing tests for some of these cases. Most of our test code is unit tests, which are most appropriate for library code. Once you start involving the GUI, or a full executable, things get a little more complicated.

Though I suppose certain core game mechanics could be tested in a similar manner to library code, using unit tests.

ldicker83: That's where I'm stumbling too. How do you automate tests on an interactive program that displays output to the user visually?

Some of the tests could be automated by simulating specific parts of the code? And simulating input? All of the user input is handled through discrete functions that should be testable. Gonna have to be something to really think about since there are a lot of variables to consider here (and I don't just mean actual program variables)

PR #786, opened 2021-02-18
Improve planet select state

ldicker83:


OPHD/States/PlanetSelectState.cpp
```diff
@@ -23,6 +23,17 @@ std::size_t planetSelection;
 constexpr std::size_t planetSelectionInvalid = std::numeric_limits<std::size_t>::max();
 
 
+namespace
+{
+	std::vector<Planet::Attributes> PlanetAttributes =
+	{
```
DanRStevens: This is a little bit off-topic for this PR, but:
Technically you can get rid of the `=`. It would then be using brace initialization syntax `{}`.

----

We might be using capitalization inconsistently when it comes to constants. I think this was a discussion that came up about a year ago.

You can initialize local variables with runtime results from functions (non-`constexpr`), and declare them as `const`, and they typically won't be capitalized. You can also initialize them with `constexpr` compile-time known values and declare them as `const`. These are pretty similar to the previous case, so we might also use lowerCamelCase. We can also move such variables to a top level global position. But should we be using a different naming convention based on whether a variable is declared locally or globally?
```cpp
void someFunction()
{
    auto mutableValue = someOtherFunction1();
    const auto runtimeValue = someOtherFunction2(); // Not constexpr compatible
    const auto anotherConst = 5; // Or should this be "AnotherConst" (constexpr compatible)
}

const auto yetAnotherConst = 6; // Or should this be "YetAnotherConst" (same, constexpr compatible, but global)
```

I'm inclined to use a consistent naming convention regardless of scope.

I'm also maybe a little concerned that constant values could be mistaken for `class` names. Occasionally they may sound pretty similar.

On the other hand, the `m` prefix is scope based. Or should we consider a `g` prefix?

OPHD/States/PlanetSelectState.cpp
```diff
@@ -135,10 +129,10 @@ State* PlanetSelectState::update()
 	{
 		return this;
 	}
-	else if (planetSelection != planetSelectionInvalid)
+	else if (mPlanetSelection != constants::NO_SELECTION)
```
DanRStevens, line 132: I would caution against re-using `constants::NO_SELECTION` in many different contexts. Particularly when the objects being selected are of widely varying types.

Sentinel values are often context dependent. For instance, array based sentinel values typically use `-1`, while pointer based sentinel values typically use `0` or `nullptr`. If a data type changes between array based or pointer based, the sentinel value may also need to be updated. Keeping a local name for the sentinel values makes such updates easier, as you don't need to sort out which uses of the sentinel name apply to that particular collection type.

I have seen some rather unfortunate cases where some collections end up with effectively two sentinel values, depending on context, because this sort of things wasn't handled in a consistent manner, and it led to some pretty unusual bugs. In particular OP2, and the garage save/load crash bug.

OPHD/States/PlanetSelectState.cpp
```diff
@@ -115,9 +115,9 @@ State* PlanetSelectState::update()
 		mPlanets[i]->update();
 	}
 
-	renderer.drawText(mFontBold, "Mercury Type", mPlanets[0]->position() + NAS2D::Vector{64 - (mFontBold.width("Mercury Type") / 2), -mFontBold.height() - 10}, NAS2D::Color::White);
-	renderer.drawText(mFontBold, "Mars Type", mPlanets[1]->position() + NAS2D::Vector{64 - (mFontBold.width("Mars Type") / 2), -mFontBold.height() - 10}, NAS2D::Color::White);
-	renderer.drawText(mFontBold, "Ganymede Type", mPlanets[2]->position() + NAS2D::Vector{64 - (mFontBold.width("Ganymede Type") / 2), -mFontBold.height() - 10}, NAS2D::Color::White);
+	renderer.drawText(mFontBold, PlanetAttributes[0].name, mPlanets[0]->position() + NAS2D::Vector{64 - (mFontBold.width(PlanetAttributes[0].name) / 2), -mFontBold.height() - 10}, NAS2D::Color::White);
+	renderer.drawText(mFontBold, PlanetAttributes[1].name, mPlanets[1]->position() + NAS2D::Vector{64 - (mFontBold.width(PlanetAttributes[1].name) / 2), -mFontBold.height() - 10}, NAS2D::Color::White);
+	renderer.drawText(mFontBold, PlanetAttributes[2].name, mPlanets[2]->position() + NAS2D::Vector{64 - (mFontBold.width(PlanetAttributes[2].name) / 2), -mFontBold.height() - 10}, NAS2D::Color::White);
```
DanRStevens: With that change, the code becomes very symmetric, and could could further modify it to use a `for` loop.

Actually, if there was an easy way to match up entries between the `mPlanets` and `PlanetAttributes` collections, you could use a for-each loop.

ldicker83: Agreed. I didn't design PlanetAttributes, I think Brett did. I just wanted to clean it up a bit to remove string literals out of the main code and into a single source that can later be improved upon.

OPHD/States/PlanetSelectState.cpp
ldicker83: The naming convention I've used is this:

```cpp
mVarName; // Class member variable
functionName(); // member or stand-alone function
ClassName;
ConstantName;
EnumName;
localVariable;
```

I used to declare constants `IN_ALL_CAPS` but you've discouraged that in favor of `UpperCamelCase` and I'm finding that I prefer the latter vs the former so new code is using that approach.

I've noticed inconsistent capitalization in code that I haven't written myself and it's confused me a number of times because something like `notActuallyLocal` is declared as module or global constant when it should have been `NotActuallyLocal`. And so on.

Something I'll be addressing moving forward and being more strict about when accepting PR's

DanRStevens: I know this is nitpicking a bit, but you mention `ConstantName` and `localVariable`, but a variable can be both constant and local. :wink:

PR #787, opened 2021-02-18
Fix misc mistakes

ldicker83:


OPHD/UI/Reports/WarehouseReport.cpp
```diff
@@ -75,6 +75,7 @@ WarehouseReport::WarehouseReport() :
 WarehouseReport::~WarehouseReport()
 {
 	Control::resized().disconnect(this, &WarehouseReport::_resized);
+	Utility<EventHandler>::get().mouseDoubleClick().disconnect(this, &WarehouseReport::doubleClicked);
```
DanRStevens, line 78: I suppose this is why I've been suggesting a top-down event distribution mechanism. It makes problems like these impossible by design. If there's no connect/disconnect for each control to event sources, there's no chance to get it wrong and leave dangling pointers that can cause crashes or weird behaviour like controls responding to clicks when they aren't visible.

DanRStevens (approved):

OPHD/UI/Reports/WarehouseReport.cpp
ldicker83: I get that but I'm standing fast on not using a top-down approach because it's more complicated and harder to get right. If you're up for overhauling and rewriting the entire GUI I'm game. :D

PR #791, opened 2021-02-20
Allow robots and buildings to be placed at view edge

### LINKED ISSUE #788: Unable to place RoboDozer in 3 out of 4 visible map corners
Test situation:
After starting a new game and placing the lander, advanced a turn and I'm trying to place a RoboDozer at each of the four visible corners of the visible map per turn:
cant_bulldoze

In the current setup, the only corner I could place it was at the top.

Perhaps this is a design decision or something the original also did or an unwritten law of isometric design, if so, the issue can be summarily closed.

The other three corners did not allow me to. It seems not possible to place a RoboDozer at the SE and SW edges (also marked in red). I also resized my window to get both an odd and even edge length, in both cases the above situation was true.

I guess this has to do with how the game decides which tile is currently visible. Even though it is visible to me as a human, the game code tells me it is in fact not visible to the game. The top corner that did work maps to the origin of the Rectangle, all other 3 corners are literal "edge" cases, where one or both coordinates are just outside of the calculated rectangle, even though they are (humanly) visible tiles. 

```
bool TileMap::isVisibleTile(NAS2D::Point<int> position, int z) const
{
	if (!NAS2D::Rectangle{mMapViewLocation.x, mMapViewLocation.y, mEdgeLength - 1, mEdgeLength - 1}.contains(position))
	{
		return false;
	}

	if (z != mCurrentDepth)
	{
		return false;
	}

	return true;
}
```

So would that mean the mEdgeLength should not be decreased by 1 in both places when checking a position with contains ?

DanRStevens:
Closes #788

Fix off by 1 error. Previously the lower left and lower right edges were not considered viewable, even though they were being visibly drawn. This prevented items from being placed on those tiles.

### DIFF
--- a/OPHD/Map/TileMap.cpp
+++ b/OPHD/Map/TileMap.cpp
@@ -554,7 +554,7 @@ Tile* TileMap::getVisibleTile(NAS2D::Point<int> position, int level)
 
 bool TileMap::isVisibleTile(NAS2D::Point<int> position, int z) const
 {
-	if (!NAS2D::Rectangle{mMapViewLocation.x, mMapViewLocation.y, mEdgeLength - 1, mEdgeLength - 1}.contains(position))
+	if (!NAS2D::Rectangle{mMapViewLocation.x, mMapViewLocation.y, mEdgeLength, mEdgeLength}.contains(position))
 	{
 		return false;
 	}


Brett208:
Thanks for working on this, I had noticed this issue in the past and not reported it. Sort of annoying to not be able to select edge tiles although it doesn't really break anything.

ldicker83:
Been meaning to look into this. Thanks!

PR #793, opened 2021-02-21
Robot inspector window

### LINKED ISSUE #55: Add a "Cancel Orders" option for deployed robots
As discussed in this forum thread: https://forum.outpost2.net/index.php/topic,6085.msg85021.html#msg85021

There should be a way to cancel a robots orders once deployed in the field. Havkyp suggested a right-click options -- I agree.

So basically this task will be adding a Cancel Orders button in a small panel that is displayed when the user right-clicks on a tile with a deployed robot in it.

Canceling a robot's orders will have the effect of recalling it without having had any effect on the terrain. Bulldozing will need to be tweaked so that if a dozer is recalled the terrain is reverted to its original type.

### LINKED ISSUE #72: Implement ability to self-destruct robots
In my play-throughs of the game thus far, the needs to self-destruct robots has become apparent. In three scenarios I'd built an additional robo-dozer before building an additional robot command center which eventually lead to the inability to deploy a newly built robo-minder leading to colony collapse because of the length of time it takes to extend the mine shaft.

In these cases, being able to self-destruct a robot would allow the player to deploy additional new robots instead of facing inevitable failure.

The case could be made that "Well why not just switch out the robots?". This is fair but ultimately the player should be punished for making such a mistake.

### LINKED ISSUE #136: Add a Robot Inspector window
Related to #55 and #72 , this window should be brought up when the user right-clicks a tile with a Robot in it.

image

ldicker83:
Closes #55 
Closes #72 
Closes #136

### DIFF
--- a/OPHD/Things/Robots/Robot.cpp
+++ b/OPHD/Things/Robots/Robot.cpp
@@ -26,6 +26,12 @@ void Robot::updateTask()
 		die();
 	}
 
+	if (mCancelTask)
+	{
+		mTurnsToCompleteTask = 0;
+		return;
+	}
+
 	mTurnsToCompleteTask--;
 
 	if (mTurnsToCompleteTask == 0)
--- a/OPHD/Things/Robots/Robot.cpp
+++ b/OPHD/Things/Robots/Robot.cpp
@@ -23,7 +23,6 @@ void Robot::updateTask()
 {
 	if (mSelfDestruct)
 	{
-		mSelfDestructCallback();
 		die();
 	}
 
--- a/OPHD/Things/Robots/Robot.cpp
+++ b/OPHD/Things/Robots/Robot.cpp
@@ -1,12 +1,14 @@
 #include "Robot.h"
 
-Robot::Robot(const std::string& name, const std::string& sprite_path) :
-	Thing(name, sprite_path, "running")
+Robot::Robot(const std::string& name, const std::string& sprite_path, Type t) :
+	Thing(name, sprite_path, "running"),
+	mType{ t }
 {}
 
 
-Robot::Robot(const std::string& name, const std::string& sprite_path, const std::string& initialAction) :
-	Thing(name, sprite_path, initialAction)
+Robot::Robot(const std::string& name, const std::string& sprite_path, const std::string& initialAction, Type t) :
+	Thing(name, sprite_path, initialAction),
+	mType{ t }
 {}
 
 
--- a/OPHD/UI/RobotInspector.cpp
+++ b/OPHD/UI/RobotInspector.cpp
@@ -94,13 +94,14 @@ void RobotInspector::btnCancelOrdersClicked()
 
 void RobotInspector::btnSelfDestructClicked()
 {
-
+	mRobot->seldDestruct(true);
+	hide();
 }
 
 
 void RobotInspector::btnCancelClicked()
 {
-
+	hide();
 }
 
 
--- a/OPHD/UI/RobotInspector.cpp
+++ b/OPHD/UI/RobotInspector.cpp
@@ -20,9 +20,9 @@ namespace
 	{
 		static const std::map<RobotType, const Image*> robotImages
 		{
-			{RobotType::ROBOT_DIGGER, &imageCache.load("ui/interface/product_robodigger.png")},
-			{RobotType::ROBOT_DOZER, &imageCache.load("ui/interface/product_robodozer.png")},
-			{RobotType::ROBOT_MINER, &imageCache.load("ui/interface/product_robominer.png")}
+			{RobotType::Digger, &imageCache.load("ui/interface/product_robodigger.png")},
+			{RobotType::Dozer, &imageCache.load("ui/interface/product_robodozer.png")},
+			{RobotType::Miner, &imageCache.load("ui/interface/product_robominer.png")}
 		};
 		return *robotImages.at(robotType);
 	}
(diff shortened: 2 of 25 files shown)


OPHD/Things/Robots/Robot.cpp
```diff
@@ -1,12 +1,14 @@
 #include "Robot.h"
 
-Robot::Robot(const std::string& name, const std::string& sprite_path) :
-	Thing(name, sprite_path, "running")
+Robot::Robot(const std::string& name, const std::string& sprite_path, Type t) :
+	Thing(name, sprite_path, "running"),
+	mType{ t }
```
DanRStevens, line 5: We should really be avoiding one letter variable names like `t`, particularly for names/types that are not universal (like `x`, `y`, or `i`). It's not like source code storage size is an issue in this day and age. And thinking speed is far more limiting than typing speed when it comes to programming.

OPHD/Things/Robots/Robot.cpp
```diff
@@ -26,6 +26,12 @@ void Robot::updateTask()
 		die();
 	}
 
+	if (mCancelTask)
+	{
+		mTurnsToCompleteTask = 0;
+		return;
+	}
```
DanRStevens, line 33: Does `mCancelTask` need to be reset after being used?

Edit: Looks like a `reset()` method was added later. Though a `reset()` method would need to be manually called. Maybe it can auto reset once the `mCancelTask` flag is used.

OPHD/UI/RobotInspector.cpp
```diff
@@ -0,0 +1,91 @@
...
+RobotInspector::RobotInspector() :
+	mFont{ fontCache.load(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL) },
+	mFontBold{ fontCache.load(constants::FONT_PRIMARY_BOLD, constants::FONT_PRIMARY_NORMAL) },
+	btnCancelOrders{ "Cancel Orders" },
+	btnSelfDestruct{ "Self Destruct" },
+	btnIdle{ "Set Idle" },
+	btnCancel{ constants::ButtonCancel }
+{
+	init();
+}
```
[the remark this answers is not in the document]
ldicker83: I am of the philosophy that the constructor should never have any code and if it does it should be relegated to a dedicated function that's called from the c'tor.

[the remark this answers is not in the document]
DanRStevens: Yes, but why?

OPHD/UI/RobotInspector.cpp
```diff
@@ -0,0 +1,91 @@
...
+	btnIdle.size({ buttonWidth, buttonHeight });
+
+	add(btnCancel, { 0, 0 });
+	btnCancel.size({ mFont.width(constants::ButtonCancel), buttonHeight });
+}
+
+
+void RobotInspector::focusOnRobot(Robot* robot)
+{
+	if (!robot) { throw std::runtime_error("RobotInspector::focusOnRobot(): nullptr passed "); }
```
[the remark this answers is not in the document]
ldicker83: Huh, fair point. Didn't think about that. I'm still thinking in the low-level and am finding that I tend to revert to c-style design. Ultimately there should never be a case off a nullptr being passed.

OPHD/Things/Robots/Robot.cpp
ldicker83: 't' kind of is but generally speaking I agree. Using 'type' works but it's also a keyword and I would rather avoid '_type'.

OPHD/Things/Robots/Robot.cpp
ldicker83: I was thinking there had to be a better way to do it and you just nailed it I think. I'll make that adjustment in a separate change set.

OPHD/UI/RobotInspector.cpp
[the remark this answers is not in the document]
ldicker83: If I end up needing to create any overloaded c'tors, I can invoke the function instead of copy/pasting. That's the biggest reason.

OPHD/Things/Robots/Robot.cpp
DanRStevens: I'm guessing if the robodozer graphic was simply drawn on top of the tile, mountains might still peek through onto surrounding tiles.

Hmm, actually, it might be interesting to have a set of doze graphics with the mountain gradually wearing down. Would also provide a visual indication of how much doze time was left. That would require graphics work though.

----

Anyway, sounds like there are multiple tasks to cleanup here, and they're not all done at the same time. Maybe code can be refactored so the effects of cancelling can all be done together.

OPHD/UI/RobotInspector.cpp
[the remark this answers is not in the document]
DanRStevens: Yes, that's why I gave a link to **delegating constructors**. That's a problem that was dealt with in C++11.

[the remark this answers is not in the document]
ldicker83: I managed to miss that link.

I guess you could say I'm still getting used to the new features of C++ since C++11. You'd think that a decade of work would give me a bit of a clue -_-

OPHD/Things/Robots/Robot.cpp
ldicker83: The robot graphic effectively is just drawn on top of the terrain tile.

There's the possibility of having robots or structures or even other 'things' that can contain a property that includes a tile to be drawn instead of the tile index. Something like `Image& Thing::terrainTile()` which can then be drawn instead of using the tileset and tile index.

Could allow for interesting effects but something that I want to have a solid grasp on before implementing.

OPHD/UI/RobotInspector.cpp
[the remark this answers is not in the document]
DanRStevens: C++11 was a bit update :wink:

OPHD/Things/Robots/Robot.cpp
ldicker83: Perhaps I'm confusing it with `typedef` and `typeid`. Did a check for reserved keywords and `type` isn't part of that. I guess I have no excuse to fully expand it now. :D

DanRStevens: I was actually also questioning if `type` was a reserved keyword when I first looked at the code. That's probably why I was a bit vague and didn't suggest renaming it to `type`. :wink:

Only noticed just as I was closing the tab that there was a method named `type()`.

Issue #115, opened 2019-11-25
Implement Connectedness Overlay


ldicker83:
Implement connectedness overlay using the changes introduced in #800

ldicker83:
Repurposing this task to use the update 'overlay' concept introduced in #800

PR #800, opened 2021-02-22
Add tile overlay color

### LINKED ISSUE #114: Find a cleaner way to change tile render color
I noted it in TileMap.cpp, I need to find a much better way to handle tile render color versus the clumsy nested if/else chain I've got going on. I'm at a point now where I want to be able to highlight tiles in different colors based on whether or not they're part of paths, comm ranges, etc.

A simple way to do this would be to add a `NAS2D::Color_4ub` member to the `Tile` class and get/set functions for it.

See #xxx about addressing potential issues with this method of tile coloration.

ldicker83:
Closes #114

Also mostly addresses #115 <-- this will be closed when the MapViewState maintains lists of tile overlays and uses the new Tile::overlay() functions to switch between overlay modes.

### DIFF
--- a/OPHD/Map/Tile.cpp
+++ b/OPHD/Map/Tile.cpp
@@ -159,9 +159,3 @@ float Tile::distanceTo(NAS2D::Point<int> point)
 	const auto direction = point - position();
 	return static_cast<float>(std::sqrt((direction.x * direction.x) + (direction.y * direction.y)));
 }
-
-
-void Tile::overlay(Overlay overlay)
-{
-	mOverlay = overlay;
-}
--- a/OPHD/Map/Tile.cpp
+++ b/OPHD/Map/Tile.cpp
@@ -7,6 +7,33 @@
 #include <cmath>
 
 
+std::map<Tile::Overlay, NAS2D::Color> OverlayColorTable =
+{
+	{ Tile::Overlay::None, NAS2D::Color::Normal },
+	{ Tile::Overlay::Communications, { 125, 200, 255 } },
+	{ Tile::Overlay::Connectedness, NAS2D::Color::Green }
+};
+
+std::map<Tile::Overlay, NAS2D::Color> OverlayHighlightColorTable =
+{
+	{ Tile::Overlay::None, NAS2D::Color{ 125, 200, 255 } },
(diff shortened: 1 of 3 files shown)


OPHD/Map/Tile.cpp
```diff
@@ -7,6 +7,44 @@
...
+
+const NAS2D::Color& overlayColor(Tile::Overlay overlay, bool isHighlighted)
+{
+	if (isHighlighted)
+	{
+		return overlayHighlightColor(overlay);
+	}
+		
+	return overlayColor(overlay);
+}
```
DanRStevens, line 33: Could probably have still used a ternary here for a one line function body.

OPHD/Map/TileMap.cpp
```diff
@@ -311,12 +311,8 @@ void TileMap::draw()
...
 				const auto subImageRect = NAS2D::Rectangle{static_cast<int>(tile.index()) * TILE_WIDTH, tsetOffset, TILE_WIDTH, TILE_HEIGHT};
 				const bool isTileHighlighted = NAS2D::Vector{col, row} == highlightOffset;
-				const bool isConnectionHighlighted = mShowConnections && tile.connected();
-				const NAS2D::Color highlightColor =
-					isTileHighlighted ?
-						isConnectionHighlighted ? NAS2D::Color{71, 224, 146} : NAS2D::Color{125, 200, 255} :
-						isConnectionHighlighted ? NAS2D::Color::Green : NAS2D::Color::Normal;
-				renderer.drawSubImage(mTileset, position, subImageRect, highlightColor);
+
+				renderer.drawSubImage(mTileset, position, subImageRect, overlayColor(tile.overlay(), isTileHighlighted));
```
DanRStevens, line 315: Looks like `mShowConnections` is no longer being used. Was wondering if that was intentional.

DanRStevens (commented):
I'm puzzling a little over why `mOverlay` is a member field of `Tile`. I assume you'd draw all tiles on the map with the same overlay setting? Maybe it should be stored on the map object, and passed in to the `Tiles` objects when drawing.

OPHD/Map/TileMap.cpp
ldicker83: Yes, because connectedness will be part of an overlay

PR #805, opened 2021-02-23
Allow savegames to be deleted

### LINKED ISSUE #802: Add option to delete a save game
To delete you'll have to manually remove it by browsing to your user directory and removing the savegame file from the savegames directory. In the future this is a function that could probably be easily added to the FileIO window. It never occurred to me that a user may want to remove their game so that would be a great quality of life improvement to the UI. I'll add that as a task.

_Originally posted by @ldicker83 in https://github.com/OutpostUniverse/OPHD/issues/799#issuecomment-783862580_

This should be a fairly simple change to the FileIO Window.

belgianguy:
closes #802 
Please review and let me know if I missed anything.

The File dialog does not gets hidden after deleting an item, as you might want to delete more items and not reopen the menu each time.

The delete button is always available when the save or load dialog is shown, but the action SAVE or LOAD is set before the dialog is shown, that's why I introduced a new button for the delete action.

File operation FILE_DELETE was added.

There is a check if the file exists before deleting it.

The delete button is disabled when nothing is selected, but is available when the filename is being changed.

The Listbox scanDirectory method is called after each deletion so the directory listing is updated accordingly.

I placed the Delete button on the left side so it's not clicked by accident when the user wants to click Load or Cancel.

I used the same look and feel as the Load button, not sure if that's the best direction, although they cannot be confused.

Perhaps we need to add a prompt after a user clicks Delete to ask the user if they are sure they want to delete <file> and in case of Yes, continue, and in case of No, just return. But that might be considered scope creep.

### DIFF
--- a/OPHD/States/MainMenuState.cpp
+++ b/OPHD/States/MainMenuState.cpp
@@ -158,36 +158,33 @@ void MainMenuState::fileIoAction(const std::string& filePath, FileIo::FileOperat
 		return;
 	}
 
-	if (fileOp == FileIo::FileOperation::FILE_LOAD)
+	std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";
+
+	if (!Utility<Filesystem>::get().exists(filename))
 	{
-		std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";
+		doNonFatalErrorMessage("Load Failed", "File '" + filename + "' was not found.");
+		return;
+	}
 
-		if (!Utility<Filesystem>::get().exists(filename))
-		{
-			doNonFatalErrorMessage("Load Failed", "File '" + filename + "' was not found.");
-			return;
-		}
+	try
+	{
+		checkSavegameVersion(filename);
 
-		try
-		{
-			checkSavegameVersion(filename);
+		GameState* gameState = new GameState();
+		MapViewState* mapview = new MapViewState(gameState->getMainReportsState(), filename);
+		mapview->_initialize();
+		mapview->activate();
 
-			GameState* gameState = new GameState();
-			MapViewState* mapview = new MapViewState(gameState->getMainReportsState(), filename);
-			mapview->_initialize();
-			mapview->activate();
+		gameState->mapviewstate(mapview);
+		mReturnState = gameState;
 
-			gameState->mapviewstate(mapview);
-			mReturnState = gameState;
-
-			Utility<Renderer>::get().fadeOut(constants::FADE_SPEED);
-			Utility<Mixer>::get().fadeOutMusic(constants::FADE_SPEED);
-		}
-		catch (const std::exception& e)
-		{
-			mReturnState = this;
-			doNonFatalErrorMessage("Load Failed", e.what());
-		}
+		Utility<Renderer>::get().fadeOut(constants::FADE_SPEED);
+		Utility<Mixer>::get().fadeOutMusic(constants::FADE_SPEED);
+	}
+	catch (const std::exception& e)
+	{
+		mReturnState = this;
+		doNonFatalErrorMessage("Load Failed", e.what());
 	}
 }
 
--- a/OPHD/States/MainMenuState.cpp
+++ b/OPHD/States/MainMenuState.cpp
@@ -189,21 +189,6 @@ void MainMenuState::fileIoAction(const std::string& filePath, FileIo::FileOperat
 			doNonFatalErrorMessage("Load Failed", e.what());
 		}
 	}
-
-	if (fileOp == FileIo::FileOperation::FILE_DELETE)
-	{
-		std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";
-
-		try
-		{
-			Utility<Filesystem>::get().del(filename);
-		}
-		catch(const std::exception& e)
-		{
-			doNonFatalErrorMessage("Delete Failed", e.what());
-		}
-	}
-
 }
 
 
--- a/OPHD/States/MainMenuState.cpp
+++ b/OPHD/States/MainMenuState.cpp
@@ -194,14 +194,13 @@ void MainMenuState::fileIoAction(const std::string& filePath, FileIo::FileOperat
 	{
 		std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";
 
-		if (!Utility<Filesystem>::get().exists(filename))
-		{
-			doNonFatalErrorMessage("Delete Failed", "File '" + filename + "' was not found.");
-			return;
-		}
-		else
+		try
 		{
 			Utility<Filesystem>::get().del(filename);
+		}
+		catch(const std::exception& e)
+		{
+			doNonFatalErrorMessage("Delete Failed", e.what());
 		}
 	}
 
--- a/OPHD/States/MainMenuState.cpp
+++ b/OPHD/States/MainMenuState.cpp
@@ -148,37 +148,63 @@ void MainMenuState::enableButtons()
  */
 void MainMenuState::fileIoAction(const std::string& filePath, FileIo::FileOperation fileOp)
 {
-	if (fileOp != FileIo::FileOperation::FILE_LOAD) { return; }
-	if (filePath.empty()) { return; }
-
-	std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";
-
-	if (!Utility<Filesystem>::get().exists(filename))
+	if (fileOp == FileIo::FileOperation::FILE_SAVE)
 	{
-		doNonFatalErrorMessage("Load Failed", "File '" + filename + "' was not found.");
 		return;
 	}
 
-	try
+	if (filePath.empty())
 	{
-		checkSavegameVersion(filename);
-
-		GameState* gameState = new GameState();
-		MapViewState* mapview = new MapViewState(gameState->getMainReportsState(), filename);
-		mapview->_initialize();
-		mapview->activate();
-
-		gameState->mapviewstate(mapview);
-		mReturnState = gameState;
-
-		Utility<Renderer>::get().fadeOut(constants::FADE_SPEED);
-		Utility<Mixer>::get().fadeOutMusic(constants::FADE_SPEED);
+		return;
 	}
-	catch (const std::exception& e)
+
+	if (fileOp == FileIo::FileOperation::FILE_LOAD)
 	{
-		mReturnState = this;
-		doNonFatalErrorMessage("Load Failed", e.what());
+		std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";
+
+		if (!Utility<Filesystem>::get().exists(filename))
+		{
+			doNonFatalErrorMessage("Load Failed", "File '" + filename + "' was not found.");
+			return;
+		}
+
+		try
+		{
+			checkSavegameVersion(filename);
+
+			GameState* gameState = new GameState();
+			MapViewState* mapview = new MapViewState(gameState->getMainReportsState(), filename);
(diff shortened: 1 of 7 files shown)


OPHD/States/MapViewStateIO.cpp
```diff
@@ -79,6 +79,17 @@ static void readRccRobots(NAS2D::Xml::XmlAttribute* attr, Structure& structure,
...
+void MapViewState::del(const std::string& filePath) {
+	if (Utility<Filesystem>::get().exists(filePath))
+	{
+		Utility<Filesystem>::get().del(filePath);
+	}
+	else
+	{
+		throw std::runtime_error("File '" + filePath + "' was not found.");
+	}
+}
```
DanRStevens, line 91: I'm wondering if this functionality should be part of the `FileIo` dialog, rather than `MapViewState`.

----

This function contains a time-of-check to time-of-use race condition. It's possible the file can be deleted by another process between the calls to `exists` and `del`.

Since `Filesystem::del` will already `throw` if the file doesn't exist, it may be better to just call `Filesystem::del`, and let the exception handling there raise an exception when the file doesn't exist.

OPHD/States/MainMenuState.cpp
```diff
@@ -148,37 +148,63 @@ void MainMenuState::enableButtons()
...
+
+		if (!Utility<Filesystem>::get().exists(filename))
+		{
+			doNonFatalErrorMessage("Delete Failed", "File '" + filename + "' was not found.");
+			return;
+		}
+		else
+		{
+			Utility<Filesystem>::get().del(filename);
+		}
```
DanRStevens, line 205: Looks like the delete code is duplicated. I think this is a further sign that perhaps the `FileIo` dialog should handle this responsibility.

DanRStevens (commented):
This looks like a reasonable attempt at implementing the feature.

I have some concern this functionality might be better handled by the `FileIo` dialog itself. That could reduce duplication for the delete code, and eliminate the need for the new enum value for a new operation type.

Curious what Leeor thinks of this.

OPHD/States/MainMenuState.cpp
ldicker83: I based the design of this on the Windows API -- the FileIO dialog's are a standard function but the dialog itself doesn't do anything, it returns to the caller the action that the user wants to do. It's then the responsibility of the caller to decide how to handle it.

I think you're right when you state above that the Filesystem will throw if there's a problem so we should just let the Filesystem exception code do the work and use a try/catch block. So the above highlighted code block could look like this:

```cpp
if (fileOp == FileIo::FileOperation::FILE_DELETE)
{
	std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";

	try
	{
		Utility<Filesystem>::get().del(filename);
	}
	catch(const std::exception& e)
	{
		doNonFatalErrorMessage("Delete Failed", e.what());
	}
}
```

I assume this was based off the previously existing code which is a little on the old side and was developed long before the current state of the Filesystem code so I think it makes sense to start the improvement with a new block of code.

Note that the try/catch block can be omitted thought I opt to catch here because these are exceptions that can almost universally be recovered from in this context (file operations).

belgianguy:
I can't follow, I had no idea about the historical context, nor about the race condition, why is it bad to have another file operation? Load and save code was duplicated in the same manner, forcing my current approach. 

Propose to close and rethink and to add more specs to the ticket. pointers about expectations, historical context, race-conditions, ...

OPHD/States/MapViewStateIO.cpp
DanRStevens: To add a bit of clarity, the method could be simplified to:
```cpp
void MapViewState::del(const std::string& filePath) {
	Utility<Filesystem>::get().del(filePath);
}
```

Which of course means we don't really need the method, and could just call: `Utility<Filesystem>::get().del(filePath);`

Internally the `Filesystem::del` method tries to delete the file, and if it didn't exist, it will `throw` and exception. Thus there is little need to detect this situation ahead of time, and write a custom `throw`.

```cpp
void Filesystem::del(const std::string& filename) const
{
	if (PHYSFS_delete(filename.c_str()) == 0)
	{
		throw std::runtime_error(std::string("Unable to delete '") + filename + "':" + PHYSFS_getErrorByCode(PHYSFS_getLastErrorCode()));
	}
}
```

The race condition is avoided because the `PHYSFS_delete` method combines the delete and existence check into a single atomic operation. If the file exists, it's deleted. If it doesn't exist, the return value indicates the file didn't exist, and that error return value gets translated into an exception.

DanRStevens:
I don't think there's any need to close, nor do any significant rework.

I think maybe simplify the delete code a bit, and let the internal exception mechanism handle the error case. That should be good enough.

I proposed a solution to get rid of the race condition. At any rate, it was never much to worry about. It's a case of either getting an exception from one code path, or getting a slightly different though basically equivalent exception from another code path. It's more potentially confusing than any sort of real problem.

I looked briefly at the code, though didn't see any removed code related to file deletion, so I don't know there is much historical context here. Plus, as stated, there was no delete functionality previously.

I haven't looked too closely at error dialogs to show exception messages. Leeor brought that up, so I'd want to take another quick look over the code to see how error messages are presented to the user. We probably want to go with Leeor's suggestion of catching any exceptions and translating them into an error dialog.

Actually, Leeor's suggestion basically includes my simplification taken all the way, and then including error handling, so perhaps focus more on Leeor's suggestion:
```cpp
if (fileOp == FileIo::FileOperation::FILE_DELETE)
{
	std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";

	try
	{
		Utility<Filesystem>::get().del(filename);
	}
	catch(const std::exception& e)
	{
		doNonFatalErrorMessage("Delete Failed", e.what());
	}
}
```

belgianguy:
Ok, I'll give it a try :) It's all very new to me, but I want to learn as I go, so all advice is well taken!

belgianguy:
Okay, added the suggested changes, I also removed the return statement in the catch of the deletion exception in MapViewState::fileIoAction, as in the case of deletion faillure, there is nothing that needs to be prevented, the file dialog doesn't get hidden on succesful deletion. 

In the case of failing to load a game that return statement in the catch clause is desired as it will prevent the file dialog from being hidden.

DanRStevens:
I noticed you're not a member of the Outpost Universe organization on GitHub. I sent you an invite. Might make it easier to collaborate. You won't need to fork a separate copy of the repository to create PRs. Instead you can create a branch right in the original repository. It would also mean you'd be listed as a potential reviewer, so if we fix an issue you reported, or some code that you wrote, it would be easier to coordinate review.

ldicker83 (approved):
This looks good to me.

Before merging this with master, I didn't check out a copy and test myself, does the FileIO window refresh its list after deletion?

I'm also thinking that Dan may have a point, specifically in the context of the delete function. It may make sense to have the delete function occur in the FileIO window itself. So that code that's gone into MapViewStateUi may work better in the click handler for the delete button in FileIO. So basically, FileIO::btnFileDeleteClicked would look like this:

```cpp
void FileIo::btnFileDeleteClicked()
{
	// Remove this line
	//mCallback(txtFileName.text(), FILE_DELETE);

	// Replace with this
	std::string filename = constants::SAVE_GAME_PATH + txtFileName.text()+ ".xml";

	try
	{
		Utility<Filesystem>::get().del(filename);
	}
	catch(const std::exception& e)
	{
		doNonFatalErrorMessage("Delete Failed", e.what());
	}

	txtFileName.text("");
	txtFileName.resetCursorPosition();
	btnFileDelete.enabled(false);
	scanDirectory(constants::SAVE_GAME_PATH);
}
```

Then the MapViewStateUi.cpp code can have the file delete operation removed from it since it'll be handled by the FileIO window itself.

Make sense?

DanRStevens (commented):
Leeor's reply may supersede some of my advice here.

Some of my comments are on existing code design, so not really relevant to this PR. Just noting it here because the diff highlighted some of those lines, and the design concerns relate to issues discussed earlier in this PR.

----

I would like to test these changes locally. I suppose that means I need to add a git remote for your fork of the repository. It's been a while, so I need to lookup how to do that again.

OPHD/States/MainMenuState.cpp
```diff
@@ -148,37 +148,62 @@ void MainMenuState::enableButtons()
...
-		Utility<Mixer>::get().fadeOutMusic(constants::FADE_SPEED);
+	if (fileOp == FileIo::FileOperation::FILE_LOAD)
+	{
+		std::string filename = constants::SAVE_GAME_PATH + filePath + ".xml";
+
+		if (!Utility<Filesystem>::get().exists(filename))
+		{
+			doNonFatalErrorMessage("Load Failed", "File '" + filename + "' was not found.");
+			return;
+		}
```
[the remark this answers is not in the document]
ldicker83: Agreed.

belgianguy:
@ldicker83 
> Before merging this with master, I didn't check out a copy and test myself, does the FileIO window refresh its list after deletion?
Yes, the scandirectory call at the end refreshes the contents of the listbox after a deletion happened.

Please review my latest changes. As there have been quite some reactions, I'm a bit hesitant to make large changes, so for now it'll be just small steps. I hope we're getting closer! :)

OPHD/States/MapViewStateUi.cpp
```diff
@@ -565,13 +565,13 @@ void MapViewState::fileIoAction(const std::string& filePath, FileIo::FileOperati
...
 	}
-	else
+	else if (fileOp == FileIo::FileOperation::FILE_SAVE)
 	{
 		save(constants::SAVE_GAME_PATH + filePath + ".xml");
+		mFileIoDialog.hide();
 	}
-
-	mFileIoDialog.hide();
 }
```
DanRStevens, line 575: Looks like the changes here are not strictly needed after backing out of the new return type.

OPHD/States/MainMenuState.cpp
```diff
@@ -148,36 +148,46 @@ void MainMenuState::enableButtons()
...
-		mReturnState = gameState;
-
-		Utility<Renderer>::get().fadeOut(constants::FADE_SPEED);
-		Utility<Mixer>::get().fadeOutMusic(constants::FADE_SPEED);
+		return;
 	}
-	catch (const std::exception& e)
+
+	if (fileOp == FileIo::FileOperation::FILE_LOAD)
 	{
```
DanRStevens, line 162: Also this, the extra guard added here isn't strictly needed anymore.

DanRStevens (approved):
Looks good. I'd be happy merging this.

I did notice a few changes added in this PR to handle the new `FILE_DELETE` return type, which has since been removed. Not sure if we want to go back and remove those changes or leave as is. The changes may add a bit of code bulk that doesn't need to be there, though perhaps someone feels it adds some clarity? At any rate, it wouldn't be a problem to leave them in.

OPHD/States/MapViewStateUi.cpp
```diff
@@ -569,6 +569,7 @@ void MapViewState::fileIoAction(const std::string& filePath, FileIo::FileOperati
 	else
 	{
 		save(constants::SAVE_GAME_PATH + filePath + ".xml");
+
 	}
```
DanRStevens, line 573: Superfluous blank line.

DanRStevens (approved):
Wow, that really cleans up the changes and makes it very easy to read the overall picture.

DanRStevens:
Ok, looks like all the talking points have been resolved, other than the one unreleated to the PR that we were going to look into as a separate PR:
https://github.com/OutpostUniverse/OPHD/pull/805#discussion_r582450711

I'm going to go ahead and merge now.

PR #807, opened 2021-02-25
Implement comm overlay

### LINKED ISSUE #803: Implement Communications Range Overlay
Implement tile coloring using Overlay changes introduced in #800

ldicker83:
Closes #803

### DIFF
--- a/OPHD/Map/Tile.h
+++ b/OPHD/Map/Tile.h
@@ -25,11 +25,11 @@ public:
 
 public:
 	Tile() = default;
-	Tile(NAS2D::Point<int> position, int depth, TerrainType index);
-	Tile(const Tile& other) = delete;
-	Tile& operator=(const Tile& other) = delete;
-	Tile(Tile&& other) noexcept;
-	Tile& operator=(Tile&& other) noexcept;
+	Tile(NAS2D::Point<int>, int, TerrainType);
+	Tile(const Tile&) = delete;
+	Tile& operator=(const Tile&) = delete;
+	Tile(Tile&&) noexcept;
+	Tile& operator=(Tile&&) noexcept;
 	~Tile();
 
 	TerrainType index() const { return mIndex; }
@@ -60,7 +60,7 @@ public:
 	bool thingIsStructure() const { return structure() != nullptr; }
 	bool thingIsRobot() const { return robot() != nullptr; }
 
-	void pushThing(Thing* thing);
+	void pushThing(Thing*);
 	void deleteThing();
 
 	void removeThing();
@@ -68,8 +68,8 @@ public:
 	Mine* mine() { return mMine; }
 	void pushMine(Mine*);
 
-	float distanceTo(Tile* tile);
-	float distanceTo(NAS2D::Point<int> point);
+	float distanceTo(Tile*);
+	float distanceTo(NAS2D::Point<int>);
 
 	void overlay(Overlay overlay) { mOverlay = overlay; }
 	Overlay overlay() const { return mOverlay; }
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -1376,6 +1376,7 @@ void MapViewState::checkCommRangeOverlay()
 	
 	for (auto cc : command)
 	{
+		if (!cc->operational()) { continue; }
 		auto& centerTile = structureManager.tileFromStructure(cc);
 		auto commAreaRect = buildAreaRectFromTile(centerTile, constants::ROBOT_COM_RANGE);
 		fillCommList(mCommRangeOverlay, *mTileMap, centerTile, commAreaRect, constants::ROBOT_COM_RANGE);
@@ -1383,6 +1384,7 @@ void MapViewState::checkCommRangeOverlay()
 
 	for (auto tower : commTowers)
 	{
+		if (!tower->operational()) { continue; }
 		auto& centerTile = structureManager.tileFromStructure(tower);
 		auto commAreaRect = buildAreaRectFromTile(centerTile, constants::COMM_TOWER_BASE_RANGE);
 		fillCommList(mCommRangeOverlay, *mTileMap, centerTile, commAreaRect, constants::COMM_TOWER_BASE_RANGE);
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -902,6 +902,11 @@ void MapViewState::placeRobot()
 				else { return; }
 			}
 
+			if (structure->structureClass() == Structure::StructureClass::Communication)
+			{
+				checkCommRangeOverlay();
+			}
+
 			auto recycledResources = StructureCatalogue::recyclingValue(structure->structureId());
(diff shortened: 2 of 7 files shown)


OPHD/States/MapViewStateUi.cpp
```diff
@@ -24,6 +24,16 @@
...
 
 
+static void setOverlay(Button& button, TileList& tileList, Tile::Overlay overlay)
+{
+	auto overlayToUse = button.toggled() ? overlay : Tile::Overlay::None;
+	for (auto tile : tileList)
+	{
+		tile->overlay(overlayToUse);
+	}
+}
```
DanRStevens, line 34: You could potentially pass a `boolean` instead of a `Button&` (which could potentially have been a `const Button&`).

By that token, you could also have passed just the `overlay`, and determined it's value outside this method. Though that starts to duplicate code, so I can see why you'd have done it this way.

----

What happens if a tile has two overlays set? Or rather, if the `tileList` collections have non-overlapping tiles, and you switch from one overlay to another? does it leave behind ghostly highlights from the old overlay?

Does the code need to have old overlays unset? Can it support multiple overlays at the same time?

OPHD/Map/Tile.h
```diff
@@ -25,11 +25,11 @@ class Tile
...
-	Tile(NAS2D::Point<int> position, int depth, TerrainType index);
-	Tile(const Tile& other) = delete;
-	Tile& operator=(const Tile& other) = delete;
-	Tile(Tile&& other) noexcept;
-	Tile& operator=(Tile&& other) noexcept;
+	Tile(NAS2D::Point<int>, int, TerrainType);
+	Tile(const Tile&) = delete;
+	Tile& operator=(const Tile&) = delete;
+	Tile(Tile&&) noexcept;
+	Tile& operator=(Tile&&) noexcept;
```
DanRStevens, line 32: I find this makes the code less readable. I know parameter names don't need to be included in the header file, though they can be really useful as a form of documentation. I would encourage always including parameter names in header files.

OPHD/States/MapViewState.cpp
```diff
@@ -69,6 +69,43 @@ const std::map<Robot::Type, RobotMeta> RobotMetaTable
...
+	{
+		std::clamp(centerTile.position().x - commRange, 0, 299),
+		std::clamp(centerTile.position().y - commRange, 0, 149)
+	};
+
+	const NAS2D::Point commAreaEndPoint
+	{
+		std::clamp(centerTile.position().x + commRange, 0, 299),
+		std::clamp(centerTile.position().y + commRange, 0, 149)
+	};
```
DanRStevens, line 84: Magic numbers here. I assume those constants relate to standard map dimensions. We should be using a named constant, especially since the values are repeated.

----

We should probably add some kind of `Rectangle::clamp` method, to allow clamping a computed `Rectangle` within some knows bounds, such as map bounds.

----

Actually, due to the symmetry from the center point, you might be able to use `Rectangle::inset` for this, with a negative inset amount:
```cpp
const auto commArea = Rectangle<int>::Create(centerTile.position(), {0, 0}).inset(-commRange);
```

Of course that that would require clamping.

OPHD/States/MapViewState.cpp
```diff
@@ -69,6 +69,43 @@ const std::map<Robot::Type, RobotMeta> RobotMetaTable
...
+			if (centerTile.distanceTo(&tile) <= commRange)
+			{
+				if (std::find(tileList.begin(), tileList.end(), &tile) == tileList.end())
+				{
+					tileList.push_back(&tile);
+				}
+			}
+		}
+	}
+}
```
DanRStevens: You may want to consider using `PointInRectangleRange` for this loop:
```cpp
for (const auto point : PointInRectangleRange{area})
{
    auto& tile = tileMap.getTile(point);
    // ...
}
```

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -517,7 +517,11 @@ void MapViewState::nextTurn()
 	updateStructuresAvailability();
 	updateRoads();
 
+	// Overlay Updates
 	checkCommRangeOverlay();
+	btnToggleConnectednessClicked();
+	btnToggleCommRangeOverlayClicked();
```
DanRStevens: This seems a bit hacky. Might be a sign the button handler code should be refactored so it can be called directly.

DanRStevens (commented):
Had a few thoughts while reading through this.

OPHD/States/MapViewStateUi.cpp
ldicker83: The code has old overlays unset. Only one overlay can be set at a time. It's not the most efficient method but it's simple and does the job.

I opted to use Button& for future overlays.

OPHD/Map/Tile.h
ldicker83: Alright. I was starting to opt to keep them out of headers but I forget not everybody is using Visual Studio or VS Code. I'll move in the direction of leaving them in the headers.

OPHD/States/MapViewState.cpp
ldicker83: > Magic numbers here. I assume those constants relate to standard map dimensions. We should be using a named constant, especially since the values are repeated.

Agreed. This was a bit of a quick-and-dirty type of deal. I was kind of surprised I didn't have named constants or that `TileMap` didn't have an option. Upon further inspection I see that `TileMap` does have `TileMap::size()` which would have served for this but for whatever reason I didn't think of looking for a `size()` type of function. For maps I usually think of a `width` and `height` function that gets me exactly the values I need.

Definitely something to address in a future change set.

> 
> We should probably add some kind of `Rectangle::clamp` method, to allow clamping a computed `Rectangle` within some knows bounds, such as map bounds.
> 
> Actually, due to the symmetry from the center point, you might be able to use `Rectangle::inset` for this, with a negative inset amount:
> 
> ```c++
> const auto commArea = Rectangle<int>::Create(centerTile.position(), {0, 0}).inset(-commRange);
> ```
> 
> Of course that that would require clamping.

We could? But this seems to work just fine here?

OPHD/States/MapViewState.cpp
ldicker83: I didn't know we had that class. There were a lot of features and functions added to NAS2D that I'm unfamiliar with. This might be a good refactor though I'm not sure it would improve performance, only code readability (which... to be fair, since this isn't run per frame may be a non-issue)

OPHD/States/MapViewStateTurn.cpp
ldicker83: It absolutely is. It's mostly just avoiding pulling code out of the buttons because... well... it effectively works.

That said, the code between the two buttons is very very similar. It will change a bit going forward as new overlay buttons are added (there's really only room for one more so not by much). It could use some attention.

OPHD/States/MapViewState.cpp
DanRStevens: Efficiency should be identical. It was mainly added to make code easier and more expressive.

PR #811, opened 2021-02-27
Add non sqrt distance check

### LINKED ISSUE #806: Refactor inCommRange(NAS2D::Point<int>)
I noticed that `bool inCommRange(NAS2D::Point<int> position)` located in MapViewStateHelper.cpp reimplements basic distance functions when we should instead be using the `Tile::distanceTo(Tile*)` function. Will make the function a lot easier to read and reduce code redundancy.

ldicker83:
Closes #806 though in a slightly different way than originally proposed.

### DIFF
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -290,6 +290,8 @@ bool selfSustained(StructureID id)
  */
 bool inCommRange(NAS2D::Point<int> position)
 {
+
+
 	const auto maxCCRangeSquared = constants::ROBOT_COM_RANGE * constants::ROBOT_COM_RANGE;
 	const auto ccDistance = position - ccLocation();
 	if (ccDistance.lengthSquared() <= maxCCRangeSquared)
@@ -315,6 +317,12 @@ bool inCommRange(NAS2D::Point<int> position)
 	}
 
 	return false;
+}
+
+
+bool isPointInRange(NAS2D::Point<int> point1, NAS2D::Point<int> point2, int distance)
+{
+	return ((point2.x - point1.x) * (point2.x - point1.x)) + ((point2.y - point1.y) * (point2.y - point1.y)) <= distance * distance;
 }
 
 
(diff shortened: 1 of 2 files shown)


OPHD/States/MapViewStateHelper.cpp
```diff
@@ -318,6 +320,12 @@ bool inCommRange(NAS2D::Point<int> position)
 }
 
 
+bool isPointInRange(NAS2D::Point<int> point1, NAS2D::Point<int> point2, int distance)
+{
+	return ((point2.x - point1.x) * (point2.x - point1.x)) + ((point2.y - point1.y) * (point2.y - point1.y)) <= distance * distance;
```
DanRStevens, line 323: This could be shortened to:
```cpp
return (point2 - point1).lengthSquared() <= distance * distance;
```

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -290,6 +290,8 @@ bool selfSustained(StructureID id)
  */
 bool inCommRange(NAS2D::Point<int> position)
 {
+
+
 	const auto maxCCRangeSquared = constants::ROBOT_COM_RANGE * constants::ROBOT_COM_RANGE;
 	const auto ccDistance = position - ccLocation();
 	if (ccDistance.lengthSquared() <= maxCCRangeSquared)
```
DanRStevens: Spurious extra blank lines added here.

Edit: Removed in later commit.

DanRStevens (approved):
Looks good. Looks like it's been improved to handle multiple CCs.

On a minor efficiency note, the distance squaring in now repeated inside the loop for comm towers. It's not super clean to fix that though. Not unless you want to provide a function that checks a list of points against a single point. Might be room to eventually add a template function taking a lambda. Maybe a little out of scope here though.

OPHD/States/MapViewStateHelper.cpp
ldicker83: Yeah, I accidentally added those.

PR #815, opened 2021-03-02
Fixed inverted com range check

rmtoth:
Got error message when placing building near command center, rather than too far from it.


DanRStevens (approved):
Hah, oh geeze, thanks.

ldicker83:
Whoops. Good find!

PR #818, opened 2021-03-05
Added Yes/No dialog option when a user clicks delete in Load/Save dia…

belgianguy:
I found out there is actually already a Yes/No prompt available.
I've added it in the "Delete Game" flow to prompt a user before he/she deletes a game.
If they click yes, the deletion is carried out, if they click no, the user is returned to the Load or Save dialog without deletion taking place.

What I'm not sure about is the way to pass strings, I do want to keep them in Strings.h as much as possible, but is there a way to setup strings that have parameter placeholders in them?

E.g.: I now have the following:

```
if(doYesNoMessage(constants::WINDOW_FILEIO_TITLE_DELETE, "Are you sure you want to delete " + txtFileName.text() + "?")) {
			Utility<Filesystem>::get().del(filename);
}
```
Is there a way to have the string in Strings.h but have some sort of placeholder text that will be filled in with the actual save game name?

And I messed up the curly brace, it should go on the next line.

### DIFF
--- a/OPHD/UI/FileIo.cpp
+++ b/OPHD/UI/FileIo.cpp
@@ -172,11 +172,13 @@ void FileIo::btnFileIoClicked()
 
 void FileIo::btnFileDeleteClicked()
 {
-	std::string filename = constants::SAVE_GAME_PATH + txtFileName.text()+ ".xml";
+	std::string filename = constants::SAVE_GAME_PATH + txtFileName.text() + ".xml";
 
 	try
 	{
-		Utility<Filesystem>::get().del(filename);
+		if(doYesNoMessage(constants::WINDOW_FILEIO_TITLE_DELETE, "Are you sure you want to delete " + txtFileName.text() + "?")) {
+			Utility<Filesystem>::get().del(filename);
+		}
 	}
 	catch(const std::exception& e)
 	{
--- a/OPHD/UI/FileIo.cpp
+++ b/OPHD/UI/FileIo.cpp
@@ -176,7 +176,8 @@ void FileIo::btnFileDeleteClicked()
 
 	try
 	{
-		if(doYesNoMessage(constants::WINDOW_FILEIO_TITLE_DELETE, "Are you sure you want to delete " + txtFileName.text() + "?")) {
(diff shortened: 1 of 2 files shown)


ldicker83:
There used to be -- we've instead opted to use functions like `std::to_string` instead of formatting strings like `printf` style functions. It does mean it takes a bit more thought on how to phrase things so that string literals can be used in cases like this.

Beyond that this looks pretty good to me.

OPHD/UI/FileIo.cpp
```diff
@@ -172,11 +172,13 @@ void FileIo::btnFileIoClicked()
...
 {
-	std::string filename = constants::SAVE_GAME_PATH + txtFileName.text()+ ".xml";
+	std::string filename = constants::SAVE_GAME_PATH + txtFileName.text() + ".xml";
 
 	try
 	{
-		Utility<Filesystem>::get().del(filename);
+		if(doYesNoMessage(constants::WINDOW_FILEIO_TITLE_DELETE, "Are you sure you want to delete " + txtFileName.text() + "?")) {
+			Utility<Filesystem>::get().del(filename);
+		}
```
DanRStevens: Minor point: I think the convention in this project is to put a space after the `if` and before the opening `(`:
```cpp
if (...)
{
    ...
}
```

DanRStevens (commented):
Change looks good to me. Glad we've got that prompt setup now.

----

I have some misgivings about the current use of `String.h`, so I'm inclined to not worry about it too much right now. I suspect we'll revisit that aspect in the future.

The new C++20 standard will have better string formatting functions. See [`std::format`](https://en.cppreference.com/w/cpp/utility/format/format). The new functions would be useful for localization, as it would allow format strings to re-order where parameter values are inserted into the format string, which will help with translations and internationalization. It's also type safe, unlike `printf` style functions. This is something we'll probably look into when we update the targeted standard.

Compilers tend to lag a bit with support for new standards, so we're probably going to wait about a year before upgrading to C++20. We've done a bit of testing with the CI builds, and so far we can only really get preliminary support for most compilers using the `c++2a` target, which is the pre-release version of `c++20`. The upgrade would also mean dropping support for the Ubuntu 18.04 build. It might be a bit early still to drop that.

PR #821, opened 2021-03-06
Added check to IconGrid to suppress changed event when the same item …

### LINKED ISSUE #820: IconGrid should not raise an event if the current selection is the same as what is already selected
Original discussion: https://github.com/OutpostUniverse/OPHD/issues/34#issuecomment-791841189

Should be as simple as storing the current selection in a local variable and checking it against the new selected index.
If they are equal, nothing changed, and we should just return as nothing changed from the view of the IconGrid state.

belgianguy:
Closes #820 

At first I thought this could cause issues when it would be desirable to click an icon multiple times, say when you have two RoboDozers and you have to click the RoboDozer icon once for placing the first one, and click the RoboDozer icon again for when placing the second one. But this did not cause any problems.

Feel free to suggest other test cases that might need to be reviewed for this change to be accepted.

### DIFF
--- a/OPHD/UI/IconGrid.cpp
+++ b/OPHD/UI/IconGrid.cpp
@@ -86,7 +86,10 @@ void IconGrid::onMouseDown(EventHandler::MouseButton button, int x, int y)
 		return;
 	}
 
+	auto previousIndex = mSelectedIndex;
 	mSelectedIndex = translateCoordsToIndex(mousePoint - startPoint);
+
+	if (previousIndex == mSelectedIndex) { return; }
 
 	if (mSelectedIndex >= mIconItemList.size())
 	{
--- a/OPHD/UI/IconGrid.cpp
+++ b/OPHD/UI/IconGrid.cpp
@@ -89,8 +89,6 @@ void IconGrid::onMouseDown(EventHandler::MouseButton button, int x, int y)
 	auto previousIndex = mSelectedIndex;
 	mSelectedIndex = translateCoordsToIndex(mousePoint - startPoint);
 
-	if (previousIndex == mSelectedIndex) { return; }
-
 	if (mSelectedIndex >= mIconItemList.size())
 	{
 		mSelectedIndex = constants::NO_SELECTION;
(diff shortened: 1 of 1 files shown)


OPHD/UI/IconGrid.cpp
```diff
@@ -86,8 +86,11 @@ void IconGrid::onMouseDown(EventHandler::MouseButton button, int x, int y)
...
 	}
 
+	auto previousIndex = mSelectedIndex;
 	mSelectedIndex = translateCoordsToIndex(mousePoint - startPoint);
 
+	if (previousIndex == mSelectedIndex) { return; }
+
 	if (mSelectedIndex >= mIconItemList.size())
 	{
 		mSelectedIndex = constants::NO_SELECTION;
```
DanRStevens, line 96: Hmm, curious, what happens if the number of items in the `IconGrid` are updated? Particularly if items are removed.

I suppose `mSelectedIndex` would potentially be reset to `NO_SELECTION` during an update that changes the changes the list `size`. Though if that doesn't happen, this function may now exit before the `size` clamp, leaving `mSelectedIndex` set to an invalid out-of-bounds index.

Might be better to change the early return check to instead be a check around the `raiseChangedEvent()` call. That might also be a little more intuitive, as it would more directly address the problem at hand.

DanRStevens (commented):
Looks like this should work. There's one aspect I'd like checked first, which comes with a bit of a suggestion.

----

That is a rather valid point about re-selecting the same item, and how we might actually want to capture that event in some cases. I'm don't remember the details of the robot placement code, so I'm not certain why it may or may not be important there. Definitely something worth checking though.

OPHD/UI/IconGrid.cpp
belgianguy: Well, the index will always refer to an integer, if we want to cover all bases and make sure the same item is selected, we'd need to translate it into whatever that integer is pointing to and compare it to whatever the incoming value is pointing to.
But I don't think that knowledge exists at the IconGrid. Or is that what the selection_meta(int selectionMetaValue is for?

belgianguy:
I have moved the check around the raiseChangedEvent call.

DanRStevens (approved):
Change looks good.

----

Actually, now that I think about it, I suspect the change will prevent triggering multiple `raiseChangedEvent` calls when clicking in the dead space so no item is selected. If `translateCoordsToIndex` works the way I think it does, the older code may have still triggered multiple calls to `raiseChangedEvent`. That may be a nice bonus for the update.

----

Btw, now that you're part of the OPU organization, you can create branches and PRs right out of the main OPU repository, without having to work through a fork.

PR #823, opened 2021-03-08
Add route overlay

ldicker83:


OPHD/States/MapViewStateUi.cpp
```diff
@@ -427,13 +438,31 @@ void MapViewState::btnToggleCommRangeOverlayClicked()
...
+void MapViewState::btnToggleRouteOverlayClicked()
+{
+	if (mBtnToggleRouteOverlay.toggled())
+	{
+		mBtnToggleConnectedness.toggle(false);
+		mBtnToggleCommRangeOverlay.toggle(false);
+
+		btnToggleCommRangeOverlayClicked();
+		btnToggleConnectednessClicked();
+	}
```
DanRStevens, line 460: Kind of wondering if we might need some kind of `ToggleButtonGroup` control, which groups several options together, and tracks which one is currently active. It can be responsible for calling a `deactivate` method on the previously active button, and an `activate` method on the newly active button.

DanRStevens (approved):
Looks good to me.

OPHD/States/MapViewStateUi.cpp
ldicker83: Would make a lot of sense since we can also group check boxes and radio buttons in the same way.

belgianguy:
I'm getting an error on this line:

```
mBtnToggleRouteOverlay.image("ui/icons/route.png");
```
It complains it cannot find the file specified, but whenever I look in the directory that holds the icons, I only see a "route_overlay.png"

```
/OPHD/data/ui/icons/route_overlay.png
```

Is that a typo or do I need to refresh my assets somehow? I tried rebuilding and updating the git submodules, but no avail.
Perhaps someone could describe the process to me (I'm on Ubuntu, if that helps).

ldicker83:
I renamed it. I thought I pushed it. I guess I forgot? /me shrugs

Sorry! I swear I'm working on getting better about that!

PR #826, opened 2021-03-10
Make UIControl extend from Control and give Window a title property

belgianguy:
I did have to reload the submodules as my game would crash on "route.png" at:

```
mBtnToggleRouteOverlay.image("ui/icons/route.png");
```
And the asset was not available on my system. When I tried to check the git status in this directory, I was greeted with the "Detached HEAD" error message. So I looked up how to update them to the latest version. It seems by default it uses a --checkout option, which points to a commit, which then causes a detached HEAD. The below command is different and I hope it no longer causes the detached head issue.

```
git submodule update --remote --rebase
```
This allowed me to start the game and not get the "route.png" error.

I reloaded my submodules, but now I have uncommitted changes of which I don't know if I need to commit and push them.

image

### DIFF
--- a/OPHD/UI/Core/UIContainer.h
+++ b/OPHD/UI/Core/UIContainer.h
@@ -13,7 +13,7 @@
  * 
  * Generally not intended to be used by itself.
  */
-class UIContainer: public TextControl
+class UIContainer: public Control
 {
 public:
 	UIContainer();
--- a/OPHD/UI/Core/Window.cpp
+++ b/OPHD/UI/Core/Window.cpp
@@ -28,7 +28,7 @@ Window::Window(std::string newTitle) :
 		imageCache.load("ui/skin/window_bottom_right.png")
 	}
 {
-	text(newTitle);
+	title(newTitle);
 
 	Utility<EventHandler>::get().mouseButtonUp().connect(this, &Window::onMouseUp);
 	Utility<EventHandler>::get().mouseMotion().connect(this, &Window::onMouseMove);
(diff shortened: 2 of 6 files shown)


DanRStevens (approved):
This is very much the solution I was expecting.

----

I noticed the CI builds aren't running for your PR. I suspect it's because the code comes from a fork. If you create branches directly in the OutpostUniverse/OPHD repository, the CI builds will probably run. Working from the original repo would also make it easier for other people to pull code changes to test out locally. No need to add an extra remote to download changes from.

----

I see you've run into another forgotten submodule update. I'm afraid I'm not familiar with the flags you used in your example command. Normally I would `git checkout master` from within the submodule folder. Might also need a `git pull` if recent changes haven't been downloaded yet. That should get you the new files, and you'd be able to run at the point, though the submodule will likely show up as modified when viewing the parent repository.

You can commit an update to the parent repository, though ideally this should be done in the PR that references new assets, and the update should be done before the reference is added. If this is forgotten, then I usually fix the broken submodule reference in its own PR. Effectively it's a bug fix PR. I try to avoid making these bug fix updates in a PR that is unrelated, and merely depends on having the fix in place.

The "Detached HEAD" message isn't really an error. More like a notice, or warning. It doesn't mean anything is wrong. That's just the state that submodule are typically in. Since the parent repository references a specific commit in the submodule repository, it doesn't really make sense to update the submodule's checked out commit when new updates are available. This is intentionally a manual process. Hence the "Detached HEAD" state. It's more a notice, to remind you to checkout a branch before making new commits (or at least create a branch after making new commits, so you actually have a name to reference them).

As for why both `data` and `nas2d-core` are listed, it's because you didn't specify a submodule in your update command, so it updated all submodules. There are some changes applied to NAS2D since the last time we updated the submodule reference. They were minor code cleanup changes, with no real visible effects, hence no need to update the submodule reference in OPHD right away. When we have need to update the reference, perhaps for a bug fix, or new feature in NAS2D, then we'll make the update.

----

If you have local changes to `.gitignore` and `makefile`, that may suggest a deficiency in our current setup (or the documentation). Maybe consider opening a PR or issue to discuss. In the case of the `Makefile`, it might even be possible to get the behaviour you want using environment variables. My local environment has a few settings for things like the number of cores/parallelism to use when compiling. Sometimes I use variables to enable/disable debug symbol generation.

Issue #830, opened 2021-03-12
Serialization overhaul


DanRStevens:
The save/load code has a lot of direct calls to the XML code in NAS2D. There is a plan to update the NAS2D XML code from TinyXML1 (with local modifications and renames) to TinyXML2. Such an upgrade will break a lot of existing code. We should try to minimize the number of direct calls to XML code to ease the transition.

References:
- https://github.com/lairworks/nas2d-core/issues/211
- https://github.com/lairworks/nas2d-core/issues/797

----

As much of the save/load code needs to save and restore arbitrary data, much of it works with `XmlElement` objects. Instead, we could potentially have objects return a `Dictionary` of data that needs to be saved, or to accept a `Dictionary` of data to process when loading.

Reference: https://github.com/lairworks/nas2d-core/blob/master/NAS2D/Dictionary.h

We could have methods that convert `Dictionary` objects to/from `XmlElement` nodes. For loading code, the ParserHelper methods can be used to validate the presence of required fields. It can also be used to detect extra unknown fields.

Reference: https://github.com/lairworks/nas2d-core/blob/master/NAS2D/ParserHelper.h

Using these objects may allow for simplifications in the loading code. It should also allow the XML references to be lifted to a higher level and reduced in number.

ldicker83:
I'm game for using a Dictionary type object. Without looking too deep into the Dictionary class, can `std::tuple`'s be used? Because in most cases I can see basic key/value pairs but I'm wondering about the elements that have multiple attributes. Would it make sense to use a `std::tuple` for this or could it still be done with key/value pairs?

I suppose that's all down to the implementation details but I'm totally in favor of doing it this way. Not having to deal directly with XML processing would simplify so many things.

DanRStevens:
The underlying code uses `stringTo<T>` from StringUtils.h. That method doesn't currently support conversion to collection types.

We could potentially upgrade `stringTo`. Though the other solution is to parse strings outside of that method, with manual calls to `split` for `std::vector` type results.

DanRStevens:
All except one use of the XML `attributes` method has been replaced with uses of `Dictionary`. That goes a long way towards decoupling from the specific of the current XML implementation.

[...]

PR #831, opened 2021-03-13
Xml improvements

Brett208:
Must be paired with assets update to compile: https://github.com/OutpostUniverse/ophd-assets/pull/1

It looks like the assets are a submodule of this repo, but I had trouble linking them on my machine for some reason.

Related to #830. Not quite a string dictionary but tried to consolidate some of the xml related code.

The last commit is pretty big. I had trouble teasing it into smaller chunks due to the serialization involved. I was experimenting with template functions and enum lookups. Hopefully it looks ok...

This PR will allow someone to change the planet attributes without recompiling the source code.

### DIFF
--- a/OPHD/States/Planet.cpp
+++ b/OPHD/States/Planet.cpp
@@ -17,21 +17,6 @@ namespace {
 	constexpr auto PlanetSize = NAS2D::Vector{PlanetRadius * 2, PlanetRadius * 2};
 }
 
-const std::unordered_map<std::string, Planet::PlanetType> Planet::planetTypeTable
-{
-	{"none", PlanetType::None},
-	{"mercury", PlanetType::Mercury},
-	{"mars", PlanetType::Mars},
-	{"ganymede", PlanetType::Ganymede}
-};
-
-const std::unordered_map<std::string, Planet::Hostility> Planet::hostilityTable
-{
-	{"none", Hostility::None},
-	{"low", Hostility::Low},
-	{"medium", Hostility::Medium},
-	{"high", Hostility::High}
-};
 
 Planet::Planet(const Attributes& attributes) :
 	mAttributes(attributes),
@@ -84,61 +69,42 @@ void Planet::update()
 	renderer.drawSubImage(mImage, mPosition, NAS2D::Rectangle<int>::Create(spriteFrameOffset, PlanetSize));
 }
 
-
-void parseElementValue(Planet::PlanetType& destination, const NAS2D::Xml::XmlElement* element)
+namespace
 {
-	destination = stringToEnum(Planet::planetTypeTable, element->getText());
-}
+	Planet::Attributes parsePlanet(const NAS2D::Xml::XmlElement* xmlNode);
 
+	void parseElementValue(Planet::PlanetType& destination, const NAS2D::Xml::XmlElement* element);
+	void parseElementValue(Planet::Hostility& destination, const NAS2D::Xml::XmlElement* element);
 
-void parseElementValue(Planet::Hostility& destination, const NAS2D::Xml::XmlElement* element)
-{
-	destination = stringToEnum(Planet::hostilityTable, element->getText());
-}
-
-
-Planet::Attributes parsePlanet(const NAS2D::Xml::XmlElement* xmlNode)
-{
-	Planet::Attributes attributes;
-
-	for (const auto* node = xmlNode->iterateChildren(nullptr);
-		node != nullptr;
-		node = xmlNode->iterateChildren(node))
+	const std::unordered_map<std::string, Planet::PlanetType> planetTypeTable
 	{
-		const auto* element = node->toElement();
+		{"none", Planet::PlanetType::None},
+		{"mercury", Planet::PlanetType::Mercury},
+		{"mars", Planet::PlanetType::Mars},
+		{"ganymede", Planet::PlanetType::Ganymede}
+	};
 
-		if (element->value() == "PlanetType")
-		{
-			parseElementValue(attributes.type, element);
-		}
-		else if (element->value() == "ImagePath") {
-			parseElementValue(attributes.imagePath, element);
-		}
-		else if (element->value() == "Hostility") {
-			parseElementValue(attributes.hostility, element);
-		}
-		else if (element->value() == "MaxDepth") {
-			parseElementValue(attributes.maxDepth, element);
-		}
-		else if (element->value() == "MaxMines") {
-			parseElementValue(attributes.maxMines, element);
-		}
-		else if (element->value() == "MapImagePath") {
-			parseElementValue(attributes.mapImagePath, element);
-		}
-		else if (element->value() == "TilesetPath") {
-			parseElementValue(attributes.tilesetPath, element);
-		}
-		else if (element->value() == "Name") {
-			parseElementValue(attributes.name, element);
-		}
-		else if (element->value() == "MeanSolarDistance") {
-			parseElementValue(attributes.meanSolarDistance, element);
-		}
+	const std::unordered_map<std::string, Planet::Hostility> hostilityTable
+	{
+		{"none", Planet::Hostility::None},
+		{"low", Planet::Hostility::Low},
+		{"medium", Planet::Hostility::Medium},
+		{"high", Planet::Hostility::High}
+	};
+
+
+	void parseElementValue(Planet::PlanetType& destination, const NAS2D::Xml::XmlElement* element)
+	{
+		destination = stringToEnum(planetTypeTable, element->getText());
 	}
 
-	return attributes;
+
+	void parseElementValue(Planet::Hostility& destination, const NAS2D::Xml::XmlElement* element)
+	{
+		destination = stringToEnum(hostilityTable, element->getText());
+	}
 }
+
 
 std::vector<Planet::Attributes> parsePlanetAttributes()
 {
@@ -161,4 +127,51 @@ std::vector<Planet::Attributes> parsePlanetAttributes()
 	}
 
 	return planetAttributes;
+}
+
+
+namespace 
+{
+	Planet::Attributes parsePlanet(const NAS2D::Xml::XmlElement* xmlNode)
+	{
+		Planet::Attributes attributes;
+
+		for (const auto* node = xmlNode->iterateChildren(nullptr);
+			node != nullptr;
+			node = xmlNode->iterateChildren(node))
+		{
+			const auto* element = node->toElement();
+
+			if (element->value() == "PlanetType")
+			{
+				parseElementValue(attributes.type, element);
+			}
+			else if (element->value() == "ImagePath") {
+				::parseElementValue(attributes.imagePath, element);
+			}
+			else if (element->value() == "Hostility") {
+				parseElementValue(attributes.hostility, element);
+			}
+			else if (element->value() == "MaxDepth") {
+				::parseElementValue(attributes.maxDepth, element);
+			}
+			else if (element->value() == "MaxMines") {
+				::parseElementValue(attributes.maxMines, element);
+			}
+			else if (element->value() == "MapImagePath") {
+				::parseElementValue(attributes.mapImagePath, element);
+			}
+			else if (element->value() == "TilesetPath") {
+				::parseElementValue(attributes.tilesetPath, element);
+			}
+			else if (element->value() == "Name") {
+				::parseElementValue(attributes.name, element);
+			}
+			else if (element->value() == "MeanSolarDistance") {
+				::parseElementValue(attributes.meanSolarDistance, element);
+			}
+		}
+
+		return attributes;
+	}
 }
--- a/OPHD/States/Planet.cpp
+++ b/OPHD/States/Planet.cpp
@@ -146,28 +146,36 @@ namespace
 			{
 				parseElementValue(attributes.type, element);
 			}
-			else if (element->value() == "ImagePath") {
+			else if (element->value() == "ImagePath")
+			{
 				::parseElementValue(attributes.imagePath, element);
 			}
-			else if (element->value() == "Hostility") {
+			else if (element->value() == "Hostility")
+			{
 				parseElementValue(attributes.hostility, element);
 			}
-			else if (element->value() == "MaxDepth") {
+			else if (element->value() == "MaxDepth")
+			{
 				::parseElementValue(attributes.maxDepth, element);
 			}
-			else if (element->value() == "MaxMines") {
+			else if (element->value() == "MaxMines")
+			{
 				::parseElementValue(attributes.maxMines, element);
 			}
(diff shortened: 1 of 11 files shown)


OPHD/States/Planet.cpp
```diff
@@ -15,6 +17,19 @@ namespace {
 	constexpr auto PlanetSize = NAS2D::Vector{PlanetRadius * 2, PlanetRadius * 2};
 }
 
+const std::unordered_map<std::string, Planet::PlanetType> Planet::planetTypeTable{
```
ldicker83, line 20: Brace placement

OPHD/States/Planet.cpp
```diff
@@ -15,6 +17,19 @@ namespace {
...
+	{"none", PlanetType::None},
+	{"mercury", PlanetType::Mercury},
+	{"mars", PlanetType::Mars},
+	{"ganymede", PlanetType::Ganymede}
+};
+
+const std::unordered_map<std::string, Planet::Hostility> Planet::hostilityTable{
```
ldicker83, line 27: Brace placement

OPHD/States/Planet.cpp
```diff
@@ -66,3 +81,80 @@ void Planet::update()
...
+{
+	Planet::Attributes attributes;
+
+	for (const auto* node = xmlNode->iterateChildren(nullptr);
+		node != nullptr;
+		node = xmlNode->iterateChildren(node))
+	{
+		const auto* element = node->toElement();
+
+		if (element->value() == "PlanetType") {
```
ldicker83, line 108: Brace placement

OPHD/States/Planet.cpp
```diff
@@ -66,3 +81,80 @@ void Planet::update()
...
+
+	std::vector<Planet::Attributes> planetAttributes;
+
+	auto rootElement = xmlDocument.firstChildElement(rootElementName);
+	for (const auto* node = rootElement->iterateChildren(nullptr);
+		node != nullptr;
+		node = rootElement->iterateChildren(node))
+	{
+		std::string elementName("Planet");
+		if (node->value() != elementName) {
```
ldicker83, line 153: Brace placement

OPHD/XmlSerializer.h
```diff
@@ -1,9 +1,49 @@
...
 // Throws a runtime_error if the xml is ill formed or the root element name is incorrect
 NAS2D::Xml::XmlDocument openXmlFile(std::string filename, std::string rootElementName);
+
+int elementToInt(const NAS2D::Xml::XmlElement* element);
+
+template<typename T>
+T stringToEnum(const std::unordered_map<std::string, T>& table, std::string value)
+{
+	auto it = table.find(NAS2D::toLowercase(value));
+	if (it != table.end()) {
```
ldicker83, line 22: Brace placement

OPHD/XmlSerializer.h
```diff
@@ -1,9 +1,49 @@
...
+
+	throw std::runtime_error("Unable to parse enum with value of " + value);
+}
+
+
+// Create an overload to parse specific enums
+template<typename T, std::enable_if_t<!std::is_enum<T>::value, bool> = true>
+void parseElementValue(T& destination, const NAS2D::Xml::XmlElement* element)
+{
+	if constexpr (std::is_same_v<T, int>) {
```
ldicker83, line 34: Brace placement

ldicker83:
Looked over it. Generally no objections, looks pretty good. I made a few notes of areas where brace placement is inconsistent with the rest of the code. With that corrected this PR should be good to go.

OPHD/XmlSerializer.h
```diff
@@ -1,9 +1,49 @@
...
 // Open an xml document
 // Throws a runtime_error if the xml is ill formed or the root element name is incorrect
 NAS2D::Xml::XmlDocument openXmlFile(std::string filename, std::string rootElementName);
+
+int elementToInt(const NAS2D::Xml::XmlElement* element);
+
+template<typename T>
+T stringToEnum(const std::unordered_map<std::string, T>& table, std::string value)
+{
+	auto it = table.find(NAS2D::toLowercase(value));
```
DanRStevens, line 21: There's a bit of an implicit assumption here that key values for the `table` will always be all lowercase. The behavior here can potentially be quite confusing if that's not the case, as lookup would always fail, even if the `value` exactly matched a key.

I would lean towards making this case sensitive. Less confusion if someone uses a table with keys that have non-lowercase letters. Plus, if we're dealing with machine processed data, the case will always be exact, and we would catch inconsistencies more quickly.

DanRStevens (commented):
I really like the idea of moving that data to a data file.

I had a few suggestions and observations looking at the last commit, though they were all pretty minor.

OPHD/States/Planet.cpp
```diff
@@ -66,3 +81,80 @@ void Planet::update()
...
+void parseElementValue(Planet::PlanetType& destination, const NAS2D::Xml::XmlElement* element)
+{
+	destination = stringToEnum(Planet::planetTypeTable, element->getText());
+}
+
+
+void parseElementValue(Planet::Hostility& destination, const NAS2D::Xml::XmlElement* element)
+{
+	destination = stringToEnum(Planet::hostilityTable, element->getText());
+}
```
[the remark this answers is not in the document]
Brett208: This is probably a premature optimization. I want to move some large strings (definitions of the planets) into the xml file in the near future and I figured we may not want to pass large chunks of data in the return.

I'll make this a return for now as suggested and we can revisit if anyone ever notices a slowdown...

OPHD/XmlSerializer.cpp
```diff
@@ -0,0 +1,24 @@
...
+		throw std::runtime_error(filename + " has malformed XML: Row: " + std::to_string(xmlDocument.errorRow()) +
+			" Column: " + std::to_string(xmlDocument.errorCol()) + " : " + xmlDocument.errorDesc());
+	}
+
+	const auto* xmlRootElement = xmlDocument.firstChildElement(rootElementName);
+	if (!xmlRootElement) {
+		throw std::runtime_error(filename + " does not contain required root tag of <" + rootElementName + ">");
+	}
+
+	return xmlDocument;
```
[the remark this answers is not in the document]
Brett208: I had assumed the xmlDocument needed to be passed out due to lifetime, but I didn't actually test. My original thought was what you were thinking about just passing the root element, but I didn't try it due to the lifetime consideration. Maybe we can test to see if it matters or put the xmlDocument into a unique_ptr to see if that would preserve it even if only the xmlRootElement pointer is passed out?

Brett208:
The unnamed namespace work is a bit hairy due to the underlying structure. There are really 2 different sets of information being represented in Planet.h (the attributes of a planet and how to draw/manipulate in the UI a planet on screen). That logic is partially split between PlanetSelectState and Planet, making it even more complex. I think disentagling that is probably not a good idea in this thread though.

Made the easier style changes in addition to the anonymous namespace. I'll work the other points later.

Thanks for the thorough and quick feedback.

OPHD/States/Planet.cpp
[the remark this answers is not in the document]
Brett208: EDIT: I was attempting to keep function names the same between parsing an enum and parsing a basic value. Since type deduction is not possible based on return values of a function, I think using a return value here isn't possible. The alternative if we want a return value would be to provide separate names for each function. I was hoping to keep function names the same to possibly allow collapsing if/else statement logic cases in future iterations.

OPHD/XmlSerializer.h
```diff
@@ -1,9 +1,49 @@
...
+		destination = static_cast<float>(std::atof(element->getText().c_str()));
+	}
+	else if constexpr (std::is_same_v<T, double>) {
+		destination = std::atof(element->getText().c_str());
+	}
+	else if constexpr (std::is_same_v<T, std::string>) {
+		destination = element->getText();
+	}
+	else {
+		throw std::logic_error("Unable to parse type " + std::string(typeid(T).name()) + " from XML");
```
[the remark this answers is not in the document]
Brett208: I didn't realize you could disable run time type information. 

Procedures: https://docs.microsoft.com/en-us/cpp/build/reference/gr-enable-run-time-type-information?view=msvc-160

I tried disabling run time type information and received about a 7 warnings: 

> Warning	C4541	'dynamic_cast' used on polymorphic type 'Control' with /GR-; unpredictable behavior may result	\OPHD\UI\Core\RadioButton.cpp	67	

The program crashes when this dynamic_cast is reached, which is before the typeid call mentioned by @DanRStevens, so I wasn't able to directly test. Oddly, it didn't provide a warning about the typeid like the dynamic_cast did. 

I guess we probably do not want to remove the concept of dynamic_cast so probably safe to use typeid? Although it is good to know these 2 features are tied to a compiler settings. I had just assumed they were always available if you had a c++11 compliant compiler since they are features of the language itself... C++ seems to constantly find the most complex ways to express itself.

Brett208:
I wasn't thrilled with the results of adding the anonymous namespace. Having to prepend the :: qualifier to some of the functions may cause noise later.

I am complete with updates unless the additions I made need further refinement before merging.

OPHD/States/Planet.cpp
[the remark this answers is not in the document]
DanRStevens: As of C++17, the compiler is guaranteed to do the optimization for you:
https://en.cppreference.com/w/cpp/language/copy_elision

When returning large objects, the compiler will effectively pass in a hidden parameter with the address of the variable being copied to, and will instead construct the value in place. Hence no copy during the return.

----

You can potentially use template methods to parameterize the return type. If the functions have different implementations, this can be handled with template specialization. Though often, just giving different functions different names may be easier and sometimes makes more sense. I'd need to look over the code again to consider this.

OPHD/XmlSerializer.cpp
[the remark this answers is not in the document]
DanRStevens: I looked a bit at the XML code. It seems lifetime is tied to the `XmlDocument`, so if it's not being returned, or otherwise preserved, the root node would go out of scope upon return, and you'd be left with a dangling pointer.

I had a local branch in NAS2D where I was trying to deal with this sort of thing. I think I attempted to pass in a lambda for the processing code, which was called with the `rootElement`, rather than try to return it.

I think maybe ignore my original comment on this for now. This might be more relevant for consideration in a future PR.

OPHD/XmlSerializer.h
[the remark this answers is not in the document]
DanRStevens: That was an interesting investigation. I had kind of forgotten that `dynamic_cast` was also related to RTTI. I used to avoid it a bit for that reason. It is useful though, and I can see us continuing to use it in this project, so probably safe to assume that RTTI is enabled.

OPHD/States/Planet.cpp
```diff
@@ -66,3 +68,118 @@ void Planet::update()
...
+	}
+
+	return planetAttributes;
+}
+
+
+namespace 
+{
+	Planet::Attributes parsePlanet(const NAS2D::Xml::XmlElement* xmlNode)
+	{
```
DanRStevens, line 136: Not that it matters, though you could potentially have put this in the first unnamed namespace, possibly even replacing the function prototype. Though there is nothing wrong about re-opening an unnamed namespace and keeping the implementation separate.

OPHD/States/Planet.cpp
```diff
@@ -66,3 +68,118 @@ void Planet::update()
...
+			const auto* element = node->toElement();
+
+			if (element->value() == "PlanetType")
+			{
+				parseElementValue(attributes.type, element);
+			}
+			else if (element->value() == "ImagePath")
+			{
+				::parseElementValue(attributes.imagePath, element);
+			}
```
DanRStevens, line 152: Hmm, yes, I see the discrepancy with the `::`. It seems you have two functions `parseElementValue`, defined in different namespaces. That could get confusing.

DanRStevens (commented):
I think this is good to merge. I just want to take a bit of a closer look at the namespace scope issue, and why `::` is needed in a few places. Maybe there's another workaround.

DanRStevens:
Yeah, maybe I'll look into that as a post-merge thing.

OPHD/States/Planet.cpp
Brett208: Thank you for the review. My goal was to keep the function names the same. I reasoned it might be possible to use a table to match argument names to locations in the structure to remove the if / else logic statements to place each value. I thought doing this would be easier if function overloads / templates were used. It gets complicated because of the detailed knowledge required about the enum to parse it properly, so the local module will need to define overloads for which enums it wants to support parsing. If this seems silly, I'm OK if the enum parse functions have custom names...

I think it might be worth trying to limit enums when possible when parsing to simplify. We could just represent hostility as a scale of 1 to 10 using an int value instead of 'low' 'medium' 'high' for example which would cut out the custom parsing logic.

DanRStevens: The problem is the different attributes have different types, and selecting which function overload or template instantiation to call is a compile time thing. It's not compatible with a runtime loop. Trying to use a table would require some kind of compile time iteration.

One of the things I was thinking about was being able to map an enum type to the lookup table used to map strings to enum values. It is possible to use templates to map types to values:
https://gpfault.net/posts/mapping-types-to-values.txt.html

I was thinking if we used such a solution, we could update the `stringToEnum` function to not require the first parameter:
```cpp
template<typename T>
T stringToEnum(const std::unordered_map<std::string, T>& table, std::string value)
```
Instead, the table could be looked up internally using such a mapping template on the template parameter.

Taking that idea further, it could be combined with the `parseElementValue` function:
```cpp
template<typename T, std::enable_if_t<!std::is_enum<T>::value, bool> = true>
void parseElementValue(T& destination, const NAS2D::Xml::XmlElement* element)
```
You could potentially remove the `std::enable_if` guard, and instead use a `if constexpr` to select whether to call `parseElementValue` or `stringTo`. And you might as well, since there would no longer be an external `table` parameter to pass in, so it would be very easy to call in such a context.

Would then make it easy to use a return value, rather than an out parameter.

Of course, that all hinges on being able to map a type to a value. I gave it a bit of a try, though if you define the type mapping in a different namespace than it's used, that seems to cause issues.

----

```cpp
// Specialize this for each EnumType with a lookup table (not shown)
template <typename EnumType>
struct EnumValueName { static const std::unordered_map<std::string, EnumType> table; };

template<typename EnumType>
T stringToEnum(std::string value)
{
    const auto& table = EnumValueName<EnumType>::table;
    // Old implementation ...
}

template<typename T>
T parseElementValue(const NAS2D::Xml::XmlElement* element)
{
    try 
    {
        const auto value = element->getText();
        if constexpr (std::is_enum_v<T>)
        {
            return stringToEnum<T>(value);
        }
        else
        {
            return NAS2D::stringTo<T>(value);
        }
    }
    // Old implementation ...
}
```

PR #832, opened 2021-03-14
Treat Air Shaft building as a Tube Intersection above and underground

### LINKED ISSUE #827: Is the Air Shaft not supposed to be tube-like in behavior? Buildings cannot be built right against it.
The Air Shaft structure (the structure that the RoboDigger builds), both above and underground, should probably also be able to function as an "intersection tube".
Currently only Tube structures are able to be built right against it, but not regular buildings, even though I would assume it would carry these fluids/gases.
They do look very much alike (tube intersection and Air Shaft), of course, it isn't in the player's best interest to do so below ground, but it shouldn't be impossible (unless there is good reason to).

MapViewStateHelper has a function
```
checkStructurePlacement
```

which returns false.

Even though the Air Shaft is a Tube type structure and it has connector direction CONNECTOR_INTERSECTION, but it is not used when comparing the connector directions.

belgianguy:
Closes #827

Added logic that treats both the above and underground Air Shaft building as a Tube structure, meaning buildings can be built against it, as well as other tubes.

### DIFF
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -90,27 +90,22 @@ bool checkTubeConnection(Tile& tile, Direction dir, ConnectorDir sourceConnector
  */
 bool checkStructurePlacement(Tile& tile, Direction dir)
 {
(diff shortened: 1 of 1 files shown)


DanRStevens (approved):
Looks good to me.

PR #835, opened 2021-03-24
Update mouse button names

ldicker83:


DanRStevens:
Oh hey, thanks for taking care of this. I only just now realized my search/replace had updated OPHD too, and left uncommitted changes to that project.

PR #837, opened 2021-03-26
renamed StructureID to StructureTypeID, since it identified a type of…

rmtoth:
… structure (from the catalog) rather than a specific structure (from the map).

This change is in preparation of renaming SKey to StructureID.

The change itself is a pure search-and-replace, no logic was altered.

### DIFF
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -113,7 +113,7 @@ enum ConnectorDir
  *			the structure Class code which is used to group like structures into
  *			lists for structure updates.
  */
-enum StructureID
+enum StructureTypeID
 {
 	SID_NONE,
 
--- a/OPHD/Map/TileMap.cpp
+++ b/OPHD/Map/TileMap.cpp
@@ -601,7 +601,7 @@ void TileMap::AdjacentCost(void* state, std::vector<micropather::StateCost>* adj
 			{
 				cost *= static_cast<float>(adjacentTile.index()) + 1.0f;
 			}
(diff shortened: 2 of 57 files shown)


DanRStevens:
Looks like a forward declaration was used for `StructureManager` when a full header include was required:
> OPHD/StructureComponent.h:42:46: error: invalid use of incomplete type 'class StructureManager' [-Werror]
>   return NAS2D::Utility<StructureManager>::get().get<T>(s);
>          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
> OPHD/StructureComponent.h:6:7: note: forward declaration of 'class StructureManager'
>  class StructureManager;
>        ^~~~~~~~~~~~~~~~

Should be easy to fix with a `#include`.

rmtoth:
Unfortunately it's not quite that easy. I've managed to create a circular dependency that I wrongly thought would be safe due to late codegen of templated code.

`StructureManager` contains `std::map<StructureComponent::ComponentTypeID, std::map<SKey, StructureComponent*>> mComponents;` and thus needs be defined after `StuctureComponent`.
`StructureComponent` uses `GetComponent`.
`GetComponent` references `StructureManager`.

DanRStevens:
Ahh, I see. That does take a little more time to solve.

@ldicker83: Just curious if you noticed the failing builds before merging. I see there's another PR that sounds related to a fix, though I don't see any obvious comment or reference to confirm they are related.

PR #839, opened 2021-03-27
Fix linux build of entityComponentSystem branch

rmtoth:
Improved definition sequence to satisfy GCC.

SKey is now defined in StructureManager.h, which makes sense anyway - it defines an instance of a structure, which is not directly related to StructureComponent.

Template specialization of member function is now done in the enclosing namespace since only MSVC allows specialization within the class definition.

Added UNUSED macro to common.h to silence compiler warnings about unused variables.

### DIFF
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -366,3 +366,8 @@ const auto formatDiff = [](int diff)
 {
 	return ((diff > 0) ? "+" : "") + std::to_string(diff);
 };
+
+/**
+ * Silence compiler warnings about unused variable
+ */
+#define UNUSED(x) (void)(x)
--- a/OPHD/StructureManager.cpp
+++ b/OPHD/StructureManager.cpp
@@ -4,6 +4,7 @@
 #include "ProductPool.h"
 #include "IOHelper.h"
 #include "PopulationPool.h"
+#include "StructureComponent.h"
 #include "Map/Tile.h"
 #include "Things/Robots/Robot.h"
 #include "Things/Structures/Structures.h"
@@ -248,6 +249,7 @@ void StructureManager::removeStructure(Structure* structure)
 	SKey s = SKey(structure);
 	for (auto& [componentTypeID, table] : mComponents)
 	{
+		UNUSED(componentTypeID);
 		auto cit = table.find(s);
 		if (cit != table.end())
 		{
--- a/OPHD/StructureManager.h
+++ b/OPHD/StructureManager.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include "StructureComponent.h"
 #include "Things/Structures/Structure.h"
 
 
@@ -13,6 +12,31 @@ namespace NAS2D {
 class Tile;
 class PopulationPool;
 struct StorableResources;
+
+class StructureComponent;
+typedef int ComponentTypeID; // TODO: replace by enum class?
+
+/**
+ * Key type for identifying a specific structure instance.
+ * The key for any given structure is guaranteed to remain unchanged for the lifetime of the structure.
+ * The key for any given structure is guaranteed to be unique during the lifetime of the structure.
+ *
+ * Every structure has a Structure instance. The Structure pointer is used as key to allow O(1) access
+ * to the Structure instance. This is an internal detail and should not be relied upon by code handling the key.
+ */
+class SKey
+{
+private:
+	Structure* mStructure;
+public:
+	SKey(Structure* structure) : mStructure(structure) {}
+
+	/** Comparison operators to allow using this type in ordered containers such as maps and sets. */
+	bool operator<(const SKey& rhs) const { return mStructure < rhs.mStructure; }
+
+	/** Do not call this function directly. It is intended only for GetComponent/TryGetComponent. */
+	Structure* getInternal() { return mStructure; }
+};
 
 
 /**
@@ -69,6 +93,8 @@ public:
 		{
 			throw std::runtime_error("Structure::Attach() was called on a Structure that already had the component!");
 		}
+#else
+		UNUSED(success);
 #endif
 	}
 
@@ -92,16 +118,6 @@ public:
 	}
 
 	/**
-	 * Return a reference to the Structure type belonging to a structure.
-	 * This allows writing code that's agnostic to the SKey type.
-	 */
-	template<>
-	Structure& get<Structure>(SKey s)
-	{
-		return *s.getInternal();
-	}
-
-	/**
 	 * Return a pointer to the given StructureComponent type belonging
 	 * to a structure, if it has the corresponding component type.
 	 * Otherwise return nullptr.
@@ -112,18 +128,8 @@ public:
 		auto& table = mComponents[ComponentTy::componentTypeID];
 		auto it = table.find(s);
 		if (it != table.end())
-			return reinterpret_cast<ComponentTy*>(it->second);
+			return static_cast<ComponentTy*>(it->second);
 		return nullptr;
-	}
-
-	/**
-	 * Return a pointer to the Structure type belonging to a structure.
-	 * This allows writing code that's agnostic to the SKey type.
-	 */
-	template<>
-	Structure* tryGet<Structure>(SKey s)
-	{
-		return s.getInternal();
 	}
 
 private:
@@ -144,8 +150,29 @@ private:
 	 * Only keys to structures that actually have a given StructureComponent type
 	 * are present in the respective sub-tables.
 	 */
-	std::map<StructureComponent::ComponentTypeID, std::map<SKey, StructureComponent*>> mComponents;
+	std::map<ComponentTypeID, std::map<SKey, StructureComponent*>> mComponents;
 
 	int mTotalEnergyOutput = 0; /**< Total energy output of all energy producers in the structure list. */
 	int mTotalEnergyUsed = 0;
 };
+
+/**
+ * Return a reference to the Structure type belonging to a structure.
+ * This allows writing code that's agnostic to whether Structure is a component or not.
+ */
+template<>
+inline Structure& StructureManager::get<Structure>(SKey s)
+{
+	return *s.getInternal();
+}
+
+/**
+ * Return a pointer to the Structure type belonging to a structure.
+ * This allows writing code that's agnostic to whether Structure is a component or not.
+ */
+template<>
+inline Structure* StructureManager::tryGet<Structure>(SKey s)
+{
+	return s.getInternal();
+}
+
--- a/OPHD/StructureComponent.h
+++ b/OPHD/StructureComponent.h
@@ -1,33 +1,7 @@
 #pragma once
 
+#include "StructureManager.h"
 #include <NAS2D/Utility.h>
-
-class Structure;
-class StructureManager;
-class StructureComponent;
-
-
-/**
- * Key type for identifying a specific structure instance.
- * The key for any given structure is guaranteed to remain unchanged for the lifetime of the structure.
- * The key for any given structure is guaranteed to be unique during the lifetime of the structure.
- *
- * Every structure has a Structure instance. The Structure pointer is used as key to allow O(1) access
- * to the Structure instance. This is an internal detail and should not be relied upon by code handling the key.
- */
-class SKey
-{
-private:
-	Structure* mStructure;
-public:
-	SKey(Structure* structure) : mStructure(structure) {}
-
-	/** Comparison operators to allow using this type in ordered containers such as maps and sets. */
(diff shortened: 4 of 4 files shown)


OPHD/Common.h
```diff
@@ -366,3 +366,8 @@ const auto formatDiff = [](int diff)
 {
 	return ((diff > 0) ? "+" : "") + std::to_string(diff);
 };
+
+/**
+ * Silence compiler warnings about unused variable
+ */
+#define UNUSED(x) (void)(x)
```
DanRStevens, line 373: This makes me a bit squeamish. I like to avoid macros whenever possible. Sometimes you can avoid unused variable warnings by commenting out the name of function parameters. For structured bindings, well, you could rewrite to avoid using structured bindings.

OPHD/StructureManager.cpp
```diff
@@ -248,6 +249,7 @@ void StructureManager::removeStructure(Structure* structure)
 	SKey s = SKey(structure);
 	for (auto& [componentTypeID, table] : mComponents)
 	{
+		UNUSED(componentTypeID);
```
DanRStevens, line 252: I would probably avoid using structured bindings here just to avoid using the `UNUSED` macro.

OPHD/StructureManager.h
```diff
@@ -69,6 +93,8 @@ class StructureManager
 		{
 			throw std::runtime_error("Structure::Attach() was called on a Structure that already had the component!");
 		}
+#else
+		UNUSED(success);
 #endif
```
DanRStevens, line 98: If the #define guard was removed the UNUSED macro could be avoided.

OPHD/StructureManager.h
```diff
@@ -112,20 +128,10 @@ class StructureManager
 		auto& table = mComponents[ComponentTy::componentTypeID];
 		auto it = table.find(s);
 		if (it != table.end())
-			return reinterpret_cast<ComponentTy*>(it->second);
+			return static_cast<ComponentTy*>(it->second);
```
DanRStevens, line 131: I would expect either `reinterpret_cast` or `dynamic_cast` here. If there's a chance it could fail, `dynamic_cast` probably makes more sense.

Though being able to avoid casts completely would be better. I'm wondering if that might be possible somehow.

DanRStevens (commented):
Had a few small thoughts on this.

OPHD/Common.h
rmtoth: what's wrong with macros? They're a powerful tool. Rewriting code to not use structured bindings just because of a false positive compiler warning seems odd.

OPHD/StructureManager.h
rmtoth: sure, I can throw in release builds if you'd prefer that

OPHD/StructureManager.h
rmtoth: reinterpret_cast would be wrong - you'd get the wrong address with multiple inheritance, for instance. dynamic_cast involves RTTI, which is not necessary in this case we know the type. The only reason to use dynamic_cast would be to dynamically discover whether an instance is in fact the type you're querying.

If you really want to avoid casts completely (why?) then I mentioned a possible solution [here](https://github.com/OutpostUniverse/OPHD/pull/836#discussion_r603556121). I do not suggest actually doing that though.

OPHD/Common.h
ldicker83: It's another stylistic choice -- we tend to avoid macros because they're a great source of errors and abuses of type system issues. Instead we've been using C++17 constexpr and similar constructs in place of the preprocessor.

OPHD/StructureManager.h
ldicker83: Throw in release is preferred.

OPHD/StructureManager.h
ldicker83: I would agree that static_cast is the correct cast to use here. Reinterpret is ... scary. Dynamic is unnecessary if we know the type we're expecting. If it's wring it explodes anyway and the compiler should be able to determine that.

As for avoiding casting... that would be ideal but would require an entire rewrite of much of the code base. I'm not even sure how I'd do it differently other than using a Data Oriented Programming approach over OOP (doable... but tremendous effort would go into it and for the gain of avoiding casts). At least with the static_cast operator we know exactly where we're subverting the type system.

OPHD/Common.h
DanRStevens: I highly recommend doing a Google search for why C++ macros shouldn't be used.

I think the main one for me is they don't respect scope. This can lead to lots of nasty conflicts, particularly with the Windows headers which define a lot of macros. If you `#define max` as a macro function, then try to use `std::max`, the macro will do a textual replacement on the call to `std::max` and you'll end up with something weird like `std::(a<b?a:b)`. Have fun diagnosing the syntax error on that, particularly since you don't normally see the result of pre-processing and may have no idea a macro was ever even invoked there.

There's issues about multiple evaluation of parameters when passed expressions with side effects. How many times are `x` and `y` incremented below? Maybe once? Maybe twice? Maybe it depends on how they compare?
```cpp
#define max(a, b) a < b ? a : b
max(x++, y++); // x++ < y++ ? x++ : y++
```

They also have issues with precedence and order of operations if parameters are not properly quoted and an expression was passed.
```cpp
#define TWO 1+1

TWO * TWO; // 1+1 * 1+1 = 1 + (1 * 1) + 1 = 3
```

DanRStevens: As for the structured binding bit. If you're not using one of the extracted values, it might not have been appropriate to use structured bindings in the first place. If you only wanted a single value, then accessing that single field makes more sense. Though yes, I get that the syntax is kind of nice, and I do like being able to use it.

OPHD/StructureManager.h
DanRStevens: That's a good point about `reinterpret_cast` being potentially wrong for multiple inheritance. Though I also tend to avoid multiple inheritance. Or at least, the only times I use multiple inheritance are when working with interfaces, where the problems of multiple inheritance generally go away.

It's pretty rare that I find the need for `dynamic_cast`. Plus `dynamic_cast` may not work if you disable RTTI in compiler settings.

For `static_cast`, I normally think of that when you are actively doing something that requires the compiler to insert code. Converting between `int` and `float` would use a `static_cast`. Or invoking a type conversion member function. With `reinterpret_cast`, there is no compiler inserted code. It just treats the memory as if it was a different compile-time known type. As that seems to be the case here, I would expect `reinterpret_cast` to be appropriate.

Question: What happens if you use `static_cast` on a type that has an overloaded type conversion method? I suspect it would produce the wrong results here, and what you really want is `reinterpret_cast`.

----

I like to avoid casts when possible because they are a way of defeating the type system and removing safety checks and warnings. If you don't really need to do it, then it's best to avoid casts. That's especially true when refactoring, as changes to types can cause casts to do invalid things, and to silence warnings while doing it. Besides, with modern template facilities, you can often get correct type safe behaviour without disabling any warnings.

In this particular case, I'm not entirely sure how to avoid casts. Maybe with a bit of thinking, something could be done. Like say, if there are a fixed number of component types, you could have separate `std::vector` to hold each type.

OPHD/Common.h
rmtoth: Those are strawman arguments.

Yes, the win32 header's `max` define is bad. Yes, poorly scoped macros are bad. Yes, you can write macros that abuse the type system. But those are simply examples of _bad code_. That doesn't mean the concept of macros is evil. It's possible to write bad code in any language.

Writing code to generate code is a well-established practice, and I simply don't believe that macros are categorically worse than any other means of doing so. You can write python scripts that generate headers. You can define macros to expand .inc files into a myriad of utility functions. In some cases you can use templates to accomplish your goals. Each way of generating code has its own strengths and weaknesses, and I'd generally argue that you should use the right tool for the right job. That said, there is value in using few tools in general, since each tool has a learning curve. But that's a very weak reason not to use macros.

This is turning quite philosophical.

OPHD/StructureManager.h
rmtoth: >For static_cast, I normally think of that when you are actively doing something that requires the compiler to insert code.

It won't insert code if there was no need to do so. Static_cast simply ensures you're converting between related types in a correct way. The cases where it will insert code are the cases where it _needs_ to insert code for the cast to be correct.

>As that seems to be the case here, I would expect reinterpret_cast to be appropriate.

attachComponent takes a StructureComponent pointer. If you pass it a pointer to a derived class (which is the intended pattern), then that's implicitly a static cast. Using reinterpret_cast to get back the original pointer is simply wrong. Forbidding use of multiple inheritance makes the invalid code happen to work, but that's a bad reason not to use the correct casting method.

>Question: What happens if you use static_cast on a type that has an overloaded type conversion method? I suspect it would produce the wrong results here, and what you really want is reinterpret_cast.

If the overloaded type conversion method does the wrong thing, then that's bad code.

rmtoth:
Addressed most comments in #840

OPHD/Common.h
DanRStevens: There isn't really a concept of poorly scoped macros. Macros just aren't scoped. To control scope of macros you mostly just isolate them at the translation unit level. Unless perhaps you want to undefine them after use, though that's often not very practical. The next best alternative is to give the name some kind of lengthy unique prefix, which makes them less pleasant to use.

Yes, you can write bad code in any language. Though with that said, some features are more error prone than others. Macros are a particularly error prone feature, and the language has evolved in a way that there are now often alternatives to using them.

The point is more that when there are alternatives that don't involve using macros, then alternative solutions should perhaps be used. Inline functions are a better alternative to function macros. Template functions are a better alternative to untyped macro functions that need to work with multiple types. Typed constants are a better alternative to untyped macro defined values.

If your use is X-Macros, then yes, you do probably need to use macros. There's no real alternative to that offered by other language features.

Similarly, template functions provide a rather limited and controlled way of doing code gen, which doesn't include every possible use case of macros. For instance, you can't declare a variable inside an enclosing scope by calling a template. You'd need a macro to accomplish that. Though that's something people often go to great lengths to avoid accidentally doing when writing macro. Adding a scope block to a macro is uglier than many people realize. See:
https://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block

As for the whole strawman thing, you did literally ask "what's wrong with macros?", and there are plenty of problems that come up when using them. And there's more that haven't been thought of here. For instance, the coming module system for C++ has complications due to macros.

----

In terms of alternatives for the unused variable warning, I generally see them in two places:

```cpp
void someFunction(int unused); // Warning
void someFunction(int /*unused*/); // No warning
```

```cpp
auto [unused, b] = someStructure; // Warning
auto b = someStructure.b; // No warning
```

OPHD/StructureManager.h
DanRStevens: I think you're correct about using [`static_cast`](https://en.cppreference.com/w/cpp/language/static_cast).

According to cppreference, this would be case (2). Now that I think about it, I rarely use `static_cast` outside of case (1). Whenever I see `static_cast`, I pretty much always think of case (1).

When dealing with derived classes, normally being able to call virtual functions means I don't need to worry about down casting. The only conversion I need to worry about is up casting, and that happens implicitly. The few cases where I have restored to down casting have generally been of the kind that requires `dynamic_cast`.

Then there's my habit of resorting to `reinterpret_cast` because of all the patching work I used to do. It has probably become my hammer.

Another interesting source of information was:
https://www.mooclab.club/threads/how-do-you-explain-the-differences-among-static_cast-reinterpret_cast-const_cast-and.4079/

----

As for what would happen if someone overloaded a type conversion method. Probably nothing special. The cast is on the pointer, not on the type.

It occurs to me that my usual use of `static_cast` with case (1) means I'm normally working on the class type, not on a pointer to the class type.

OPHD/Common.h
rmtoth: >The point is more that when there are alternatives that don't involve using macros, then alternative solutions should perhaps be used. Inline functions are a better alternative to function macros. Template functions are a better alternative to untyped macro functions that need to work with multiple types. Typed constants are a better alternative to untyped macro defined values.

>If your use is X-Macros, then yes, you do probably need to use macros. There's no real alternative to that offered by other language features.

As often is the case, I think our views are actually not that different now that we've elaborated a bit. I mistook your initial feedback for "you should never use macros," but re-reading that part I see you never wrote anything to that effect.

> As for the whole strawman thing, you did literally ask "what's wrong with macros?"

You're right - I should have asked "what's wrong with this macro?" :-) But let's not go into that discussion.

PR #840, opened 2021-03-30
Ecs staging

rmtoth:
Fix most review comments in PR #836 and PR ##839. Some notable points that I did not address:

`StructureManager` is still using raw `StructureComponent` pointers in its backing store. Using `unique_ptr` would be challenging because of the dependency between classes:
`StructureComponent` uses `StructureManager`, and `StructureManager` has `StructureComponent` pointers. `StructureManager` can have `StructureComponent` raw pointers since it's sufficient to forward-declare the type. `unique_ptr` on the other hand references the destructor and requires the type to be fully declared.

I kept the `getComponent` functions. Looking closer at my fork, I see that I actually do use them a fair bit, when porting code like this over to the ECL:
```
if (structure->isRobotCommand())
{
	deleteRobotsInRCC(robot, &GetComponent<RobotCommand>(structure), mRobotPool, mRobotList, &tile);
}
```

Longer term we should eliminate as much of such code as possible, and we may be able to remove the getters in the end. But we should keep them for now until they're no longer necessary at some point in the future.

I still have an `SKey` class. In an attempt to make it less ugly, I removed the getter and gave friend access to `StructureManager`, which is the only class that should be using the key anyway. Let me know if I should rather just erase the `SKey` class and use `Structure*` directly, I don't mind either way.

I considered adding `static_assert` to check that `ComponentTy` actually inherits `StructureComponent` in various places, but the use of `static_cast` makes the compiler already complain if you attempt to pass an unexpected type. If you want we can still add `static_assert` just to make the compiler's error message slightly more obvious.

### DIFF
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -366,8 +366,3 @@ const auto formatDiff = [](int diff)
 {
 	return ((diff > 0) ? "+" : "") + std::to_string(diff);
 };
-
-/**
- * Silence compiler warnings about unused variable
- */
-#define UNUSED(x) (void)(x)
--- a/OPHD/StructureComponent.h
+++ b/OPHD/StructureComponent.h
@@ -16,16 +16,6 @@ inline T& getComponent(SKey s)
 	return NAS2D::Utility<StructureManager>::get().get<T>(s);
 }
 
-/**
- * Return a reference to the Structure type belonging to a structure.
- * This allows writing code that's agnostic to the SKey type.
- */
-template<>
-inline Structure& getComponent<Structure>(SKey s)
-{
-	return *s.getInternal();
-}
-
 
 /**
  * Return a pointer to the given StructureComponent type belonging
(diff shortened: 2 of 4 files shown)


rmtoth:
I'm on the fence about having both `getComponent` and `tryGetComponent`. On one hand we could just remove the reference version and only keep the pointer version. On the other hand if you want to get an exception rather than a crash on programmer errors, that moves the responsibility of checking for nullptr into every single callsite rather than being performed in the reference getter. Opinions?

DanRStevens:
I generally prefer exceptions to sentinel values. Having to check for sentinel values makes calling code kind of ugly, which can discourage people from writing proper error checking code.

If errors are expected though, that may change a bit. It's much easier to write an `if` around a return value than a `try`/`catch` block around an exception. Exceptions are not a good way to do standard flow control. Admittedly though, there's a bit of gray area over what is considered "expected", and how exactly flow control is being done. An example of an expected problem might be checking if a user input field contains only digits, knowing that users can type any character. Standard flow control I find a little harder to define.

At any rate, I find it acceptable to offer both forms as two separate functions when both possibilities may be needed.

DanRStevens (approved):
Looks good.

PR #841, opened 2021-03-31
added component enumeration function

rmtoth:
This change adds a helper class suitable for range-based for-loops, as well as functions to obtain instances of the helper class from the structure manager.

Intended usage pattern is:
```
for ( auto& componentInstance : enumerateComponents<SomeComponentType>()) {
  componentInstance.someFunction();
}
```

### DIFF
--- a/OPHD/StructureManager.h
+++ b/OPHD/StructureManager.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <NAS2D/Utility.h>
 #include "Things/Structures/Structure.h"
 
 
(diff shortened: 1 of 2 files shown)


rmtoth:
I think the global convenience functions `getComponent`, `tryGetComponent` and `enumerateComponent` should be moved from StructureComponent.h into StructureManager.h, since they're more closely related to the `StructureManager` functions they're giving convenience access to than to the `StructureComponent` class itself. Thoughts?

OPHD/StructureManager.h
```diff
@@ -37,6 +37,38 @@ class SKey
...
+	class Iterator
+	{
+	private:
+		std::map<SKey, StructureComponent*>::iterator mIt;
+	public:
+		Iterator(std::map<SKey, StructureComponent*>::iterator it) : mIt(it) {}
+		bool operator!= (const Iterator& rhs) const { return mIt != rhs.mIt; }
+		Iterator& operator++() { ++mIt; return *this; }
+		operator ComponentTy* () const { return static_cast<ComponentTy*>(mIt->second); }
+		ComponentTy* operator->() const { return static_cast<ComponentTy*>(mIt->second); }
```
DanRStevens, line 58: The spacing is a bit inconsistent here before the opening parenthesis for the parameter list. I would recommend no spacing.

OPHD/StructureManager.h
```diff
@@ -37,6 +37,38 @@ class SKey
 };
 
 
+/**
+ * Object suitable for range-based for loops.
+ * This class allows code to iterate over instances of a given component
+ * without exposing the internal component storage container type.
+ */
+template<typename ComponentTy>
+class ComponentRange
```
DanRStevens, line 46: There are now 3 top level classes in one file. Might be good to either break them out into separate files, or make them inner classes.

DanRStevens (approved):
Code looks good. Made some minor notes.

----

I would agree those functions should be moved to `StructureManager`. They all depend on having a `StructureManager` object to operate, so yes, it makes sense for them to become member functions. In particular, they all start with:
```cpp
NAS2D::Utility<StructureManager>::get()
```

rmtoth:
I wasn't actually referring to making the convenience functions static members, since my main intent with having them was to reduce verbosity. I only meant moving the global functions to the other file. But perhaps the `StructureManager::` prefix isn't too annoying. I'll leave it outside for now, but feel free to move it if you have a strong opinion about it.

DanRStevens:
Ahh, I see now.

I was actually expecting those functions to become regular (non-static) members, as they require an instance of `StructureManager` to operate on. But there are already member functions for that functionality. You're just delegating to the existing member functions and giving a global method the wraps lookup of the global `StructureManager` variable. I had missed that point when I first looked at it.

Anyway, change looks fine. The code is probably better housed in the new file.

PR #845, opened 2021-04-06
Throw in release

DanRStevens:
As discussed in https://github.com/OutpostUniverse/OPHD/pull/836#discussion_r603837928 there are a few places where the code was only throwing in debug mode. This removes the guards so the throws can happen in any configuration.

### DIFF
--- a/OPHD/Things/Structures/Structure.cpp
+++ b/OPHD/Things/Structures/Structure.cpp
@@ -310,8 +310,5 @@ void Structure::die()
 {
 	Thing::die();
 
-	#if defined(_DEBUG)
-	std::cout << "Holy shit, a Structure died!!!" << std::endl;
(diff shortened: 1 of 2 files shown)


OPHD/Things/Structures/Structure.cpp
```diff
@@ -310,8 +310,5 @@ void Structure::die()
 {
 	Thing::die();
 
-	#if defined(_DEBUG)
-	std::cout << "Holy shit, a Structure died!!!" << std::endl;
```
ldicker83, line 314: Always the professional huh? :D

ldicker83:
This looks good to me.

OPHD/Things/Structures/Structure.cpp
DanRStevens: Well, I did leave in "He's dead Jim!" :wink:

PR #846, opened 2021-04-06
let each structure component add information to the structure inspector window

rmtoth:


rmtoth:
This will currently add information in ascending ComponentTypeID order. We can figure out how to best define desired ordering later.

OPHD/UI/StructureInspector.cpp
```diff
@@ -128,6 +130,19 @@ void StructureInspector::drawStructureSpecificTable(NAS2D::Point<int> position,
 	stringTable.computeRelativeCellPositions();
 	stringTable.position(position);
 	stringTable.draw(renderer);
+	position.y = stringTable.screenRect().endPoint().y + 25;
+
+	for (auto& component : NAS2D::Utility<StructureManager>::get().structureComponents(mStructure))
+	{
+		StringTable componentStringTable = component.second->createInspectorViewTable();
```
DanRStevens, line 137: A `struct` with named fields rather than a `std::pair` would make for more readable code here. At present, it's not super obvious that `component.second` refers to a `StructureComponent` object. Named fields would make this distinction more clear.

Naming also contributes a bit to the confusion. The loop variable is named `component`, but it's not actually a `StructureComponent`. Even the name method name `structureComponents` would seem to imply a collection of `StructureComponent` objects, rather than a collection of `std::pair`, where one field is a `StructureComponent`.

DanRStevens (approved):
Code looks good.

Had one comment on code readability, though it perhaps pertains more to previously merged support code.

OPHD/UI/StructureInspector.cpp
rmtoth: Do you have suggestions on better names? My imagination isn't cooperating with me right now.

Another option would be having `structureComponents` actually returning a vector of `StructureComponent*` rather than of pair/struct, and add virtual getters to `StructureComponent` to obtain additional information if/when needed. Not sure what would be cleanest. I suggest postponing this decision until we see what the various use cases look like.

DanRStevens: Yeah, might be better to wait until use cases are more clear. If you went with a `std::vector<StructureComponent*>`, that would eliminate the name concerns.

PR #847, opened 2021-04-07
Fix release configuration log file redirect

DanRStevens:
Apply crash fixes from OP2-Landlord to OPHD for release configuration log redirect. Old code did not restore the old `rdbuf` before exiting, which caused crashes when exiting.

To see the crash on Linux with the old code, compile with:
```sh
make CXXFLAGS_EXTRA=-DNDEBUG
```

As the Linux environment doesn't normally define `NDEBUG` (MSVC specific define), this was not normally seen for Linux builds.

This was noticed while reviewing pre-processor directives for #845.

### DIFF
--- a/OPHD/main.cpp
+++ b/OPHD/main.cpp
@@ -29,14 +29,13 @@ using namespace NAS2D;
 
 int main(int argc, char *argv[])
 {
-	//Crude way of redirecting stream buffer when building in release (no console)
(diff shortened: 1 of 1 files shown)


ldicker83:
Huh. I wonder if this was that heap assert I was getting during exit? Would be great to not see that anymore. :D

Anyway, looks good to me.

Issue #843, opened 2021-04-05
Structure construction and serialization in the entityComponentSystem branch


rmtoth:
Opening a new issue to track creation and serialization aspects of the entity component system (ECS).
I think it might be best to wait with refactoring these aspects of the engine until the ECS can be merged into master, built on top of the existing creation/serialization code structure, to avoid straying too far in the feature branch.

The existing code for creation/serialization/deserialization is a bit spread out over various places in the code, but I think it can roughly be summized as follows:
* `StructureCatalog` is a database of what kinds of structures exist, and has the responsibility of serving the construction GUI as well as instanciating the various `Structure` subclasses.
* `StructureManager` is a tracking class for all the structures that exist on the map. It is responsible for destroying structures. It is also responsible for serializing structures.
* `MapViewStateIO` is responsible for deserializing structures.

I think the responsibilities should be slightly shifted around as follows:

* `StructureCatalog` remains responsible for serving the construction GUI. It maintains a database of _structure types_ that can be constructed. A structure type defines fixed and initial parameters of a structure, as well as the set of structure components attached to it and fixed/initial parameters of the components.
* `StructureManager` remains responsible for tracking all the structures that exist on the map. It should be responsible for user-initiated creation of new structure and component instances based on structure type descriptions obtained from `StructureCatalog`. It should be responsible for destroying structures and their components, thereby owning the entire lifetime of these object types. It should be responsible both for serializing and deserializing structures and their components.
* `MapViewStateIO` is responsible for loading/saving the entire game state, but delegates serialization/deserialization of structures to `StructureManager`.

I imagine a structure type definition to be _data_ rather than code. As data, it can be loaded from an XML file or similar, and allows for easy modding without building the game from source.
```
<catalog version="1">
	<structureType id="Agridome" sprite="agridome.sprite" maxAge="600" turnsToBuild="3">
		<!-- user-facing information -->
		<desc locale="en" name="Agricultural Dome">Textual description of structure for use in tooltips and structure inspector.</desc>
		<desc locale="sv" name="Jordbrukskupol">Textbeskrivning av byggnad för användning i verktygstips och inspektionsrutor.</desc>
		<!-- construction requirements -->
		<tech id="Xeno nourishment"/>
		<cost resources="20,10,5,0"/>
		<!-- operating requirements -->
		<requireComponent id="CHAP" amount="5"/>
		<consume resources="1,0,0,0" energy="2" workers="5"/>
		<!-- structure services -->
[...]

PR #849, opened 2021-04-08
Component food production

rmtoth:
Refactored food production from being an abstract Structure subclass to being an entity component.
This is a larger change, so I'll try to add review comments explaining the various changes.


OPHD/States/MapViewStateDraw.cpp
```diff
@@ -170,11 +171,16 @@ void MapViewState::drawResourceInfo()
 	}
 
 	// Capacity (Storage, Food, Energy)
+	int foodStorage = 0;
+	for (auto& foodProduction : enumerateComponent<FoodProduction>())
+	{
+		foodStorage += foodProduction.foodCapacity();
+	}
```
rmtoth, line 178: The prior method of counting food storage capacity did not take Command Center capacity into account. Now actually counting the capacity, so that every food production structure can have an arbitrary capacity.

OPHD/States/MapViewStateEvent.cpp
```diff
@@ -111,9 +111,10 @@ void MapViewState::deployColonistLander()
  */
 void MapViewState::deployCargoLander()
 {
-	auto cc = static_cast<CommandCenter*>(mTileMap->getTile(ccLocation(), 0).structure());
-	cc->foodLevel(cc->foodLevel() + 125);
-	cc->storage() += StorableResources{ 25, 25, 15, 15 };
+	Structure* structure = mTileMap->getTile(ccLocation(), 0).structure();
+	auto& foodProduction = getComponent<FoodProduction>(structure);
+	foodProduction.foodLevel(foodProduction.foodLevel() + 125);
+	structure->storage() += StorableResources{ 25, 25, 15, 15 };
```
rmtoth, line 117: This is just a mechanical translation, but in general I feel that the deployXxLander functionality could be made more elegant (later, perhaps when converting the landers to components).

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -89,15 +89,10 @@ void MapViewState::updatePopulation()
...
 	int hospitals = structureManager.getCountInState(Structure::StructureClass::MedicalCenter, StructureState::Operational);
 
-	auto foodproducers = structureManager.structureList(Structure::StructureClass::FoodProduction);
-	auto& command = structureManager.structureList(Structure::StructureClass::Command);
-	foodproducers.insert(foodproducers.end(), command.begin(), command.end());
-
 	int remainder = mPopulation.update(mCurrentMorale, mFood, residences, universities, nurseries, hospitals);
 
-	for (auto structure : foodproducers)
+	for (auto& foodProducer : enumerateComponent<FoodProduction>())
```
rmtoth, line 94: I ditched the "consume from Command Centers last" behavior as it didn't fit in well with the notion that all food producing components are the same. In practice the difference will be minor since `MapViewState::transferFoodToCommandCenter` will make sure the CCs are near-full anyway (full minus up to one turn's consumption).

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -412,21 +407,22 @@ void MapViewState::updateBiowasteRecycling()
 }
 
 
-void MapViewState::countFood()
+void MapViewState::produceFood()
 {
-	mFood = 0;
+	for (auto& foodProducer : enumerateComponent<FoodProduction>())
+	{
+		foodProducer.produce();
```
rmtoth, line 414: Food was earlier produced in `Agridome::think`. It is now being performed in a separate function. This could arguably be a static function in `FoodProduction`, along with global food capacity/level counting, but adding the functionality here seemed like a smaller change to the overall code structure.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -412,21 +407,22 @@ void MapViewState::updateBiowasteRecycling()
...
+	mFood = 0;
 
-	for (auto structure : structures)
+	for (auto& foodProducer : enumerateComponent<FoodProduction>())
 	{
-		if (structure->operational() || structure->isIdle())
-		{
-			mFood += static_cast<FoodProduction*>(structure)->foodLevel();
-		}
+		mFood += foodProducer.foodLevel();
```
rmtoth, line 425: the operational/idle checks were superfluous, since the level was set to zero if the structure is disabled anyway.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -509,11 +505,13 @@ void MapViewState::nextTurn()
 
 	mPreviousMorale = mCurrentMorale;
 
+	produceFood();
```
rmtoth, line 508: this was done slightly earlier before, in `StructureManager::update` -> `StructureManager::updateStructures` -> `Agridome::think`. Food levels are not used between here and the prior production point.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -509,11 +505,13 @@ void MapViewState::nextTurn()
...
 	transferFoodToCommandCenter();
 	updateResidentialCapacity();
 
 	countFood();
 	updatePopulation();
+	countFood();
```
rmtoth, line 514: `updatePopulation` reduces the food amount, so re-counting here to make the GUI show the correct level.

OPHD/StructureCatalogue.cpp
```diff
@@ -29,11 +30,32 @@ Structure* StructureCatalogue::get(StructureTypeID type)
...
+			structure->turnsToBuild(3);
+			structure->requiresCHAP(true);
+			structure->resourcesIn({ 1, 0, 0, 0 });
+			structure->energyRequired(2);
+
+			const int AGRIDOME_CAPACITY = 1000;
+			const int AGRIDOME_BASE_PRODUCUCTION = 10;
+			FoodProduction* foodProduction = new FoodProduction(structure);
+			foodProduction->initialize(AGRIDOME_BASE_PRODUCUCTION, AGRIDOME_CAPACITY);
+			NAS2D::Utility<StructureManager>::get().attachComponent(structure, foodProduction);
```
rmtoth, line 56: this is quite ugly, but temporary. The initialize call will be replaced by structure type deserialization as discussed in #843 . I also dislike calling `StructureManager::attachComponent` on a `Structure*` that has not yet been added with `StructureManager::addStructure`, but this is the only place I have the actual component type. Again, this won't be an issue once #843 is addressed.

OPHD/Things/Structures/FoodProduction.h
```diff
@@ -22,18 +28,34 @@ class FoodProduction : public Structure
...
+		{
+			foodLevel(foodLevel() + mFoodProductionRate);
+			if (mFoodProductionRate && mFoodLevel == mFoodCapacity)
+			{
+				structure().idle(IdleReason::InternalStorageFull);
+			}
+		}
+		else if (structure().disabled())
+		{
+			foodLevel(0);
```
rmtoth, line 53: this was done in `Agridome::disabledStateSet` before. Now being cleared in the food-growing phase of the new turn.

OPHD/StructureCatalogue.cpp
rmtoth: I considered putting this code in a temporary `CreateAgridome` function to avoid the temporary clutter in `StructureCatalogue::get` but a bunch of the `Structure` members are protected. It seemed less offensive to put this here than to make the members public or giving friend access to a bunch of structure factory functions.

OPHD/States/MapViewStateEvent.cpp
ldicker83: It could be -- the deploy functionality was sort of hacked together to 'just get it to work' so I could move on to other things.

This code looks a lot cleaner too, I like this.

OPHD/States/MapViewStateIO.cpp
```diff
@@ -408,7 +408,7 @@ void MapViewState::readStructures(Xml::XmlElement* element)
 		if (structureId == StructureTypeID::SID_AGRIDOME ||
 			structureId == StructureTypeID::SID_COMMAND_CENTER)
 		{
-			auto& foodProduction = *static_cast<FoodProduction*>(&structure);
```
ldicker83, line 411: I like seeing this in code and seeing the improvement -- this is so much better than all the casting.

Question though -- since it appears to return a reference, what if this is called on a structure that doesn't have a `FoodProduction` component? Does it throw, barf or something else?

OPHD/States/MapViewStateTurn.cpp
ldicker83: Fair enough. I find this acceptable.

OPHD/States/MapViewStateTurn.cpp
ldicker83: This is an acceptable change. It makes more sense for a food production facility to `produce()` than to `think()` food into existance.

OPHD/States/MapViewStateTurn.cpp
ldicker83: Good catch.

OPHD/StructureCatalogue.cpp
ldicker83: This is what I'd call 'falsework' (from construction lingo) and is acceptable. It makes sense to me to have something temporarily in place until it can be replaced with something better. It's the nature of iterative work I think.

OPHD/Things/Structures/CommandCenter.h
```diff
@@ -33,3 +34,4 @@ class CommandCenter: public FoodProduction
 		return 0;
 	}
 };
+#endif
```
ldicker83: I take it this is in preparation of moving toward a data based approach? E.g., pulling the hard coding out and putting definitions into some external file?

OPHD/States/MapViewStateTurn.cpp
rmtoth: the alterantive would be to add a `think()` method to `StructureComponent` and invoke that as part of the structure update loop. I think that could be more difficult to get right with respect to ordering various phases though, so I think explicit logic like this might be better.

ldicker83:
Okay, so all-in-all this looks pretty good. I had a couple of comments but nothing blocking merging so I'll go ahead and merge this into the ECS branch.

@DanRStevens, if you have any comments by all means. I'm pretty satisfied with this though.

OPHD/States/MapViewStateIO.cpp
rmtoth: In this case we are confident there will be a `FoodProduction` component, and deem it a programming error if that's not true. `getComponent` will throw if the component does not exist. If you have similar logic where you aren't sure whether the structure will have the component or not, you can use the `tryGetComponent` function instead, which returns a pointer that can be `nullptr` if the structure didn't have the queried component.

OPHD/Things/Structures/CommandCenter.h
rmtoth: Yes. This should be cleaned up when tackling issue #843.

OPHD/States/MapViewStateIO.cpp
ldicker83: Nice, I like that.

DanRStevens:
Changes look good to me. Sorry it took me a little while to get around to reviewing things.

I'd don't quite yet follow why the ugly deserialization code is there, though I assume that will become obvious when that part is updated. I'm fine with temporarily ugly code.

----

In terms of the PR comments, those might be better placed as git commit messages. Each of those comments seemed to pertain to a particular small change set, so it may have made sense to make the small section changes their own commit, and comment them with the git commit message.

As a bonus, if we ever need to move hosts, it's quite easy to push a git repository up to a new provider, which would preserve all git commit messages. The PR comments however would not be preserved by such a move.

My personal preference is for fairly small granular commits.

rmtoth:
>Sorry it took me a little while to get around to reviewing things.

Nothing to worry about :-)

>I'd don't quite yet follow why the ugly deserialization code is there, though I assume that will become obvious when that part is updated.

Are you referring to `StructureCatalog::get`? It's mainly there because we no longer have per-SID structure subclasses doing that in the c'tor. All of that should disappear when we tackle #843 .

>comment them with the git commit message

Sure, I can be more verbose in the commit messages. I usually just type a one-liner with -m, but for these larger changes I agree that they may arrant more context.

>My personal preference is for fairly small granular commits.

I know the commit was larger than you'd prefer, but it would have been difficult/tedious/time-consuming to split into several commits that would each compile and run. As a heads up, I intend to keep converting one structure class at a time, and those changes will likely be of a similar nature. Some structure classes will be smaller changes, but others might become bigger. Some may have obvious ways to break into several functional commits, but some of them won't. I hope familiarity with prior commits make the patterns in subsequent similar commits easier to recognize and review.

DanRStevens:
Yes, I figured out the `StructureCataglog::get` thing when reviewing the next change set. Makes sense now.

I think your PR messages are actually a pretty good indication of where certain changes could have been broken up into separate commits.

PR #850, opened 2021-04-10
Converted PowerStructure to structure component. 

rmtoth:
Mean solar distance is routed through update functions rather than through structure catalogue. Solar distance is not accessible when populating the inspector, so using inaccurate value for now.

Is there any way of accessing planetary attributes from arbitrary locations in the code? I couldn't find anything obvious. I removed the mean solar distance from `StructureCatalogue` which seems like the wrong place to cache it. I'm piping it through `StructureManager::update` for now, but it should probably be queried from global planetary attributes in `PowerStructure::productionRate` instead, which would also fix the inspector view (see comment in code).


DanRStevens (approved):
Looks good.

For `StructureCataclogue::get`, I just realized that extra code is going there because the associated `Structure` derived classes are being removed, and that data was previously set by their constructors. That makes more sense now.

I assume that data will eventually be loaded externally.

Meanwhile, I wonder if the code might look a bit nicer if the `Structure` constructor could take arguments for the `maxAge`, `turnsToBuild`, and `requiresCHAP` values.

ldicker83:
Mean solar distance -- I would imagine that's something we'd want accessible to anything also involved with the `MapViewState `-- maybe we should use a global accessor kind of like is already done with the `StructureManager`? e.g., `NAS2D::Utility<StructureManager>`.

PR #879, opened 2021-05-13
Add RadioButtonGroup as separate Control

belgianguy:
This introduces a separate RadioButtonGroup Control that simplifies UIContainer logic.


OPHD/UI/Core/RadioButton.cpp
```diff
@@ -1,30 +1,23 @@
...
-
-#include <algorithm>
-
+#include "RadioButtonGroup.h"
 
 using namespace NAS2D;
 
-
-RadioButton::RadioButton(std::string newText) :
+RadioButtonGroup::RadioButton::RadioButton(std::string newText, RadioButtonGroup* parentContainer, NAS2D::DelegateX<void> delegate) :
```
DanRStevens, line 5: I would probably try to group the `newText` and `delegate` parameters next to each other, as they are kind of interrelated. The `parentContainer` parameter is more standalone.

OPHD/UI/Core/RadioButton.cpp
```diff
@@ -1,30 +1,23 @@
...
 	mFont{fontCache.load(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL)},
-	mSkin{imageCache.load("ui/skin/checkbox.png")},
-	mLabel{newText}
+	mSkin{imageCache.load("ui/skin/radio.png")},
+	mLabel{newText},
+	mParentContainer{parentContainer}
 {
-	Utility<EventHandler>::get().mouseButtonDown().connect(this, &RadioButton::onMouseDown);
+	hasFocus(true);
```
DanRStevens, line 11: This `hasFocus` call stands out as looking a bit strange and unexpected. Should this perhaps be handled by the `RadioButtonGroup` when an `index` is selected?

OPHD/UI/Core/RadioButton.cpp
```diff
@@ -33,60 +26,57 @@ RadioButton::~RadioButton()
...
 {
-	for (auto* sibling : mParentContainer->controls())
-	{
-		if (auto* asRadioButton = dynamic_cast<RadioButton*>(sibling))
-		{
-			asRadioButton->checked(false);
-		}
-	}
+	mParentContainer->clearSelection();
+	mRbgDelegate();
```
DanRStevens, line 60: I kind of feel like `clearSelection` should be handled internally in the `mRbgDelegate()` callback method. That way `clearSelection` could potentially be made `private` or `protected`.

----

This is the wrong function to put this code in. This method is returning the `SignalSource` object. It is not the `onClick` handler method.

OPHD/UI/Core/RadioButton.cpp
```diff
@@ -33,60 +26,57 @@ RadioButton::~RadioButton()
...
+
 	checked(true);
-	return mSignal;
+	return mCallback;
```
DanRStevens, line 63: Same, wrong method. This should go in the `onClick` handler method.

OPHD/UI/Core/RadioButtonGroup.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+void RadioButtonGroup::update()
+{
+	if (!visible()) { return; }
+
+	if( mRadioButtons.size() > 0 && !hasOneSelectedItem() ) {
+		clearSelection();
+		mRadioButtons[0].click();
+	}
```
DanRStevens, line 47: I feel like this part could potentially be improved.

I feel like the `RadioButtonGroup` class should maintain an invariant of having at most one button selected. (No buttons selected should be fine too). The exposed methods should make sure this is always true, and so there would be no need to check this in the `update()` method.

OPHD/UI/Core/RadioButtonGroup.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+bool RadioButtonGroup::hasOneSelectedItem()
+{
+	int count = 0;
+	for (auto &control : mRadioButtons)
+	{
+		if(control.checked()) { count++; }
+	}
+	return count == 1;
+}
```
DanRStevens, line 73: I would really like for this method to not be needed.

OPHD/UI/Core/RadioButtonGroup.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void RadioButtonGroup::clearSelection()
+{
+	if (!visible()) { return; }
+
+	for (auto &control : mRadioButtons)
+	{
+		control.checked(false);
+	}
+}
```
DanRStevens, line 63: I'm not sure this method need the `visible()` guard. It seems like you should be able to clear a `RadioButtonGroup` selection, even when it is not currently visible.

OPHD/UI/Core/RadioButtonGroup.cpp
```diff
@@ -0,0 +1,78 @@
...
+	{
+		if(control.checked()) { count++; }
+	}
+	return count == 1;
+}
+
+void RadioButtonGroup::setSelected(std::size_t index)
+{
+	mRadioButtons[index].checked(true);
+}
```
DanRStevens, line 78: I think this method can be updated to remove the need for the `hasOneSelectedItem`. Perhaps by clearing the selection of the old index before setting the new index. The `RadioButtonGroup` would need to track the index of the currently selected item (if any).

```cpp
if (mIndex != NoItemSelected)
{
    mRadioButtons[mIndex].checked(false);
}
mIndex = index;
mRadioButtons[index].checked(true);
```

OPHD/UI/Core/RadioButtonGroup.h
```diff
@@ -0,0 +1,67 @@
...
+        void parentContainer(RadioButtonGroup* parent);
+
+    private:
+        const NAS2D::Font& mFont;
+        const NAS2D::Image& mSkin;
+        Label mLabel;
+        ClickCallback mCallback; /**< Object to notify when the Button is activated. */
+        RadioButtonGroup* mParentContainer{nullptr};
+        bool mChecked{false};
+        void onMouseDown(NAS2D::EventHandler::MouseButton button, int x, int y);
```
DanRStevens, line 62: This method should be declared along side the other event handler methods, rather than mixed in with the variables.

This method is probably going to handle the `onClick` event. Probably doesn't matter if it's one event handler, or split into two. Might provide some potentially subclassing benefits if split into two.

DanRStevens (commented):
Wow, that cleaned up the history quite nicely. Thank you for handling that. Definitely much easier to read and review with the squashed history.

I noted a few places to look at for potential improvement. Some of them correspond with previous comments from the issue thread. I tried giving a bit of detail to suggestion how they might be improved. With the cleanup, I think I now have a better sense of how I expect the API to function.

OPHD/UI/Core/RadioButton.cpp
belgianguy: Moved parentContainer to the front.

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: Agreed, removed the `visible()` guard.

OPHD/UI/Core/RadioButton.cpp
belgianguy: I tried adding it in the setSelected method, but then the clicks don't register.
I found that placing it in `update()` also works. But I don't know if that is any better.

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: Okay, I have removed it locally for now, still testing if this assertion can be made to hold without it.

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: Added the `std::size_t mIndex` member and initializing it to `constants::NO_SELECTION` and replaced the body by the code above.

OPHD/UI/Core/RadioButtonGroup.h
belgianguy: Moved it into the protected section where the other event handler methods.

OPHD/UI/Core/RadioButton.cpp
belgianguy: ok, I don't have an `onClick()` method at this time, but I do have an `onMouseDown()` that now is used to execute the action.

OPHD/UI/Core/RadioButton.cpp
belgianguy: Ok, I'm now treating the method that returns the SignalSource just as a getter for the SignalSource (meaning removing any state altering logic and only return the object. Is that the correct interpretation?

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: Removed in latest push.

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: Changed the implementation accordingly.

OPHD/UI/Core/RadioButton.cpp
belgianguy: I think there is still some confusion here, I noticed the mCallback (which is different from the mRbgDelegate) was in fact not used, I've removed mCallback and its getter/setter to make the situation clearer.

Member `mRbgDelegate` contains the function that is executed in the targeted UI Control (here: FactoryProduction window `chkIdleClicked`).

Although as an inner class of RadioButtonGroup I don't really see the problem with `mParentContainer` calling `clearSelection()`.

belgianguy:
Hi @DanRStevens , I've pushed a new version, can you check if the topics above were handled correctly? I've added a comment where possible.

OPHD/UI/Core/RadioButton.cpp
DanRStevens: What is the purpose of the `hasFocus` call? I'm thinking it allows a `Control` to be highlighted so the user knows where input will go. I see tracking the active control as being relevant only to keyboard input. I don't know why it would be called on every `RadioButton`. I'm not super familiar with how it's used in the API though.

OPHD/UI/Core/RadioButton.cpp
DanRStevens: Sounds like `mRbgDelegate` is being used for what I assumed `mSignal` was being used for. Might be a good idea to rename the field. Having "RBG" in the name makes me assume the callback is from the `RadioButton` to the `RadioButtonGroup` object, so it can implement one-hot. I expected `mSignal` to be for the final user method that's called when a corresponding option is selected.

Looks like there is no longer any accessor method for the `SignalSource` interface.

OPHD/UI/Core/RadioButtonGroup.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+void RadioButtonGroup::clearSelection()
+{
+	for (auto &control : mRadioButtons)
+	{
+		control.checked(false);
+	}
+	mIndex = constants::NO_SELECTION;
+}
```
DanRStevens, line 58: This can probably be implemented without a loop. Assuming the at-most-one invariant is maintained, you could borrow the unselect code from the `setSelected` method. Actually, the `setSelected` method could delegate to `clearSelection`, so you only need the code in one place.

OPHD/UI/Core/RadioButtonGroup.h
```diff
@@ -0,0 +1,64 @@
...
+		std::string name;
+		NAS2D::DelegateX<void> delegate;
+	};
+
+	RadioButtonGroup() = default;
+	RadioButtonGroup(std::vector<ButtonInfo> buttonInfos);
+	void setSelected(std::size_t index);
+	void onMove(NAS2D::Vector<int> displacement) override;
+	void update() override;
+	void clearSelection();
```
DanRStevens, line 33: Maybe group the `clearSelection` and `setSelected` methods together.

----

We should probably be consistent regarding "selection" versus "selected".

OPHD/UI/Core/RadioButtonGroup.h
```diff
@@ -0,0 +1,64 @@
...
+        bool checked() const;
+        void text(const std::string& text);
+        const std::string& text() const;
+        void update() override;
+
+    protected:
+        void onResize() override;
+        void onTextChange() override;
+        void parentContainer(RadioButtonGroup* parent);
+        void onMouseDown(NAS2D::EventHandler::MouseButton button, int x, int y);
```
DanRStevens, line 51: We should probably move `parentContainer` out from the middle of the event handler methods. Actually, do we even need a `parentContainer` setter method anymore? Seems like that could/should be set once in the constructor and then never changed.

OPHD/UI/Core/RadioButtonGroup.h
```diff
@@ -0,0 +1,64 @@
...
+
+    private:
+        const NAS2D::Font& mFont;
+        const NAS2D::Image& mSkin;
+        Label mLabel;
+        RadioButtonGroup* mParentContainer{nullptr};
+        bool mChecked{false};
+        NAS2D::DelegateX<void> mRbgDelegate;
```
DanRStevens, line 59: We should probably revisit `DelegateX` versus `Delegate` at some point. I'm still not clear on why we're using `DelegateX` in this one class.

DanRStevens (commented):
Ok, I'm rather liking how this is shaping up.

Added a few more comments for things I just noticed.

OPHD/UI/Core/RadioButton.cpp
belgianguy: I have looked into where the hasFocus() call is being used to apply said information, and I see there is a total of 2 checks in  `Window.cpp`:

```
void Window::onMouseDown(EventHandler::MouseButton button, int x, int y)
{
	if (!visible() || !hasFocus()) { return; }

	UIContainer::onMouseDown(button, x, y);

	const auto titleBarBounds = NAS2D::Rectangle{mRect.x, mRect.y, mRect.width, sWindowTitleBarHeight};
	mMouseDrag = (button == EventHandler::MouseButton::Left && titleBarBounds.contains(NAS2D::Point{x, y}));
}

```

Could it be that this is blocking mouse clicks when its focus is set to `false`?

image

OPHD/UI/Core/RadioButtonGroup.h
ldicker83: I'm also of the opinion of not adding set/get to function names. E.g., _selection(std::size_t)_ ought to be fine.

That stated, I am aware that my brief naming convention like this has caused some ambiguity so this is open for debate. We should really come up with a standard for this and make it consistent across the code base (and probably hard define it in a developer doc).

ldicker83:
Want to give a bit of a poke here. This looks to be almost entirely finished but don't want to merge as is with open/unresolved comments.

OPHD/UI/Core/RadioButton.cpp
DanRStevens: Ahh, yes, good catch. That is very strange to be checking in a `onMouseDown` event handler. Typically "focus" is only related to keyboard input, not mouse input.

I would recommend removing the `!hasFocus()` check from `onMouseDown`.

@ldicker83 : Did you want to weight in on that? You might be more familiar with why that check was added in the first place.

OPHD/UI/Core/RadioButtonGroup.h
DanRStevens: I was just thinking the `setSelected` method is an active method that changes state. Normally functions have verb names. What about naming it `select`?

Similarly, I would be ok with naming `clearSelection` as just `clear`. That's maybe a little more ambiguous, though in the context of a `RadioButtonGroup`, I imagine most people would be able to guess correctly.

----

Of course my above suggestion is perhaps getting away from the property naming syntax, which I think is what Leeor was originally going for. That is, name the property as a noun or adjective, and have both get and set forms of it, with a single name. I find that style more common in languages with property syntax support. In that case, reading or writing to a property is transparently redirected to function calls, without the code looking like a function call. I've seen it in VB, and I believe C# had similar support. I'm not so used to seeing it in C++. I wouldn't object to that style, just noting there is less syntax support for properties in C++.

DanRStevens:
Yeah, it's pretty close, though there are a few minor things left to look at. I think maybe just give him a bit of time to get around to it. I suspect he's somewhat time limited in when he can work on that stuff.

OPHD/UI/Core/RadioButton.cpp
DanRStevens: I tried checking `git blame` to see the history of that line. Though it seems it was just copied from other files, and use of `hasFocus` in mouse event handlers is quite widespread in the code base. I opened #884 to track.

I suppose we can mark this as "resolved", and worry about it as part of the other issue. It would make sense to address that issue for all controls simultaneously.

ldicker83:
Totally fair, no pressure!

OPHD/UI/Core/RadioButton.cpp
belgianguy: okay, renamed field mRbgDelegate to mSignal.

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: There's a problem here, the `index` parameter is actually only used when building the control, it is not used when other interactions are done. Meaning there is no mapping from `int` to a valid RadioButton in the vector.

Naively you could pass a RadioButton `this` pointer to the RadioButtonGroup and have a lookup table of some sort, but that might be more complicated than needed.

DanRStevens: The `mIndex` field already tracks the currently selected index. Here I'm assuming it's always set correctly. Is there some doubt about this? The code in `setSelected` seems to assume it is set correct.

Code to borrow from `setSelected` to implement `clearSelection`, with the last line updated:
```cpp
if (mIndex != constants::NO_SELECTION)
{
    mRadioButtons[mIndex].checked(false);
}
mIndex = constants::NO_SELECTION;
```

----

I suppose this relates to why I expected there to be two separate callbacks when a button is selected. One for the button specific interaction, and another for the `RadioButtonGroup` to ensure at-most-one-hot. This also relates to why I expected `clearSelection` to be called from an event handler on the `RadioButtonGroup`, rather than through the button with `mParentContainer->clearSelection();`.

If you remember, I had expected the `mRbgDelegate()` to call the at-most-one-hot event handler on the `RadioButtonGroup`:
```cpp
mParentContainer->clearSelection();
mRbgDelegate();
```

If that had been the case, it would have made sense to call `clearSelection()` from within the event handler on `RadioButtonGroup`, rather than just before the event handler was called. Though it seems `mRbgDelegate` really ended up being `mSignal`, which was a different concept entirely.

----

Anyway, my thought is, when a new button is selected the `RadioButtonGroup` somehow knows this, probably by calling an event handler method. When you clear a selection, the `RadioButtonGroup` will know what the previously selected button was, and can unselect it.

Looking back at the code now, it seems there is currently no code to inform the `RadioButtonGroup` of the update. Maybe we do need something like an `mRgbOneHot` delegate (similar to what I thought `mRgbDelegate` was supposed to be).

As you've mentioned, we can't just call `mParentContainer->setSelected` since the child control doesn't know it's own index. We could substitute with the `RadioButton` control address (`this`). Since the `RadioButton` controls are all in the same `std::vector`, it's is possible to convert an element address to an index in constant time (no loops).

Reference:
https://www.geeksforgeeks.org/computing-index-using-pointers-returned-stl-functions-c/

DanRStevens: Actually, I was just thinking. We don't really need a fully general `Delegate` system for the at-most-one-hot method. We know for sure the target control is `RadioButtonGroup`, and what the exact method is, and the `RadioButton` control is already tied to `RadioButtonGroup` in terms of design. Since it already has a `mParentContainer` reference, it can just call the event handler directly. Could potentially even make the event handler `private` or `protected`, and give `friend` access to the `RadioButton` control. That would lock down access to the event handler, and prevent abuse from outside code.

Or perhaps instead of an event handler style method, we could have an overload of `setSelected`, which takes a `RadioButton` control address instead of an index.

```cpp
mParentContainer->setSelected(this);
```

The method could use the technique from the article linked above to convert the address to an index (basically subtract the control address from the array base address), and then delegate to the index overload of `setSelected`.

belgianguy: Hi @DanRStevens I checked out the link provided, but that seems to detail the std::max_elements function, whereas we probably need something like this which can find a "random" entry in a vector based on its element pointer, https://thispointer.com/c-how-to-find-an-element-in-vector-and-get-its-index/ using std::find and std::distance (but I assume the find loops over the container). Would the std::find and std::distance combination be okay to use here to go from the pointer of the RadioButton to its index?

DanRStevens: I just doubled checked the link I sent you earlier. The article makes use of `std::max_element` in their example, but that's just an example problem where you would get an element address, rather than an element index. The bulk of the article is about converting that element address into an element index.

In the example they use the expression:
```cpp
max_element(vi.begin(), vi.end()) - vi.begin()
```

Though really the essential point is (for an element within an array):
```cpp
auto index = elementAddress - arrayAddress; // There is an implicit division by the element size
```

Remember that if you use `p++` on a pointer, it increments the address by the size of the pointed to element, so internally it would compute `p + 1 * sizeof(*p)`. Similarly if you were to write `p + 2` for a pointer, it really means `p + 2 * sizeof(*p)`. When you subtract pointers, it does the reverse operation, including the divide, so `(p + 2) - p == 2`.

Using [`std::distance`](https://en.cppreference.com/w/cpp/iterator/distance) is basically the same as subtracting pointers. Or at least, for *LegacyRandomAccessIterator* it's the same as subtracting pointers. Think contiguous containers, such as `std::array` and `std::vector`, as opposed to non-contiguous containers, such as `std::list`, or any of the tree data structures. For more general pointer types, `std::distance` can abstract the same concept of element distance, where the result is not actually subtracting pointers, but may require iterator traversal, such as for a tree data structure. In the case of a tree, the memory for the various nodes may not be at well defined address offsets from each other, especially after a tree rebalancing operation, though there is still a defined concept of distance, which can be determined by traversing the tree.

----

The article you linked also deals with converting an element address to an index. Though in that article they use an example with `std::find`, which returns an element address, which they then want to convert to an index. The use of `std::find` isn't relevant to the conversion from address to index. It's just an example function that will give you an address, which you might want to convert to an index. The particulars of `std::find` are irrelevant to the conversion, much like the result of `max_element` was. It's just that they both returned addresses rather than indexes. It's `std::distance` that does the actual conversion.

----

Btw, fun fact, if you want to divide, and you know there will be no remainder, then there is a simplification when working modulo a register size. Instead of dividing, you can find a multiplicative inverse (modulo the register size), and a shift value, and instead multiple (with wrap around) and shift to get the effect of division. That means you can divide by using the much faster multiply and shift instructions, rather than the slow and long delayed division instruction. This is particularly useful with arrays since two element addresses within the same array will always differ by a multiple of the element size, and so if you subtract them and divide, there will never be any remainder. Hence the multiplicative inverse technique will apply. Of course, if the division is a power of two, you only need the shift.

There was a section in Hacker's Delight about using multiplicative inverses to do division.

belgianguy: Hi @DanRStevens Thanks for that and you were completely right! I didn't know that was possible! Another thing learned!
I'll also keep in mind the multiplicative inverses, that's a nice win.

I've now implemented it as such:

```
void setSelected(std::size_t index);
void setSelected(RadioButtonGroup::RadioButton* button);
```

```
void RadioButtonGroup::setSelected(std::size_t index)
{
	clearSelection();
	mIndex = index;
	mRadioButtons[index].checked(true);
}
```

```
void RadioButtonGroup::setSelected(RadioButtonGroup::RadioButton* button)
{
	auto index = button - mRadioButtons.data();
	setSelected(index);
}
```

OPHD/UI/Core/RadioButtonGroup.h
belgianguy: Renamed `setSelected` to `select` and renamed `clearSelection` to `clear`

OPHD/UI/Core/RadioButtonGroup.h
belgianguy: Agreed, it's cleaner if it's only set through the constructor. The setter method has been removed.

OPHD/UI/Core/RadioButtonGroup.h
belgianguy: This is still a rather tough one for me, does @ldicker83 perhaps know how we would have to approach this? As we did try some configurations but the compiler would have none of it.

OPHD/UI/Core/RadioButtonGroup.cpp
```diff
@@ -0,0 +1,71 @@
...
+void RadioButtonGroup::select(std::size_t index)
+{
+	clear();
+	mIndex = index;
+	mRadioButtons[index].checked(true);
+}
+
+void RadioButtonGroup::select(RadioButtonGroup::RadioButton* button)
+{
+	auto index = button - mRadioButtons.data();
```
DanRStevens, line 69: I would maybe be tempted to use `std::distance(button, mRadioButtons.data())` here. The division inherent in that subtraction is not super obvious, though would be more expected when calling a method such as `distance`. The two are of course completely equivalent in this case. Just that there is a bit more of a high level abstraction when using `std::distance`.

Using `std::distance` might require an additional header include.

OPHD/UI/Core/RadioButtonGroup.h
```diff
@@ -0,0 +1,65 @@
...
+		NAS2D::DelegateX<void> delegate;
+	};
+
+	RadioButtonGroup() = default;
+	RadioButtonGroup(std::vector<ButtonInfo> buttonInfos);
+	void onMove(NAS2D::Vector<int> displacement) override;
+	void update() override;
+	void clear();
+	void select(std::size_t index);
+	void select(RadioButtonGroup::RadioButton* button);
```
DanRStevens, line 60: This section probably needs to be broken up a bit with blank lines. Special member functions (constructors, destructors, copy/move operators, etc.) are usually separated from custom named members by a blank line. If member functions are grouped into related sets (like `clear`, and the two overloads of `select`), there may also be a blank line between them and other sets of methods.

The `onMove` event handler should probably be in a `protected` section. It's not expected that client code can call this method directly.

DanRStevens (commented):
Ok, excellent! This has cleaned up really nicely. I'm quite happy with the state of the code now.

I noted a few more extremely minor points while reviewing the recent changes. All pretty much optional suggestions.

The one outstanding point from above is the use of `DelegateX` versus `Delegate`. I don't think this needs to be a show stopper, though I would like to at least investigate before merging and the API effectively becomes public. I should have a bit of time on the weekend. If anyone else wants to join in on Discord, they'd certainly be welcome.

I'm looking to merge this on the weekend after the `Delegate` investigation, whether or not there are any additional changes made.

OPHD/UI/Core/RadioButtonGroup.cpp
DanRStevens: I suppose I should draw attention to the note on Complexity from the [`std::distance`](https://en.cppreference.com/w/cpp/iterator/distance) documentation: 

> **Complexity**
>
> Linear.
> However, if `InputIt` additionally meets the requirements of `LegacyRandomAccessIterator`, complexity is constant.

The final note about "complexity is constant" is the one that's relevant here. Basically it's a roundabout way of saying the complexity is the same as the subtraction operation. In this case, the pointer meets the definition of `LegacyRandomAccessIterator`.

OPHD/UI/Core/RadioButton.cpp
```diff
@@ -33,60 +25,41 @@ RadioButton::~RadioButton()
...
-void RadioButton::onMouseDown(EventHandler::MouseButton button, int x, int y)
+void RadioButtonGroup::RadioButton::onMouseDown(EventHandler::MouseButton button, int x, int y)
 {
 	if (!enabled() || !visible() || !hasFocus()) { return; }
 
 	if (button == EventHandler::MouseButton::Left && mRect.contains(Point{x, y}))
 	{
-		click();
 		mSignal();
+		mParentContainer->select(this);
```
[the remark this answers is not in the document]
belgianguy: Agreed, they have been swapped.

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: Agreed, I'm all for code that explains itself, and that distance could indeed add some clarity.
I did need to swap the orders of the parameters to get it to work though:

```
void RadioButtonGroup::select(RadioButtonGroup::RadioButton* button)
{
	auto index = std::distance(mRadioButtons.data(), button);
	select(index);
}
```

I also tried to make sure that this wasn't limited to cases wherein the last item was selected by default, so I tried to select the first item (at position 0), this caused a compilation error because in the case of 0, it needs a cast to make sure it knows which overload to pick:

```
    add(rbg, {mProductGrid.size().x + 12, 25});
    rbg.size({50,50});
    rbg.select(static_cast<std::size_t>(0));
```

OPHD/UI/Core/RadioButtonGroup.h
belgianguy: Done, I also re-arranged the .cpp files structure to better match the new order in the .h file.
The `onMove` has been moved to the `protected` section.

DanRStevens: Ahh excellent. Always good to have the .cpp and .h file list things in the same order.

OPHD/UI/Core/RadioButtonGroup.cpp
DanRStevens: Good catch on the parameter order to `std::distance`.

----

To resolve the overload resolution, you could potentially have used `0u` instead of the `static_cast`. Very interesting that this came up though. I suppose this is one of the reasons they introduced `nullptr` as a keyword, since it's a typed "0", forcing it to be a pointer. Similarly `0u` would be a typed "0", forcing it to be an unsigned integer.

Of course, there is another change that could eliminate this entirely. Change the parameter from a pointer to a reference. I would actually recommend this anyway, since references are non-nullable, and you really don't want someone to pass `nullptr` to that function. It also has the nice side effect of clearing up any ambiguity over what `0` would mean, since `0` isn't a valid reference.

OPHD/UI/Core/RadioButtonGroup.h
DanRStevens: Ahh, because most code uses `Signal` rather than `Delegate`. The template parameter list has similar styles for `Signal` and `DelegateX`, but a different style for `Delegate`.

Actually, this file is a bit odd since `mSignal` is defined as a `Delegate`, rather than as a `Signal` object. That makes the API a bit inconsistent here, since `RadioButton` objects can then have only a single listener, while most other controls allow for an arbitrary number of listeners. Of course, we removed the accessor for the `mSignal` object, so there is no way to attach more listeners anyway.

Granted, it makes sense to supply a single listener when constructing the `RadioButton`, so I'm fine with using `Delegate`/`DelegateX` for the `RadioButtonInfo` struct and the `RadioButton` constructor.

I'm wondering if we should perhaps switch back to using a `Signal` object for `mSignal` internally. I could go either way on adding back the accessor for the `SignalSource` slice of that object. I suppose it would make the API more consistent if we added it back.

----

Ok, so if we want to use `Delegate` instead of `DelegateX`, we would specify a template parameter of `void()` rather than `void`.

If we updated `mSignal` to be a `Signal`, then in the `RadioButton` constructor, we would need to change the assignment to a `connect` call:
```cpp
mSignal.connect(delegate);
```

OPHD/UI/Core/RadioButtonGroup.cpp
belgianguy: Ah yes, that's a very fitting solution, with the added bonus that it's more robust against nullptr values. Fixed.

OPHD/UI/Core/RadioButtonGroup.h
DanRStevens: Actually, if we could make use of `Signal<>::DelegateType` to define the type. Or event `decltype(mSignal)::DelegateType` (that might required `typename` in front). That way the delegate declaration would be automatically updated in case the `mSignal` declaration was ever changed.

DanRStevens (approved):

belgianguy:
Sample code for RadioButtonGroup and RadioButton usage:

FactoryProduction.h
```cpp
	RadioButtonGroup rbg{{
		   {"One", {this, &FactoryProduction::onCheckBoxIdleChange}},
		   {"Two", {this, &FactoryProduction::onCheckBoxIdleChange}},
		   {"Three", {this, &FactoryProduction::onCheckBoxIdleChange}}
        }};
```

FactoryProduction.cpp
```cpp
    add(rbg, {mProductGrid.size().x + 12, 25});
    rbg.size({50,50});
    rbg.select(0);
```

This would add three RadioButtons in a vertical list in the FactoryProduction window.

PR #891, opened 2021-05-23
Fixes for RadioButtonGroup

DanRStevens:
The Windows project file update was accidentally missed from #879. The CI builds didn't run since the PR was from a repository fork rather than the main repository. It was caught by CI after the PR was merged into the main repository.

Additionally, with the increased warning level for the Clang compiler, it was noted that `RadioButton` should have explicitly declared copy and move constructors. Technically the default implementation is incorrect, since it won't properly handle the global event source, however, this won't matter in practice since the code is written so as to never copy or move `RadioButton` objects. In particular, the `std::vector::reserve` call prevents the vector from being resized, which would require copy or move. At any rate, the methods are needed to form a `std::vector` of the objects.

Added a `static_cast` to eliminate an MSVC conversion warning.

Added double blank lines between method implementations to match project convention.

### DIFF
--- a/OPHD/UI/Core/RadioButton.cpp
+++ b/OPHD/UI/Core/RadioButton.cpp
@@ -1,6 +1,8 @@
 #include "RadioButtonGroup.h"
 
+
 using namespace NAS2D;
+
 
 RadioButtonGroup::RadioButton::RadioButton(RadioButtonGroup* parentContainer, std::string newText, NAS2D::Delegate<void()> delegate) :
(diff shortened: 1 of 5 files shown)


DanRStevens:
It's a touch ironic that I was commenting earlier today how we usually wait until CI has finished before merging, and how it often catches things even when I think it's good. Then I go and merge something without CI. Though to be fair, there were no listed failures, since CI didn't run at all. I tend to be blind to a lack of success, provided there is also a lack of failure.

PR #892, opened 2021-05-23
Remove unused code

ldicker83:


DanRStevens (approved):
I assume just use the table directly?

ldicker83:
> I assume just use the table directly?

It's not actually used anywhere. I figure if/when it's needed I can put accessors back in. More importantly I noticed the "Difficulty from string" function and was like, yeah, that needs to not be a thing.

Issue #824, opened 2021-03-09
Investigate feasibility of using an ECS (Entity Component System) for Structures


rmtoth:
This suggestion was brought up in #714 as a way of getting more convenient access to StructureClass-specific functionality, rather than having to typecast to a specific Structure subclass.

DanRStevens:
Re-posting links for ease of reference:
- https://en.wikipedia.org/wiki/Entity_component_system
- https://medium.com/ingeniouslysimple/entities-components-and-systems-89c31464240d

----

So my understanding of this idea so far, is structure abilities would have associated types, with one instance per each structure that has that ability. The structure would contain all the fields needed to implement that ability. The associated types would be stored in their own collections, and there would be a mapping between the ability instance and the structure that has it. In database terms, this would be a one-to-many relationship (1 structure, has many abilities, with each ability being owned by exactly 1 structure).

For example, we might have abilities for `FoodStorage`, `WarehouseStorage`, `PowerProduction`, `FactoryProduction`, etc. A Command Center would have abilities for `FoodStorage` and `PowerProduction`. An Agridome would have only `FoodStorage`. If you built 1 Command Center and 1 Agridome, there would be 2 entries in the `FoodStorage` table, and 1 entry in the `PowerProduction` table.

The Structure class itself would be very minimal, or perhaps even non-existent? If all meaningful properties were components, then a `Structure` might simply be the ID that ties all the component fields together.

----

This sort of design necessarily means a lot of references. Each ability component would have a reference to the structure that owns it. Do we need mapping in the reverse direction? How would a structure know what components it has, or does it even care? We could have a collection of component references, though I'm not sure that has much value, since components are of different types, so there'd be no real purpose in iterating over such a collection. You could potentially have fields in the Structure to point to each of the possible components, of which there is at most one instance, though that might waste space since most Structures will have only 1 or 2 components. Or you could just scan all the component tables for references back to the structure. That would be much slower, but not waste memory. But do we even care? If what we really want is to iterate over specific abilities, and do processing that way, then maybe we don't actually care about the structure itself, or mapping from a structure to abilities. Maybe that's only relevant if a structure is destroyed, and we need to cleanup associated components. That should happen rarely enough that speed isn't of much concern. I'm guessing that is the design approach of choice.

----

[...]

PR #898, opened 2021-05-24
Comm range refactor

Brett208:
I've been really inactive for a long time. Glad to see all the work on the project though by others. Some minor exploratory work today since I had some time.

Any issue if I create an `IStructureWithAreaAffect` interface? I want to move pulling the communication range from an external constant to a get function defined by this structure interface. This interface could also be used in the future to pull the police station ranges or any other structure that affects an area. To be less generic it could be namesd 'ICommunicationStructure'. 

I think this will also make changing the affect range of various structures via research easier if we go that route. 

It may also allow collapsing the 2 for statements within `checkCommRangeOverlay`.

```C++
for (auto cc : command)
{
	if (!cc->operational()) { continue; }
	auto& centerTile = structureManager.tileFromStructure(cc);
	auto commAreaRect = buildAreaRectFromTile(centerTile, constants::ROBOT_COM_RANGE);
	fillCommList(centerTile, commAreaRect, constants::ROBOT_COM_RANGE);
}

for (auto tower : commTowers)
{
	if (!tower->operational()) { continue; }
	auto& centerTile = structureManager.tileFromStructure(tower);
	auto commAreaRect = buildAreaRectFromTile(centerTile, constants::COMM_TOWER_BASE_RANGE);
	fillCommList(centerTile, commAreaRect, constants::COMM_TOWER_BASE_RANGE);
}
```

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -87,7 +87,7 @@ static NAS2D::Rectangle<int> buildAreaRectFromTile(const Tile& centerTile, int c
 }
 
 
-static void fillCommList(TileList& tileList, TileMap& tileMap, Tile& centerTile, const NAS2D::Rectangle<int>& area, int commRange)
+void MapViewState::fillCommList(TileList& tileList, TileMap& tileMap, Tile& centerTile, const NAS2D::Rectangle<int>& area, int commRange)
 {
 	for (int y = 0; y < area.height; ++y)
 	{
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -87,25 +87,6 @@ static NAS2D::Rectangle<int> buildAreaRectFromTile(const Tile& centerTile, int c
 }
 
 
-void MapViewState::fillCommList(Tile& centerTile, const NAS2D::Rectangle<int>& area, int commRange)
-{
-	for (int y = 0; y < area.height; ++y)
-	{
-		for (int x = 0; x < area.width; ++x)
-		{
-			auto& tile = (*mTileMap).getTile({ x + area.x, y + area.y });
-			if (isPointInRange(centerTile.position(), tile.position(), commRange))
-			{
-				if (std::find(mCommRangeOverlay.begin(), mCommRangeOverlay.end(), &tile) == mCommRangeOverlay.end())
-				{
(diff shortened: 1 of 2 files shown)


Brett208:
Oh, I forgot to squash the commit _Remove Class Qualifier from function prototype_ with _Make fillCommList a member function of MapViewState instead of static._ before pushing. Not sure how @ldicker83 feels about abusing history post public push.

DanRStevens (approved):
You've been missed :slightly_smiling_face: 

I agree, that was an odd use of `static`. Technically, yes, the method as written didn't access any member variables, and as such should be declared `static`. However, as you've noted, it simply passed member variables in as parameters. In that case, yes, I agree it should just be made a regular member function and access the variables directly.

I don't think Leeor has a problem with `git rebase` on branches or PRs.

DanRStevens:
Oh, in terms of the proposed new `IStructureWithAreaAffect` interface, I just wanted to mention there was some incomplete work done to move towards an Entity Component System (ECS). It may related to that idea. See #824, and to a lesser extent #843. There is a branch for the current work, but no associated PR.

With that said, I'm not too sure how a common interface for the effect ranges of various buildings would be used. Particularly since those buildings will have different effects within those areas. I suppose it might help for map highlighting. I can't really think of other uses.

Another concern is the interface might potentially lead to multiple inheritance, which I would kind of like to avoid. Though I suppose that's not a terrible thing for interfaces.

Anyway, if you think there is value in it then give it a try and see.

belgianguy:
In the original there were a few buildings that also had a range effect applied:

Above ground:
- Command Centers: new Command Centers cannot be placed within 30 tiles of an existing Command Center 
- Communication Tower: Unless a Communications Satellite was brought, these are needed to be able to deploy your rovers further out
- DIRT: Protects buildings in a 5 tile radius around itself
- Hot Lab: Explosive damage area of effect
- Police: Benevolent: Reduce Crime level by half in radius of 5 tiles, while improving Morale by one point. Hostile: Reduce Crime level to 0 in radius of 5 tiles, Morale also reduced by 1 point.

Underground:
- Administration center: Light Industry and Laboratories immeddiately adjacent get an additional 1 point production rate/research progress. (range of 1?)
- Police (UG): Same as above ground Police
- Tokamak Core: The game implied people didn't like living too close to this.

PR #900, opened 2021-05-24
Normalizes calls to show/hide functions in controls

ldicker83:


DanRStevens:
Now that I see this, yes. That's both terser and possibly easier to read.

Actually, in general, I like to avoid boolean parameters, since they tend to be not very self documenting at the point of call. In this case I don't think it was too hard to read with just a single parameter, though when there are multiple parameters, booleans aren't great for clarity. Similar for enum parameters in some cases. Particularly two valued enums, which could result in two functions, rather than an extra parameter.

ldicker83:
Yeah, it's one of those where I have the `visible(bool)` function there for cases where I want to set the visibility based on a boolean expression. Using it the way it was is technically not incorrect but it's... I dunno. `show()`/`hide()` just seems more natural.

PR #901, opened 2021-05-25
Comm range refactor2

Brett208:
Sort of took a middle of the road approach and didn't actually create an interface for structures effecting an area.

I think all these changes are moving in the right direction?

Sort of want to try to start tracking police station coverage similar to comm coverage in the near future.


DanRStevens (approved):
I like this change.

PR #904, opened 2021-05-25
Surface police range overlay

Brett208:
If this looks good, the next step is to either get the underground views showing or make the surface police actually do something useful.

It wasn't too difficult to modify the comm range code to also handle police coverage, so nice job on the original design there. It also wasn't difficult to modify the UI code positioning, so appreciate the time put in developing it.

I'm not beholden to using red or this shade of red if someone has a better color.

See new icon on display below.

Police Overlay
.

### DIFF
--- a/OPHD/Constants/Strings.h
+++ b/OPHD/Constants/Strings.h
@@ -210,6 +210,7 @@ namespace constants
 	const std::string ToolTipBtnConnectedness = "Toggles display of the Tile Connection Overlay";
 	const std::string ToolTipBtnCommRange = "Toggles display of the Communications Range Overlay";
 	const std::string ToolTipBtnRoutes = "Toggles display of the Trucking Routes Overlay";
+	const std::string ToolTipBtnPolice = "Toggles display of the Police Station Coverage";
 	const std::string ToolTipRefinedResources = "Breakdown of storage levels for Smelted Resources";
 	const std::string ToolTipResourceStorage = "Smelted Resources Storage. Current Level / Capacity";
 	const std::string ToolTipFoodStorage = "Food Storage. Current Level / Capacity";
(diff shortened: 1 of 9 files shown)


DanRStevens (approved):
Technical side looks good to me.

We should probably merge the icon update to `master` in the assets repository before merging this.

Curious if @ldicker83 wants to comment on any design aspect?

belgianguy:
Nice work! Indeed it should then be written out:
* What defines the game element of Crime? (Do other Structures 'emit' crime that the Police station then neutralizes?)
* What causes Crime to rise and how do we define that? (Low Morale? Idle Workers? Low food?)
* How is it represented in game? (As a counter? Low values is low crime, high values is high crime?)
* What other game elements are affected by it? (Residential units transforming into Red Light Districts, ... )

How about blue-ish? Must take care not to be the same blue as the mouse tile selection highlight.
About the iconography, perhaps a solid fill fits better with the others?
Will OPHD also feature the DIRT structure? Then we might need to think of another "protection" icon.

Brett208:
@ldicker83, have you recorded your thoughts on how to implement the police stations somewhere? I poked around the forum but didn't see anything obvious. Probably need to talk a bit before continuing to make sure this aligns with your vision.

I had considered implementing chance of stealing of food or metals from agridomes and storage facilities if they are outside police coverage for above ground. Based on the original game, I think the behaviour for below ground is moral modifier if residence is covered by a police station, plus stopping other structures from turning into red light districts. I hadn't put much thought into a more comprehensive crime rate/percent calculation. A binary calculation where presence of protection is present or not present would be simpler to display and implement. Anyways, curious to hear where Leeor wants to go, or this can be shelved for now as well.

ldicker83:
Not too keen on the icon that's there. Here's one from the same set that can be included in the credits list that's in the data repository:

police

Put this one through the ringer so it's fully compressed and stripped of unnecessary metadata.

> @ldicker83, have you recorded your thoughts on how to implement the police stations somewhere? I poked around the forum but didn't see anything obvious. Probably need to talk a bit before continuing to make sure this aligns with your vision.

I have not. Honestly I haven't given it a whole lot of thought yet so it's entirely up for debate. If I had to take a quick crack at it I'd say that some buildings 'emit' crime at a constant rate as @belgianguy suggested which can be affected by a difficulty modifier and adjusted based on current morale. Police would also affect that.

I'd recommend a separate task/issue to further the discussion.

PR #916, opened 2021-05-30
Add StructureManager::getStructures<StructureType>() template method

### LINKED ISSUE #714: Add typecast method to StructureManager to get lists of derived structure types
From PR #713, @DanRStevens mentioned a way to improved some of the code by adding a templated method to StructureManager that allows getting derived structure types:

> This gets me thinking, maybe we could have a templated `StructureManager` method that allows getting structures of a certain type, and would use that exact type for the return value. Then you could write something like:
> ```cpp
> const auto warehouses = structureManager.structureList<Warehouse>();
> for (auto warehouse : warehouses)
> {
>   warehouse->warehouseSpecificMethod();
> }
> ```
> 
> _Originally posted by @DanRStevens in https://github.com/OutpostUniverse/OPHD/pull/713#discussion_r495585281_

This sound like a great way to improve code use

DanRStevens:
Closes #714

Add a `constexpr` template method `structureTypeToClass` to convert a structure type (C++ class) to a `StructureClass` enum value.

Implement `StructureManager::getStructures<StructureType>()` template method to get a list of derived structure types.

Remove a lot of no longer needed casts from `Structure*` to derived types.

### DIFF
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -468,7 +468,7 @@ int getTruckAvailability()
 {
 	int trucksAvailable = 0;
 
-	auto& warehouseList = NAS2D::Utility<StructureManager>::get().structures<Warehouse>();
+	auto& warehouseList = NAS2D::Utility<StructureManager>::get().getStructures<Warehouse>();
 	for (auto warehouse : warehouseList)
 	{
 		trucksAvailable += warehouse->products().count(ProductType::PRODUCT_TRUCK);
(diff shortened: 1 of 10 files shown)


DanRStevens:
I avoided the name `structureList` since I didn't want a template method to be overloaded with a regular method. The original non-template method still exists, and is used by the new template method. Hence the new name `getStructures`.

ldicker83:
This will make the code so much cleaner and easier to deal with. Good work! :D

PR #924, opened 2021-05-31
Use std::back_inserter in call to copy_if

ldicker83:
Fixes an exception being thrown with the message "can't dereference value-initialized vector iterator". This was due to attempting to copy elements from one container into another that wasn't properly sized.


DanRStevens (approved):

DanRStevens:
Bug was first introduced in #913, commit 164582339c0fb8c5dd6d4287b9d4a9bd22f74e36, then temporarily fixed in fab9b999f60ddbedb5faa8bf2f27f349630f1bcc, before being re-introduced in #916 commit ac9a39363852d507b5a3e016317bc685d328734e.

It seems `std::copy_if` will not expand the destination collection if passed an iterator. Instead, `std::back_inserter` should be passed, which is initialized with the collection (and hence gains access to enlarge the collection).

PR #931, opened 2021-06-01
Fix bug in StructureManager::removeStructure

### LINKED ISSUE #930: Exception Thrown when digging an elevator shaft deeper
Build: Current Master branch on 31May2021
Configuration: x64 Debug Build, running in debug mode in VS2019
OS: Windows 10 64 bit

When trying to make an elevator shaft deeper by digging on top of an existing shaft, an exception is thrown (IE trying to go from UG1 to UG2 on a Mars type planet).

DiggingDownOnTopOfElevator

DiggingDownOnTopOfElevator VisualStudio View

DiggingDownOnTopOfElevator Disassembly View

I'm forgetting where to find the saved game, trying to rummage through the AppData folders and it is getting late. Should be easy to reproduce, but can dig it up later if needed.

DanRStevens:
Closes #930

When structures are removed from collections, the collection `end()` marker may move. Hence checking if an item was found, by checking if `iterator != collection.end()` must be done before the item is removed from the collection.

### DIFF
--- a/OPHD/StructureManager.cpp
+++ b/OPHD/StructureManager.cpp
@@ -260,19 +260,21 @@ void StructureManager::removeStructure(Structure* structure)
 	StructureList& structures = mStructureLists[structure->structureClass()];
 
 	const auto it = std::find(structures.begin(), structures.end(), structure);
-	if (it != structures.end())
+	const auto isFoundStructureTable = it != structures.end();
+	if (isFoundStructureTable)
 	{
 		structures.erase(it);
 	}
 
 	const auto tileTableIt = mStructureTileTable.find(structure);
-	if (tileTableIt != mStructureTileTable.end())
+	const auto isFoundTileTable = tileTableIt != mStructureTileTable.end();
+	if (isFoundTileTable)
 	{
 		tileTableIt->second->deleteThing();
 		mStructureTileTable.erase(tileTableIt);
 	}
 
-	if ((it == structures.end()) || (tileTableIt == mStructureTileTable.end()))
+	if (!isFoundStructureTable || !isFoundTileTable)
 	{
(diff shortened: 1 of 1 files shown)


OPHD/StructureManager.cpp
```diff
@@ -260,19 +260,21 @@ void StructureManager::removeStructure(Structure* structure)
...
-	if (tileTableIt != mStructureTileTable.end())
+	const auto isFoundTileTable = tileTableIt != mStructureTileTable.end();
+	if (isFoundTileTable)
 	{
 		tileTableIt->second->deleteThing();
 		mStructureTileTable.erase(tileTableIt);
 	}
 
-	if ((it == structures.end()) || (tileTableIt == mStructureTileTable.end()))
+	if (!isFoundStructureTable || !isFoundTileTable)
```
ldicker83, line 277: I'm wondering if this check is even necessary anymore? I originally had the check in there to ensure that I wasn't doing stupid stuff during debugging but it seems like a relatively simple change. By removing this check the above two comparisons can be left as they were in what I would consider a more expressive format versus adding boolean flags.

If the check is deemed necessary this code looks fine, otherwise I'd remove the check and change the above two lines back to what they were.

Brett208 (approved):
Just checked and this fixes the issue I was seeing. Thanks for the quick fix.

OPHD/StructureManager.cpp
DanRStevens: You're probably right. I had briefly considered just removing the check. Though I was a little concerned that with the active development we could easily re-introduce a case where we accidentally try to delete a non-existent building. I figured it might be better to leave the check intact for now.

Anyway, I'm going to merge as is, as it's the fastest way to solve the bug, though I'm quite open to removing the check in a followup PR if people want to go that route.

ldicker83: Sounds good to me. I have no objections.

Issue #911, opened 2021-05-28
Implement Crime


Brett208:
From @belgianguy 

> Nice work! Indeed it should then be written out:
> 
> * What defines the game element of Crime? (Do other Structures 'emit' crime that the Police station then neutralizes?)
> * What causes Crime to rise and how do we define that? (Low Morale? Idle Workers? Low food?)
> * How is it represented in game? (As a counter? Low values is low crime, high values is high crime?)
> * What other game elements are affected by it? (Residential units transforming into Red Light Districts, ... )
> 
> How about blue-ish? Must take care not to be the same blue as the mouse tile selection highlight.
> About the iconography, perhaps a solid fill fits better with the others?
> Will OPHD also feature the DIRT structure? Then we might need to think of another "protection" icon.

From @ldicker83 

> Not too keen on the icon that's there. Here's one from the same set that can be included in the credits list that's in the data repository:
> 
> ![police](https://user-images.githubusercontent.com/13858902/119896411-f96b3e80-bf0c-11eb-8668-f7335c802ba6.png)
> 
> Put this one through the ringer so it's fully compressed and stripped of unnecessary metadata.
> 
> > @ldicker83, have you recorded your thoughts on how to implement the police stations somewhere? I poked around the forum but didn't see anything obvious. Probably need to talk a bit before continuing to make sure this aligns with your vision.
> 
> I have not. Honestly I haven't given it a whole lot of thought yet so it's entirely up for debate. If I had to take a quick crack at it I'd say that some buildings 'emit' crime at a constant rate as @belgianguy suggested which can be affected by a difficulty modifier and adjusted based on current morale. Police would also affect that.
> 
> I'd recommend a separate task/issue to further the discussion.

See PR #910 or #904 for initial discussion. That PR implements a view of police station coverage and concept of police station affecting range.

belgianguy:
I've reviewed all the resources of the original game, including the manual, and it seems a really underlit game element, I could only find the following:
* There is no indication in the population overview how high the crime is nor any other report that states a number/state.
* Overcrowded Residential Units cause crime to rise (ingame outpost.hlp, Residential)
* High crime causes Residential Units to convert into Red Light Districts
* Police buildings lower crime, how much depends on their crime fighting setting
* Placing more Police buildings in troubled areas stops Residential Unit conversion into Red Light Districts
* Benevolent Police divide crime by 2 in a range of 5 tiles and raise Morale by one, another manual says that Benevolent is +1 Morale
[...]

PR #932, opened 2021-06-03
Get all OreProcessing derived structures, not just Smelter

ldicker83:
Fixes an issue in route processing where the StructureManager was asked to return all Smelter's, not OreProcessing structures. This would skip anything except the derived Smelter class so the SEED Smelter would be ignored. This should resolve the behavior reported by @Brett208 here: https://github.com/OutpostUniverse/OPHD/issues/911#issuecomment-853507343

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -33,7 +33,7 @@ static inline void pullFoodFromStructure(FoodProduction* producer, int& remainde
 }
 
 
-static RouteList findRoutes(micropather::MicroPather* solver, TileMap* tilemap, Structure* mine, const std::vector<Smelter*>& smelters)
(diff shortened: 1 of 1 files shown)


DanRStevens:
Ahh, it seems my search/replace update in PR #932 has some unintended changes. It seems `Structure::StructureClass::Smelter` doesn't correspond exactly with `class` `Smelter`, but rather corresponds most closely with `class` `OreRefining`, which includes `Smelter` and `SeedSmelter`.

There were two places affected by that change. You may want to investigate the other in `transportResourcesToStorage`.
```cpp
void MapViewState::transportResourcesToStorage()
{
	auto& smelterList = NAS2D::Utility<StructureManager>::get().getStructures<Smelter>();
	for (auto smelter : smelterList)
```

I believe that one needs to be updated as well, since it used to use `structureList(Structure::StructureClass::Smelter)`.

PR #933, opened 2021-06-04
Fix potential undefined behavior in casting

ldicker83:
Fixes a case of `static_cast`'ing from a base class to a `Smelter` which could actually be a `SEED Smelter`. This probably didn't have immediate effects because all of the functions used are implemented in `OreRefining` but there are all sorts of nefarious things that could happen here. Might even be responsible for some of the mysterious heap corruption issues the debugger has been complaining about.

Anyway, point being this is clearly incorrect and is how obscure and annoying bugs are introduced. This will hopefully prevent explosions going forward.

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -282,7 +282,7 @@ void MapViewState::transportOreFromMines()
 		if (routeIt != routeTable.end())
 		{
 			const auto& route = routeIt->second;
-			const auto smelter = static_cast<Smelter*>(static_cast<Tile*>(route.path.back())->structure());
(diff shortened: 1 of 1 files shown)


DanRStevens:
Hmm, nice catch. That would be really easy to miss.

I suppose this is one of the reasons to avoid casting. There's no way the compiler can really check that.

ldicker83:
> Hmm, nice catch. That would be really easy to miss.
> 
> I suppose this is one of the reasons to avoid casting. There's no way the compiler can really check that.

Generally speaking agreed, but at least with explicit cast operators we can actually see where the type system is being subverted. I just happened to notice this when I was going over a few things and noticed that we were using `Smelter*` pointers instead of `OreRefining*`... wherps.

Might be good to audit other sections of code that use interfaces for derived types like Warehouse, etc.

PR #935, opened 2021-06-04
Fix truck assignment not saved

### LINKED ISSUE #928: Trucks disappear upon save/load
If I have M mines and N trucks (where N > M) and :
1. Save the game.
2. Close the game.
3. Load the game.

There will now be only M trucks (1 per mine, M = N).
Available trucks always resets to 0.

In my case I had 2 mines and 6 trucks, 5 trucks assigned to Mine A and 1 truck assigned to Mine B.
I saved, closed and later opened and loaded the same game.
Mines A and B had A truck each and there were 0 available trucks.

Unfortunately the resources spent to create the trucks doesn't reappear 😝

ldicker83:
Closes #928

### DIFF
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -97,7 +97,6 @@ void MapViewState::save(const std::string& filePath)
 	mTileMap->serialize(root, mPlanetAttributes);
 	Utility<StructureManager>::get().serialize(root);
 	writeRobots(root, mRobotPool, mRobotList);
-	//writeResources(root, mPlayerResources, "resources");
 	writeResources(root, mResourceBreakdownPanel.previousResources(), "prev_resources");
 
 	XmlElement* turns = new XmlElement("turns");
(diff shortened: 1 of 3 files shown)


OPHD/States/MapViewStateIO.cpp
```diff
@@ -393,6 +393,24 @@ void MapViewState::readStructures(Xml::XmlElement* element)
 			mineFacility.mine(mine);
 			mineFacility.maxDepth(mTileMap->maxDepth());
 			mineFacility.extensionComplete().connect(this, &MapViewState::onMineFacilityExtend);
+
+			auto trucks = structureNode->firstChildElement("trucks");
+			if (trucks == nullptr)
+			{
+				throw std::runtime_error("MapViewState::readStructures(): MineFacility structure saved without a trucks node.");
+			}
```
DanRStevens, line 401: Just wanted to comment that a check like this will invalidate all saved games where a mine facility exists.

ldicker83: True. Maybe I should change this to try to read data if it's not `nullptr` versus throwing if it's not found? It's not really a change that needs to throw away older saves.

DanRStevens: That's kind of what I was thinking. Could potentially just accept any lost data, or perhaps restore using default values.

PR #939, opened 2021-06-05
Add take me there to notification window

### LINKED ISSUE #923: Add a "take me there" button to NotificationWindow
This is dependent on #922 

Similar to the Take Me There buttons in the full-screen UI's, this will only show up in the Notification Window when a position is added to a Notification. This can be determined if the position is anything other than { -1, -1 }

ldicker83:
Closes #923

### DIFF
--- a/OPHD/UI/NotificationWindow.cpp
+++ b/OPHD/UI/NotificationWindow.cpp
@@ -19,6 +19,11 @@ NotificationWindow::NotificationWindow():
 	btnOkay.size({ 50, 20 });
 	btnOkay.click().connect(this, &NotificationWindow::btnOkayClicked);
 
+	add(btnTakeMeThere, { 10, 195 });
+	btnTakeMeThere.size({ 125, 20 });
+	btnTakeMeThere.click().connect(this, &NotificationWindow::btnTakeMeThereClicked);
+	btnTakeMeThere.hide();
+
 	add(mMessageArea, { 5, 65 });
 	mMessageArea.size({ size().x - 10, 125 });
 	mMessageArea.font(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL);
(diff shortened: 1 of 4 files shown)


OPHD/UI/NotificationWindow.cpp
```diff
@@ -39,12 +46,20 @@ void NotificationWindow::btnOkayClicked()
...
+}
+
+
 void NotificationWindow::update()
 {
 	if (!visible()) { return; }
 
 	Window::update();
 
+	btnTakeMeThere.visible(mTakeMeThereHidden); // bit of a hack
```
DanRStevens, line 62: Considering the inverted truth nature of this, that is quite surprising and confusing.

OPHD/UI/NotificationWindow.h
```diff
@@ -26,4 +33,6 @@ class NotificationWindow : public Window
 	Button btnTakeMeThere{ "Take Me There" };
 	TextArea mMessageArea;
 	bool mTakeMeThereHidden{ false };
+
+	Signal mTakeMeThereClicked;
```
DanRStevens, line 37: Wondering if we should use a present tense for this?
`mTakeMeThereClick`

ldicker83: Well, considering that the action has already happened (e.g., the button has been clicked, the mouse button was let go of), it seems appropriate.

I'm not really opposed to a present tense but I don't know if there's really a value in that.

OPHD/UI/NotificationWindow.cpp
ldicker83: How so? Variable naming is, admittedly, very poor but the functionality is correct.

OPHD/UI/NotificationWindow.h
DanRStevens: A lot of the event handler functions were updated to use present tense. It might be more consistent if everything used present tense.

If you think about it, it's not really until the event handler code has run that the event is really complete. So the point where that object is relevant, the event is still live being processed.

ldicker83: I mean, you bring up a good point. I'll have to adjust the way I'm thinking about it. I looked into some of the .Net code for C# GUI applications and they follow the present tense format as well so I'll start to conform to thinking in the present tense as well. Might take a bit of time to adjust though! :D

PR #943, opened 2021-06-06
Show brief on notification hover

### LINKED ISSUE #942: Show brief on hover over NotificationArea
Show a brief message popup when hovering over notification area icons. This doesn't need to have a timer, just a basic highlight with the notification brief

ldicker83:
Closes #942

### DIFF
--- a/OPHD/UI/NotificationArea.cpp
+++ b/OPHD/UI/NotificationArea.cpp
@@ -132,9 +132,10 @@ void NotificationArea::onMouseMove(int x, int y, int /*dX*/, int /*dY*/)
 			mNotificationIndex = count;
 
 			const int stringWidth = mFont.width(mNotificationList[count].brief) + 8;
-			const int briefPositionX = positionX() - stringWidth - 4;
+			const int briefPositionX = positionX() - stringWidth;
+			const int briefPositionY = rect.y + (rect.height / 2) - (mFont.height() / 2) - 2;
 
-			mNotificationBriefRect = { briefPositionX, rect.y, stringWidth, mFont.height() + 4 };
+			mNotificationBriefRect = { briefPositionX, briefPositionY, stringWidth, mFont.height() + 4 };
 
 			return;
 		}
@@ -189,8 +190,6 @@ void NotificationArea::update()
 
 		if (mNotificationIndex == count)
 		{
-			renderer.drawBox(rect);
-
 			renderer.drawBoxFilled(mNotificationBriefRect, Color::DarkGray);
 			renderer.drawBox(mNotificationBriefRect, Color::Black);
 
--- a/OPHD/UI/NotificationArea.cpp
+++ b/OPHD/UI/NotificationArea.cpp
@@ -56,7 +56,8 @@ static constexpr int Offset = constants::MARGIN_TIGHT + 32;
 
 
 NotificationArea::NotificationArea() :
-	mIcons{ imageCache.load("ui/icons.png") }
+	mIcons{ imageCache.load("ui/icons.png") },
+	mFont{ fontCache.load(constants::FONT_PRIMARY, constants::FONT_PRIMARY_NORMAL) }
 {
 	auto& eventhandler = Utility<EventHandler>::get();
 
@@ -128,6 +129,9 @@ void NotificationArea::onMouseMove(int x, int y, int /*dX*/, int /*dY*/)
 		if (rect.contains({ x, y }))
 		{
 			mNotificationIndex = count;
+
+			mNotificationBriefRect = { 100, 100, 100, 25 };
+
 			return;
 		}
 
--- a/OPHD/UI/NotificationArea.cpp
+++ b/OPHD/UI/NotificationArea.cpp
@@ -111,6 +111,7 @@ void NotificationArea::onMouseDown(EventHandler::MouseButton button, int x, int 
 			mNotificationList.erase(mNotificationList.begin() + count);
 			mNotificationRectList.erase(mNotificationRectList.begin() + count);
 			updateRectListPositions();
+			onMouseMove(x, y, 0, 0);
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


OPHD/UI/NotificationArea.cpp
```diff
@@ -118,9 +118,23 @@ void NotificationArea::onMouseDown(EventHandler::MouseButton button, int x, int
...
+	for (auto& rect : mNotificationRectList)
+	{
+		if (rect.contains({ x, y }))
+		{
+			mNotificationIndex = count;
+			return;
+		}
+
+		count++;
+	}
```
DanRStevens: Nothing wrong with this. Just wanted to mention it would be possible to use [`std::find_if`](https://en.cppreference.com/w/cpp/algorithm/find) and [`std::distance`](https://en.cppreference.com/w/cpp/iterator/distance) to compute this.

```cpp
	const auto contains = [point = Point<int>{x, y}](const Rectangle<int>& rect) { return rect.contains(point); };
	const auto iterator = std::find_if(mNotificationRectList.begin(), mNotificationRectList.end(), contains);
	if (iterator != mNotificationRectList.end()) {
		mNotificationIndex = std::distance(mNotificationRectList.begin(), iterator);
		// Rest of code for found case
	}
```

I think I prefer the readability of ranged-for loops over some STL algorithms. Mind you, with the Range changes coming in C++20, those STL algorithms should become must shorter and more readable, so I'm starting to think of how the STL can be used more. In particular, the Range changes should allow less repetition, as you can then pass collections instead of pairs of iterators, so no need for the `begin()` and `end()` pairs.

OPHD/UI/NotificationArea.cpp
```diff
@@ -118,9 +118,23 @@ void NotificationArea::onMouseDown(EventHandler::MouseButton button, int x, int
...
+
+	mNotificationIndex = SIZE_MAX;
```
DanRStevens, line 146: This would read better with a named constant that gave better semantics:
```cpp
mNotificationIndex = NoSelection;
```

Same for initialization:
```cpp
std::size_t mNotificationIndex{NoSelection};
```

For that, the class should probably define a `static constexpr` value that gives this a more meaningful name.
```cpp
static constexpr auto NoSelection{SIZE_MAX};
```

Or possibly even better, use a typed expression rather than the untyped `SIZE_MAX` pre-processor macro. That will ensure the `auto` deduces the correct type:
```cpp
static constexpr auto NoSelection{std::numeric_limits<std::size_t>::max()};
```

OPHD/UI/NotificationArea.h
```diff
@@ -65,10 +65,13 @@ class NotificationArea : public Control
 	void updateRectListPositions();
 
 	const NAS2D::Image& mIcons;
+	const NAS2D::Font& mFont;
+
 	std::vector<Notification> mNotificationList;
 	std::vector<NAS2D::Rectangle<int>> mNotificationRectList;
 
 	std::size_t mNotificationIndex{ SIZE_MAX };
+	NAS2D::Rectangle<int> mNotificationBriefRect{ 0 };
```
DanRStevens, line 75: You can initialize using empty `{}`. That might be less confusing, as here there is a 4 element struct, where only 1 element is listed and 3 are omitted. Specifying one element kind of makes this look like it's initializing a scalar type, which it's not.
```cpp
NAS2D::Rectangle<int> mNotificationBriefRect{};
```

----

If you want to get pedantic about how the standard treats this, using C++ legalese:

Using `{}` on a struct causes [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization). When fewer parameters are given than struct elements, this leads to value initialization of the unspecified elements.
> If the number of initializer clauses is less than the number of members or initializer list is completely empty, the remaining members are value-initialized.

When [value initialization](https://en.cppreference.com/w/cpp/language/value_initialization) is done on the element type, which in this case is the primitive type `int`, which is not a class type, this translates to zero initialization.
> 4) otherwise, the object is zero-initialized.

When [zero initialization](https://en.cppreference.com/w/cpp/language/zero_initialization) is done, the element values are set to zero of the corresponding type.
> If T is a scalar type, the object's initial value is the integral constant zero explicitly converted to T.

So according to the requirements of the C++ standard, `{}` applied to a struct of `int` values will set each `int` value to `0`.

OPHD/UI/NotificationArea.cpp
```diff
@@ -130,7 +131,10 @@ void NotificationArea::onMouseMove(int x, int y, int /*dX*/, int /*dY*/)
 		{
 			mNotificationIndex = count;
 
-			mNotificationBriefRect = { 100, 100, 100, 25 };
+			const int stringWidth = mFont.width(mNotificationList[count].brief) + 8;
+			const int briefPositionX = positionX() - stringWidth - 4;
+
+			mNotificationBriefRect = { briefPositionX, rect.y, stringWidth, mFont.height() + 4 };
```
DanRStevens: Might be wise to use named constants for the magic numbers used here. Perhaps `constexpr int margin = 4;`. Then width can add `margin * 2`, while the others use `- margin` and `+ margin`.

At least, that's my rough interpretation of what those magic numbers represent. Maybe not since x and y margins might not be the same here.

DanRStevens (commented):
Got a bit curious about some of the finer points of C++ while looking over this code. I've included some comments about what I researched. Included are some minor suggestions for improvements, though really, there's nothing wrong with the code.

OPHD/UI/NotificationArea.cpp
ldicker83: I was thinking about doing something like that and then it just managed to poof out of my mind. I was looking for `std::numeric_limits<std::size_t>::max()` but... yeah. Go figure -_-

OPHD/UI/NotificationArea.cpp
ldicker83: UIConstants has these already defined, managed to forget to change the magic numbers to the named constants. Will slate this for an update going forward.

OPHD/UI/NotificationArea.h
ldicker83: I figured a single 0 would fill the entire structure with 0 -- I'm still getting used to the `{}` initializer syntax and making assumptions based on the 'C' part of the language.

PR #947, opened 2021-06-06
Basic crime

Brett208:
**Implements the following:**
* Expands police overlay to show underground coverage
* Adds Crime rate to structures
  * Structures may be marked to support crime similar to how they are marked to require the CHAP
* Saves/Loads crime rate to saved games (doesn't break backwards compatibility)
* Shows structure crime rate in structure inspector window
* Increases or decreases crime rate of each structure each turn based on police station coverage
* Calculates the mean crime rate of the colony
* Displays the mean crime rate on the population panel
* Decreases or Increases morale based on crime rate
* Saves/loads mean crime rate (doesn't break backwards compatibility)

Relates to Issue #911. More work is required before this issue could be considered closed.

Was hoping to get this done before version 0.8.3 was released, but was too slow. :)

### DIFF
--- a/OPHD/States/CrimeRateUpdate.cpp
+++ b/OPHD/States/CrimeRateUpdate.cpp
@@ -1,0 +1,88 @@
+#include "CrimeRateUpdate.h"
+#include "../UI/PopulationPanel.h"
+#include "../Things/Structures/Structure.h"
+#include "../StructureManager.h"
+#include <NAS2D/Utility.h>
+
+namespace CrimeRate
+{
+	bool isProtectedByPolice(const std::vector<TileList>& policeOverlays, Structure* structure);
+	int calculateMoraleChange(int meanCrimeRate);
+	void setPopulationPanel(int moraleChange, int meanCrimeRate, PopulationPanel& populationPanel);
+
+
+	int update(const std::vector<TileList>& policeOverlays, PopulationPanel& populationPanel)
+	{
+		const auto& structuresWithCrime = NAS2D::Utility<StructureManager>::get().structuresWithCrime();
+
+		// Colony will not have a crime rate until at least one structure that supports crime is built
+		if (structuresWithCrime.empty()) {
+			setPopulationPanel(0, 0, populationPanel);
+			return 0;
+		}
+
+		double accumulatedCrime = 0;
+
+		for (auto structure : structuresWithCrime)
+		{
+			int crimeRateChange = isProtectedByPolice(policeOverlays, structure) ? -1 : 1;
+			structure->IncreaseCrimeRate(crimeRateChange);
+
+			accumulatedCrime += structure->crimeRate();
+		}
+
+		int meanCrimeRate = static_cast<int>(accumulatedCrime / structuresWithCrime.size());
+		int moraleChange = calculateMoraleChange(meanCrimeRate);
+
+		setPopulationPanel(moraleChange, meanCrimeRate, populationPanel);
+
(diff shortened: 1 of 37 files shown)


OPHD/States/CrimeRateUpdate.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+		// Colony will not have a crime rate until at least one structure that supports crime is built
+		if (structuresWithCrime.empty()) {
+			setPopulationPanel(0, 0, populationPanel);
+			return 0;
+		}
+
+		double accumulatedCrime = 0;
+
+		for (auto structure : structuresWithCrime)
```
Brett208, line 26: If you combine this loop with the loop in the function `isProtectedByPolice`, you end up with an O(n^2) algorithm. I think we could tamp it down to O(n) by reformatting `policeOverlays` into a graph that the values could be pulled off of in an O(1) fashion. Of course, then you are paying for the creation of the graph. 

Another solution could be to try some sort of binary search here.

I am OK addressing now or leaving and seeing if it becomes a performance issue, but wanted to point it out for specific review. Someone else may have a better solution on how to address than what I have suggested.

OPHD/States/CrimeRateUpdate.cpp
```diff
@@ -0,0 +1,88 @@
...
+		return false;
+	}
+
+
+	int calculateMoraleChange(int meanCrimeRate)
+	{
+		if (meanCrimeRate > 50)
+		{
+			// Reduce morale by 1 for every 10% above 50%
+			return -1 * (meanCrimeRate / 10 - 4);
```
DanRStevens, line 64: I doubt this really matters after optimizations, but you could have just written:
```cpp
return -(meanCrimeRate / 10 - 4);
```

OPHD/States/CrimeRateUpdate.h
```diff
@@ -0,0 +1,12 @@
...
+#include "../Map/Tile.h"
+#include <vector>
+
+class PopulationPanel;
+
+namespace CrimeRate
+{
+	// Returns change in current morale caused by colony's mean crime rate
+	int update(const std::vector<TileList>& policeOverlays, PopulationPanel& populationPanel);
+}
```
DanRStevens, line 12: The meaning of the `int` return type isn't super obvious for a function named `update`. I suspect if this was named in a more descriptive manner, the namespace could be dropped as well.

DanRStevens (approved):
I wouldn't be too worried about performance yet. Besides, police stations cover a 2D area of the map, so you might reasonably expect area coverage problems to be O(n^2) anyway.

ldicker83 (approved):

OPHD/States/CrimeRateUpdate.h
Brett208: Yeah, I agree the return isn't very obvious without the comment. I am having trouble deciding how this module should pass the change in morale back to the main game state to make it less confusing.

ldicker83: This could probably be done fairly cleanly by declaring a module level variable, say, CrimeChange or MoraleChange or whatever you want to call it, put it in an unnamed namespace and then provide a get/set type of function. Since the CrimeRateUpdate isn't a distinct class this is probably the cleanest way to do it.

Actually it makes me want to change the way Population handles food consumption -- it current returns 'actual food consumed' in the update() function which should probably be a separate function. Classic case of a single function trying to do more than one thing.

PR #952, opened 2021-06-12
Main menu Button array

DanRStevens:
The main menu `Button` code is quite repetitive. Using array processing simplifies things, and ensures uniform handling.

Mind you, the always disabled "Options" button is slightly non-uniform. Maybe we could get rid of it for now?

Anyway, noticed some of the potential for cleanup after working on #951, which had a way of separating the unique portions from the repetitive portions.


ldicker83:
Definitely dig this update. Nice clean way to do a simple button menu like this.

PR #953, opened 2021-06-12
Temporarily remove the Options button from the main menu

DanRStevens:
The Options feature isn't currently implemented, so the button was set to always be disabled. Perhaps we should just remove the button for now, until the feature is implemented. At this point, it's very little code to change to add or remove a main menu button.

Relates to cleanup in #952, and feature from #951.


ldicker83:
Works for me. The previous PR makes it very easy to reintroduce the options button.

PR #968, opened 2021-06-15
Agridome pilfering

Brett208:
Working on #911 

Checking to make sure approach is sound before filling in more structures to have crime.

Added the module `CrimeExecution` to actually implement crime. This would allow triggering criminal activity in a more scripted manner if ever desired in the future and hopefully breaks the code into more manageable pieces.

I made `CrimeRateUpdate` into a class to move `moraleChange` to a separate get function based on feedback in other PRs.

If this looks good, will probably try to follow up with allowing smelters and storage tanks to be stolen from. I figured the food in the Command Center should probably be off limits from pilfering.

Right now, up to 100 food is stolen each time crime occurs. That should be updated to be a bit more of a random amount if the overall structure of the code looks good. Also, I didn't really know how to label food, so I just said units. Maybe tons of food?

FoodPilfered

### DIFF
--- a/OPHD/States/CrimeExecution.cpp
+++ b/OPHD/States/CrimeExecution.cpp
@@ -8,7 +8,7 @@
 CrimeExecution::CrimeExecution(NotificationArea& notificationArea) : mNotificationArea(notificationArea) {}
 
 
-void CrimeExecution::executeCrimes(std::vector<Structure*> structuresCommittingCrime)
+void CrimeExecution::executeCrimes(const std::vector<Structure*>& structuresCommittingCrime)
 {
 	for (auto& structure : structuresCommittingCrime)
 	{
@@ -26,13 +26,6 @@ void CrimeExecution::executeCrimes(std::vector<Structure*> structuresCommittingC
 			break;
 		}
 	}
-}
-
-
-void CrimeExecution::executeCrimes(std::vector<Structure*>&& structuresCommittingCrime)
-{
-	// Delegate to lvalue function overload
-	executeCrimes(structuresCommittingCrime);
 }
 
 
--- a/OPHD/States/CrimeExecution.cpp
+++ b/OPHD/States/CrimeExecution.cpp
@@ -22,6 +22,8 @@ void CrimeExecution::executeCrimes(std::vector<Structure*> structuresCommittingC
 		case StructureID::SID_AGRIDOME:
 			executeStealingFood(static_cast<FoodProduction&>(*structure));
 			break;
+		default:
+			break;
 		}
 	}
 }
--- a/OPHD/States/CrimeExecution.cpp
+++ b/OPHD/States/CrimeExecution.cpp
@@ -44,7 +44,6 @@ void CrimeExecution::executeStealingFood(FoodProduction& structure)
 		
 		structure.foodLevel(-foodStolen);
 
-
 		const auto& structureTile = NAS2D::Utility<StructureManager>::get().tileFromStructure(&structure);
 		
 		mNotificationArea.push("Food Stolen", NAS2D::stringFrom(foodStolen) + " units of food was pilfered.",
--- a/OPHD/States/CrimeExecution.cpp
+++ b/OPHD/States/CrimeExecution.cpp
@@ -1,0 +1,52 @@
+#include "CrimeExecution.h"
+#include "../StructureManager.h"
+#include <NAS2D/StringUtils.h>
+#include <NAS2D/Utility.h>
+#include <algorithm>
+
+
+CrimeExecution::CrimeExecution(NotificationArea& notificationArea) : mNotificationArea(notificationArea) {}
+
+
+void CrimeExecution::executeCrimes(std::vector<Structure*> structuresCommittingCrime)
+{
+	for (auto& structure : structuresCommittingCrime)
+	{
+		if (structure == nullptr) 
+		{
+			continue;
+		}
+
+		switch (structure->structureId())
+		{
+		case StructureID::SID_AGRIDOME:
+			executeStealingFood(static_cast<FoodProduction&>(*structure));
+			break;
+		}
+	}
+}
+
+
+void CrimeExecution::executeCrimes(std::vector<Structure*>&& structuresCommittingCrime)
+{
+	// Delegate to lvalue function overload
+	executeCrimes(structuresCommittingCrime);
+}
+
+
+void CrimeExecution::executeStealingFood(FoodProduction& structure)
+{
+	if (structure.foodLevel() > 0)
+	{
+		int foodStolen = std::clamp(100, 0, structure.foodLevel());
+		
+		structure.foodLevel(-foodStolen);
+
+
+		const auto& structureTile = NAS2D::Utility<StructureManager>::get().tileFromStructure(&structure);
+		
+		mNotificationArea.push("Food Stolen", NAS2D::stringFrom(foodStolen) + " units of food was pilfered.",
+			structureTile.position(),
+			NotificationArea::NotificationType::Warning);
+	}
+}
--- a/OPHD/States/CrimeExecution.h
+++ b/OPHD/States/CrimeExecution.h
@@ -10,8 +10,7 @@ class CrimeExecution
 public:
 	CrimeExecution(NotificationArea& notificationArea);
 
-	void executeCrimes(std::vector<Structure*> structuresCommittingCrime);
-	void executeCrimes(std::vector<Structure*>&& structuresCommittingCrime);
+	void executeCrimes(const std::vector<Structure*>& structuresCommittingCrime);
 
 	void executeStealingFood(FoodProduction& structure);
 
--- a/OPHD/States/CrimeExecution.h
+++ b/OPHD/States/CrimeExecution.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include "../UI/NotificationArea.h"
+#include "../Things/Structures/Structures.h"
+#include <vector>
+
+
+class CrimeExecution 
+{
+public:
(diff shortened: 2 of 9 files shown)


OPHD/States/CrimeExecution.h
```diff
@@ -0,0 +1,20 @@
...
+#include <vector>
+
+
+class CrimeExecution 
+{
+public:
+	CrimeExecution(NotificationArea& notificationArea);
+
+	void executeCrimes(std::vector<Structure*> structuresCommittingCrime);
+	void executeCrimes(std::vector<Structure*>&& structuresCommittingCrime);
```
DanRStevens, line 14: Is there any plan to update or store the contents of the passed parameter? If not, I would recommend a single overload passing by `const&`. A `const&` parameter binds to both lvalues and rvalues. Looking at the code, it appears read-only access is sufficient.

If you need write access to the object, and want to be able to accept both named variables and temporary objects, then it makes sense to have both `&` and `&&` overloads.

Here it's by-value and rvalue reference. So write access would be granted to temporaries, but named variables would be copied. I suspect this was not the intention.

OPHD/States/CrimeExecution.cpp
```diff
@@ -0,0 +1,52 @@
...
+		}
+	}
+}
+
+
+void CrimeExecution::executeCrimes(std::vector<Structure*>&& structuresCommittingCrime)
+{
+	// Delegate to lvalue function overload
+	executeCrimes(structuresCommittingCrime);
+}
```
DanRStevens, line 36: Hmm, this binds to temporaries, but then forces a copy of the temporary when passing it to the by-value overload. Could eliminated this overload get the same effect.

Though I assume this wasn't what was intended.

OPHD/States/CrimeRateUpdate.cpp
```diff
@@ -3,13 +3,24 @@
...
+#include <random>
+#include <functional>
+
+
+namespace {
+	std::random_device randomDevice;
+	std::mt19937 generator(randomDevice());
+	std::uniform_int_distribution<int> distribution(0, 1000);
+	auto random = std::bind(distribution, std::ref(generator));
+}
```
DanRStevens: I believe I once expressed some concern in the past about putting random number code in individual modules. I don't remember the exact reason. I believe it was something to do with reproducibility, or how this might impact saved games, or game recordings, or multiplayer if that is ever implemented. If random numbers are generated and used in the same order on all clients, then there is no need to synchronize multiplayer clients for each value. Instead, a random number state can be shared once at the beginning, and then all clients will generate the same values from there. Similar for game recordings that make use of random numbers. You can store the initial state once, and then generate the same sequence during playback, without having to record the outcome of every random number generated.

It may be a bit early to worry about that design aspect though. In particular, I don't have any suggestion for how to improve this. I guess at some point we might need to develop some high level policy on how random numbers should be generated and used by the game engine.

DanRStevens (commented):
General design looks good. I noticed one oddity that looks unintended.

OPHD/States/CrimeExecution.h
Brett208: Yeah, this should be const& as you point out. I can update tomorrow.

OPHD/States/MapViewState.h
```diff
@@ -246,6 +247,7 @@ class MapViewState : public Wrapper
 private:
 	MainReportsUiState& mMainReportsState;
 	TileMap* mTileMap = nullptr;
```
ldicker83, line 250: Minor note but something to consider going forward, we've been preferring brace initializers versus assignment. This is mostly a stylistic choice (I'm sure there are some differences in the generated code) but going forward brace init is the preferred style.

OPHD/States/CrimeRateUpdate.cpp
```diff
@@ -5,14 +5,16 @@
...
 
-int CrimeRateUpdate::update(const std::vector<TileList>& policeOverlays, PopulationPanel& populationPanel)
+void CrimeRateUpdate::update(const std::vector<TileList>& policeOverlays, PopulationPanel& populationPanel)
 {
+	mMoraleChange = 0;
+
 	const auto& structuresWithCrime = NAS2D::Utility<StructureManager>::get().structuresWithCrime();
 
 	// Colony will not have a crime rate until at least one structure that supports crime is built
 	if (structuresWithCrime.empty()) {
```
ldicker83, line 29: Another minor nitpick, the general styling here should be brace placement on a new line. Not going to complain too loudly but this is the preferred style choice for the project and for the sake of consistency. :D

OPHD/States/CrimeRateUpdate.h
```diff
@@ -9,10 +9,13 @@ class Structure;
 class CrimeRateUpdate
 {
 public:
-	// Returns change in current morale caused by colony's mean crime rate
-	int update(const std::vector<TileList>& policeOverlays, PopulationPanel& populationPanel);
+	void update(const std::vector<TileList>& policeOverlays, PopulationPanel& populationPanel);
+
+	int getMoraleChange() const { return mMoraleChange; }
```
ldicker83, line 16: I'd like @DanRStevens to weigh in here -- I know we've talked about verb-based function names but my general choice for most of the code for getters/setters is to drop the get/set prefix in favor of something like `int moraleChange() const 
 { return mMoraleChange; }`, basically treating it as a property versus a get/set type function.

I'm not going to be overly picky about it going forward but I would like to make a decision as to the preferred naming scheme here.

Something like the below `isProtectedByPolice` makes perfect sense versus `setPopulationPanel` which would probably be better served as `updatePopulationPanel`.

Thoughts?

OPHD/States/CrimeRateUpdate.cpp
ldicker83: Agreed that this is a bit of a tough one mostly from the aspect of multiplayer. I've considered a multiplayer aspect but when I really think about how it would be done in this game, I'm not sure it would actually be 'fun' so I haven't done much with it.

That stated, this is fine for me. There's a bit of magic numbering going on here, something I've tried to mitigate or at least note when they show up. Besides that this is fine.

OPHD/States/CrimeRateUpdate.cpp
```diff
@@ -27,6 +38,13 @@ void CrimeRateUpdate::update(const std::vector<TileList>& policeOverlays)
 		int crimeRateChange = isProtectedByPolice(policeOverlays, structure) ? -1 : 1;
 		structure->increaseCrimeRate(crimeRateChange);
 
+		// Crime Rate of 0% means no crime
+		// Crime Rate of 100% means crime occurs 10% of the time
+		if (structure->crimeRate() + random() > 1000)
```
ldicker83: Some magic numbers going on here that aren't documented -- not super picky about it but something to be aware of.

This is probably a good place to have difficulty modifiers (e.g., easy might be 5%, hard might be 20%) though this is probably beyond the scope of this PR and change set.

OPHD/States/CrimeRateUpdate.cpp
```diff
@@ -27,6 +38,13 @@ void CrimeRateUpdate::update(const std::vector<TileList>& policeOverlays)
...
+		{
+			mStructuresCommittingCrimes.push_back(structure);
+		}
+
```
ldicker83, line 47: I hadn't thought of doing it this way... but generally I like it.

I recently added a crime icon with the thinking that structures that have high crime rates (or need maintenance) could have the icon displayed over top of them. In this case it would be displayed where crime is actually happening versus just the crime rate itself being high. I like it, simplifies that part of it a bit and boils it down to individual criminal acts that can also push notifications (you show an example of that which I really like).

Nice work!

OPHD/States/CrimeRateUpdate.h
DanRStevens: I have pretty mixed feelings on the naming convention. I do sort of like the shorter property naming convention, and was never very fond of "get" and "set", though C++ doesn't have property syntax support, so property uses are indistinguishable from normal function calls. That can be an issue because some noun phrases (properties) are easily confused with verb phrases (function calls), so it becomes very easy to mistake what a method does.

I think I would prefer verb phrases for functions and methods.

----

The `setPopulationPanel` method could probably have been named a bit better.

OPHD/States/CrimeRateUpdate.cpp
Brett208: I knew this was the general style you were going for, but I guess I forgot or missed these lines. Will fix momentarily.

OPHD/States/CrimeRateUpdate.cpp
Brett208: I would like to merge without resolving this issue. I did open issue #974 to track. I may be able to take a stab at resolving this Random Number Generator issue after this PR closes.

I would approach it as unifying the location of the random number generator in the code base more so than prepping for multiplayer.

OPHD/States/CrimeRateUpdate.cpp
Brett208: Glad the module separate is appreciated. Thank you.

I would encourage highlighting high crime rate as you were planning. I sort of like the idea of also marking structures that recently had major criminal activity if it wasn't too busy. I think highlighting high crime structures might be a bit higher priority since the message log will tell the player when specific criminal activity occurs.

OPHD/States/CrimeRateUpdate.cpp
Brett208: Yeah, agree on the difficulty modifier being a task here. I'll add it to the umbrella Implement Crime task.

Brett208:
I tried to address the bulk of the issues brought up by review. Thank you @ldicker83 and @DanRStevens for the thorough review of the code. I think it is better for the changes.

PR #977, opened 2021-06-17
Convert constants to constexpr

ldicker83:
Does a bit of refactoring in naming schemes and definition of constants prefering constexpr where applicable.

This is a bit of a poorly named branch because it's actually a refactor more than anything else.

Also, this is going to seem like a HUGE changeset, but it really isn't. Lots of lines changed to reflect improved naming conventions.


Brett208:
I appreciate pulling the underscores from the naming convention here. :)

ldicker83:
> I appreciate pulling the underscores from the naming convention here. :)

Woot!

Yeah, it's been bugging me that these have sat around like this for a long time after we decided that we weren't going to use ALL_CAPS_NAMES like that anymore.

Once the checks finish I'm going to merge into master and call it a night. :D

OPHD/Constants/Numbers.h
```diff
@@ -8,18 +8,18 @@
  */
 namespace constants
 {
-	const int FADE_SPEED = 300;
+	inline constexpr int FADE_SPEED{ 300 };
```
DanRStevens: Btw, `inline` is implied by `constexpr`. You only need to specify `constexpr`.

DanRStevens (commented):
Yes!

I am so happy to see the CamelCase change.

OPHD/Constants/Numbers.h
ldicker83: That's not what the article I was reading said, but reading the actual standard does confirm this. Wherps. Oh well, I'll worry about it another time. :D

PR #978, opened 2021-06-17
Allow setting difficulty to increase or decrease the rate at which crimes occur

Brett208:
Related to issue #911.

Just realized the crime issue was literally the number 911...

So on Normal chance of crime occurring scales linear from 0 to 10% chance each turn. On Hard, the scale is doubled, so max of 20% chance of crime. On Easy and Beginner, it is reduces to 7.5% and 5% respectively. I'm not beholden to any specific number, just tried to pick somewhat sensible starting places.


DanRStevens:
> Just realized the crime issue was literally the number 911...

Hah, nice! :smile:

Issue #974, opened 2021-06-15
Unify Random Number Generation


Brett208:
Create a static module that generates random numbers. Pull standalone random number generators out of the various locations in the code base.

This will allow tweaking the random number generator in the future without having to edit disparate parts of the code base. There may be some minor performance gains.

Brett208:
population.cpp uses `default_random`. If this code is pulled and the new RandomNumberGenerator used, it will be implemented with `mt19937`, I think this is OK to switch out, but wanted to check if there was a specific reason for the split in the code base with some areas using the default and others specifying mt19937.

In population.cpp:
```C++
std::default_random_engine pop_generator;
std::uniform_int_distribution<int> pop_distribution(0, 100);
```

Much appreciate for @DanRStevens for taking the time to fix the deficient template function I wrote to close this issue out.

Brett208:
Random number generation in TileMap.cpp should be updated to use the new random number generator. It will take a bit of work to refactor:

```C++
std::random_device rd;
std::mt19937 generator(rd());
std::uniform_int_distribution<int> distributionWidth(5, MAP_WIDTH - 5);
std::uniform_int_distribution<int> distributionHeight(5, MAP_HEIGHT - 5);

auto mwidth = std::bind(distributionWidth, std::ref(generator));
auto mheight = std::bind(distributionHeight, std::ref(generator));
auto randPoint = [&mwidth, &mheight]() { return NAS2D::Point{mwidth(), mheight()}; };

auto generateMines = [&](int mineCountAtYield, MineProductionRate yield) {
	for (int i = 0; i < mineCountAtYield; ++i) {
		addMineSet(randPoint(), mMineLocations, yield);
	}
};
```

DanRStevens:
I think it's fine to switch the random number generator. We probably do want to be specific about it. The default is implementation defined, so there's no guarantee of consistency between platforms, or builds with different compilers. That could be an issue later on if we implement game recording or multiplayer, both of which would require consistent repeatable random numbers.

ldicker83:
`mt19937` is the preferred method, it tends to produce high quality values.

Was this intended to be closed with the associated PR?

Brett208:
Yes it was. All random number generation has been consolidated into the new RandomNumberGenerator. You should be able to use it now to generate random numbers as needed.

PR #980, opened 2021-06-18
First draft of a random number generator

Brett208:
Does not work yet

Related to #974 

```C++
randomNumber.generate<int>(0, 1000);
```

Produces the error:

Error (active)	E0304	no instance of function template "RandomNumberGenerator::generate" matches the argument list	   
    argument types are: (int, int)
    object type is: RandomNumberGenerator

I think the template is written incorrectly?


DanRStevens (commented):
With a few slight changes, I was able to get the following to compile:
```cpp
RandomNumberGenerator generator;
auto number = generator.generate<int>(0, 1000);
```

Brett208:
Hey, thanks for the help! I'm showing my lack of skill with template programming and reading online documentation...

I'm assuming since there were not any comments that this looks like a reasonable starting approach. I'll plan to cleanup and move the PR out of draft mode.

DanRStevens:
Yes, it looked reasonable.

When idea I was toying with, was to have an overload where the `min` and `max` values were template value parameters (as opposed to either template type parameters, or normal function parameters). That would allow the `min` `max` range check to be done at compile time. Of course, such an overload would only be usable when the range was a fixed compile time value.

Brett208:
Hmm, I think maybe I should ditch the template for this implementation. Maybe I somehow committed the suggested changes incorrectly?

OPHD/RandomNumberGenerator.h
```diff
@@ -0,0 +1,42 @@
...
+#include <random>
+
+
+class RandomNumberGenerator
+{
+public:
+	RandomNumberGenerator() : generator(randomDevice()) {};
+
+	template<typename T, std::enable_if_t<std::is_arithmetic_v<T>>>
+	T generate(T min, T max)
```
DanRStevens, line 14: This is one way to fix that problem:
```cpp
	template<typename T, typename = std::enable_if_t<std::is_arithmetic_v<T>>>
	T generate(T min, T max)
```

I may have forgotten to include this in my previous review. I had made the change temporarily, but thought I had removed it afterwards, which is why I didn't include it as part of the review.

I had also experimented with using `std::enable_if_t` on the return type, rather than a template parameter. Normally that's what I've done in the past. I used to avoid extra template type parameters, since that means the user could potentially specify that extra parameter at the point of call:

It weird, but valid to include an explicit type for the second type parameter, such as `void`, or `int`, using the above syntax for the function definition.
```cpp
randomNumber.generate<int, void>(0, 1000)
```

If you want to avoid that possibility, you can place `std::enable_if_t` on the return type instead:
```suggestion
	template<typename T>
	std::enable_if_t<std::is_arithmetic_v<T>, T>
	generate(T min, T max)
```

Note that you must be explicit with the second type parameter to `std::enable_if_t`, since the default is `void`, and we want the return type to be `T` rather than `void`.

Brett208: I figured specifying `template<typename T, std::enable_if_t<std::is_arithmetic_v<T>>>` covered all instances of T in the function declaration. Based on what you are saying, I'm thinking that this is not the case and it only covers the 2 arguments min and max.

I remember reading long ago that some of the template material was embedded in the function declaration. I was wondering if that affected the syntax here?

Brett208:
Ok, I think this branch is ready for merging now. Thank you Dan for helping out.

I pointed out in the issue supporting this PR 2 other places where the random number generator is not being used, but I think we can work those in followon branches if this looks good.

OPHD/RandomNumberGenerator.h
DanRStevens: Using `std::enable_if` is a bit weird. It basically works by creating invalid syntax when something should not be enabled, and relying on SFINAE (Substitution Failure Is Not An Error), which ignores the badly formed template function rather than stopping with a compile error. If the expression passed to `std::enable_if` is false, then there is no `::type` field in the generated struct, and so trying to access it will be invalid code. In particular, the `std::enable_if_t` alias has a built in access for the `::type` field.

Based on how that all works, `std::enable_if` disables the function as a whole when the expression is `false`. It's as if the compiler doesn't even see that method for types that produce a `false` expression.

This also means that `std::enable_if` has to be used where a type is expected. Hence people often put it in the template type parameter list, often as some kind of default. It can also be placed on the return value, or on any of the function parameters. It's a bit weird to use on function parameters though, since they vary in number, and there's no natural choice for which parameter to put it around. The return type is a lot more natural since there is always exactly one, even if it's declared to be `void`. Putting it in the template type parameter list is perhaps more natural for some people, since it means the function signature looks normal, but then there are implications when giving explicit type parameters at the point of call.

My personal preference is generally to use the return type.

As a bit of an aside, you can also use the trailing return type syntax with `auto` and `->`, and put `std::enable_if` on a return type listed after the function parameter list.

```cpp
// Without std::enable_if_t
template<typename T>
auto generate(T min, T max) -> T
```

```cpp
// Guarded with std::enable_if_t
template<typename T>
auto generate(T min, T max) -> std::enable_if_t<std::is_arithmetic_v<T>, T>
```

DanRStevens (approved):

ldicker83:
> I pointed out in the issue supporting this PR 2 other places where the random number generator is not being used, but I think we can work those in followon branches if this looks good.

Agreed

Issue #588, opened 2020-08-09
Crash on turn update after using refined resources cheat


ldicker83:
I discovered that with the switch to StorableResources, using the refined resources 'cheat' causes an exception of "incorrect type/value" to be thrown during turn processing. This is related to handling of resource movement and capping resource to what's actually able to be stored and the std::clamp function (it doesn't like negative minimum values).

So that stated, instead of just adding 1000 to each refined resource type, the fix should involve simply capping the resource at their maximum values for the given storage.

PR #982, opened 2021-06-19
Crime execution improvements

Brett208:
Adds stealing from smelters and storage tanks.

Until #974 is resolved, the way resources are stolen will be fairly uninteresting.

One may be able to argue that stealing raw materials is a bit silly...

Resources Stolen

ResourcesStolenDetails


OPHD/States/CrimeExecution.h
```diff
@@ -10,10 +11,13 @@ class CrimeExecution
...
 	CrimeExecution(NotificationArea& notificationArea);
 
+	void difficulty(Difficulty difficulty) { mDifficulty = difficulty; }
+
 	void executeCrimes(const std::vector<Structure*>& structuresCommittingCrime);
 
 	void executeStealingFood(FoodProduction& structure);
 
 private:
+	Difficulty mDifficulty{ Difficulty::Medium };
```
DanRStevens, line 24: I feel like this should probably be a constructor parameter.

OPHD/States/CrimeExecution.cpp
```diff
@@ -44,3 +47,29 @@ void CrimeExecution::stealFood(FoodProduction& structure)
...
+{
+	if (structure.storage().isEmpty())
+	{
+		return;
+	}
+
+	auto resourceIndicesWithStock = structure.storage().getIndicesWithStock();
+
+	// TODO: Rob randomly from stock, for now it just robs from first available indice with resources
+	auto indexToStealFrom = resourceIndicesWithStock[0];
```
DanRStevens, line 78: This is quite reasonable for initial review. I'd probably want to fix it before merging though.

OPHD/States/CrimeExecution.cpp
```diff
@@ -44,3 +47,29 @@ void CrimeExecution::stealFood(FoodProduction& structure)
...
+
+	int amountStolen = std::clamp(100, 0, structure.storage().resources[indexToStealFrom]);
```
DanRStevens, line 80: You may want to check the parameter order of [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp). The value comes first. Using an alternate parameter order may appear to work in some cases, but can be problematic and throw exceptions when `low >= high`. That can potentially happen on rare occasions when variable values are passed for either or both of the later two parameters.

OPHD/States/CrimeExecution.h
DanRStevens: Actually, make it a constructor parameter, and declare it as `const`, so we know if it was ever not set, as unset `const` values will produce a compiler warning.

I don't think it makes sense to change that during a game.

ldicker83:
Yikes, 100 units of refined resources? That's like... 100 truck loads of materials lost 0.0

OPHD/States/CrimeExecution.h
Brett208: This was my initial design. C++ makes it difficult though. `MapViewState` is constructed before saved games are loaded. So, to pass `CrimeExecution` the difficulty setting in the constructor initialization list of `MapViewState`, we would probably need to load the saved game before constructing `MapViewState`. Or make `CrimeExecution` a `unique_ptr` so it can be initialized after MapViewState is constructed. 

It is not uncommon to allow a player to change difficulty settings mid play-through in some games, wasn't sure what Leeor's plan was here.

Brett208: Oh, I don't think difficulty is actually set in game yet, so sort of academic in some respects for me to be adding it to crime. I wasn't ready to add a GUI setting for difficulty on the menu if someone else wanted to tackle that.

OPHD/States/CrimeExecution.cpp
Brett208: I think this is correct. The goal is to steal 100 resources, but not greater than the stored resources. 

Maybe easier to read if I set amountStolen to 100, then pass it into clamp? This is a little arbitrary, since I want to use the random number generator to make the amount stolen variable in the future.

OPHD/States/CrimeExecution.cpp
Brett208: Sounds good. Thanks for unblocking the RandomNumberGenerator. I work on getting it finished tomorrow to unblock this PR.

Brett208:
@ldicker83, good point. I hadn't thought of the balance yet. What do you think about RANDOM between 1-5 resources then? Maybe make it a bit higher on hard difficulty.

Also, the Agridome losing 100 food is probably too high as well now that I think about it.

-Brett

OPHD/States/CrimeExecution.cpp
DanRStevens: Logically that should be equivalent to stealing all available resources, but clamping the value at 100. That would be safer, considering how `std::clamp` is implemented.

> Reference to lo if v is less than lo, reference to hi if hi is less than v, otherwise reference to v.

> The behavior is undefined if the value of lo is greater than hi.

The constants should be `lo` and `hi`. It's a safer way to use the function.

See bug #588, and fixes #589, and #590.

In particular, take a look at what the fix was:
https://github.com/OutpostUniverse/OPHD/pull/589/commits/27ddeb366cf47d4f67d208f1f8efa4aed0a9b05b

OPHD/States/CrimeExecution.h
DanRStevens: Actually, you bring up a good point. I think the way new games versus loaded games are being initialized needs to be refactored. There's a lot of modification of effectively static values after construction due to the way the current code is structured. It's one of those things I've wanted to address, but never had the time. I suspect it will be quite a big ball to unravel.

Anyway, I suppose we have to stick with something that works with the current design.

ldicker83:
> @ldicker83, good point. I hadn't thought of the balance yet. What do you think about RANDOM between 1-5 resources then? Maybe make it a bit higher on hard difficulty.

> Also, the Agridome losing 100 food is probably too high as well now that I think about it.

1 - 5 works for me, that makes a lot more sense. Food could probably be a bit more, like 10 - 15 units per event and more for higher difficulties. These would probably be accounted for as organized criminal activity and underground / black-market transfer of goods.

OPHD/States/CrimeExecution.cpp
Brett208: I guess my assumption was the amount in storage was always 0 or positive. If the amount in storage is negative, it would maybe be good to throw an exception? 

I'm pulling std::clamp since it is fairly easy to implement without the standard library call, which will remove the contention.

DanRStevens: I suppose my other concern is the code will serve as an example to be copied later, where assumptions will be different, and potentially more dangerous.

As for throwing an exception, this would be the wrong place for that exception. It would be confusing to get the exception here, rather than the code location that sets the negative value. From that perspective, your code is absolutely correct to assume the value should be non-negative. If we throw an error, it should be done elsewhere. Throwing the exception here would lead to a confusing debugging experience.

DanRStevens:
Anyway, I'm fine with you using `std::clamp`. I just think constant clamp values should be the 2nd and 3rd parameter, while the variable value should be the 1st parameter. That will always be safe, regardless of range. (Unless the constant values are wrong, in which case it's always wrong, so you'll know very quickly).

Brett208:
@ldicker83, I'm hoping this was closed without merge by accident? I was planning on pulling in the random number generator before merging as recommended in some of the comments. However, if your intention was to merge as is, I'm happy adding the random number generator in a separate PR, just don't want to see these changes lost.

-Brett

DanRStevens:
Hmm, odd. There was no comment, so I assume that was done by accident?

Brett208:
OK, I'm ready to merge this branch unless there are any remaining issues someone else notes. I would like to make the actual resource quantities stolen random but that can be in a separate PR.

Below is an shot of rare metals being stolen during play testing.

RareMetals

ldicker83:
> @ldicker83, I'm hoping this was closed without merge by accident? I was planning on pulling in the random number generator before merging as recommended in some of the comments. However, if your intention was to merge as is, I'm happy adding the random number generator in a separate PR, just don't want to see these changes lost.
> 
> -Brett

I don't remember closing this so I probably hit something by mistake. Whoops? I might have hit Close with comment by mistake.

DanRStevens:
Blame it on a cat jumping on your keyboard while you weren't looking :wink:

PR #983, opened 2021-06-19
Show breakout of which resources are stored within a Storage Tanks

Brett208:


Brett208:
StorageTanksInspector

OPHD/Things/Structures/StorageTanks.h
```diff
@@ -26,10 +26,27 @@ class StorageTanks : public Structure
...
+		StringTable stringTable(2, 5);
+
+		stringTable.setColumnText(
+			0,
+			{ 
+				"Storage Capacity",
+				ResourceNamesRefined[0],
+				ResourceNamesRefined[2],
+				ResourceNamesRefined[1],
+				ResourceNamesRefined[3],
```
DanRStevens, line 38: This use of out or order indexes concerns me. Is there a way we can fix this? I think I've seen this elsewhere in existing code too. Maybe we should re-order what the values mean if we typically use the values in a different order.

DanRStevens (approved):

PR #984, opened 2021-06-19
Fix route cost constant error

### LINKED ISSUE #981: Smelters are not refining ore
When starting a new game on mars, and building a mine, the seed smelter is not processing raw resources. I'm unsure if this is a trucking route issue or a smelter issue? I can provide a saved game if needed, but it seems reproducible quickly each game I start.

Build: Windows x64 debug, 18 June 21 master branch

ldicker83:
Closes #981

### DIFF
--- a/OPHD/Map/TileMap.cpp
+++ b/OPHD/Map/TileMap.cpp
@@ -577,7 +577,7 @@ void TileMap::AdjacentCost(void* state, std::vector<micropather::StateCost>* adj
 		}
 
 		auto& adjacentTile = getTile(position, 0);
-		float cost = constants::RobotCommRange;
+		float cost = constants::RouteBaseCost;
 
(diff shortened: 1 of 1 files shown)


DanRStevens (approved):
Well that would do it.

If only there was an easy way to create new types that were still effectively usable as the base type in C++, but not interoperable with other types. Sadly `using` aliases don't suffice, as they don't actually create a new type, just a new alias, so the new name is still interoperable with the base type and all it's other aliases.

ldicker83:
> Well that would do it.
> 
> If only there was an easy way to create new types that were still effectively usable as the base type in C++, but not interoperable with other types. Sadly `using` aliases don't suffice, as they don't actually create a new type, just a new alias, so the new name is still interoperable with the base type and all it's other aliases.

? There's no using alias involved here? This was a `constexpr` and I just accidentally used the wrong constant.

Issue #987, opened 2021-06-20
Windows compile warnings


DanRStevens:
I've noticed a few new MSVC compiler warnings, which seem related to the new Crime update. They use the code analysis style of text, so I assume they aren't visible in a normal developer compile.

> C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\utility(193,54): warning C4244: 'initializing': conversion from '_Ty' to '_Ty2', possible loss of data [C:\projects\ophd\OPHD\ophd.vcxproj]
>           with
>           [
>               _Ty=double
>           ]
>           and
>           [
>               _Ty2=float
>           ] (compiling source file States\CrimeRateUpdate.cpp)
> C:\projects\ophd\OPHD\States\CrimeRateUpdate.h(31): message : see reference to function template instantiation 'std::pair<const Difficulty,float>::pair<Difficulty,double,0>(_Other1 &&,_Other2 &&) noexcept' being compiled [C:\projects\ophd\OPHD\ophd.vcxproj]
>           with
>           [
>               _Other1=Difficulty,
>               _Other2=double
>           ] (compiling source file States\CrimeRateUpdate.cpp)
> C:\projects\ophd\OPHD\States\CrimeRateUpdate.h(26): message : see reference to function template instantiation 'std::pair<const Difficulty,float>::pair<Difficulty,double,0>(_Other1 &&,_Other2 &&) noexcept' being compiled [C:\projects\ophd\OPHD\ophd.vcxproj]
>           with
>           [
>               _Other1=Difficulty,
>               _Other2=double
>           ] (compiling source file States\CrimeRateUpdate.cpp)
> C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\utility(193,54): warning C4244: 'initializing': conversion from '_Ty' to '_Ty2', possible loss of data [C:\projects\ophd\OPHD\ophd.vcxproj]
>           with
>           [
>               _Ty=int
>           ]
>           and
>           [
>               _Ty2=float
>           ] (compiling source file States\CrimeRateUpdate.cpp)
[...]

PR #988, opened 2021-06-20
Explicitly define difficulty modifiers as floating point

ldicker83:
Fixes a type mismatch warning by explicitly declaring difficulty modifiers in Crime module as floating point.


DanRStevens:
I suspect this was the change that addressed issue #987.

ldicker83:
Yus. There as unintended implicit type conversions going on. It's why I recommend using 'f' in floating point expressions to ensure that this doesn't become an issue.

PR #994, opened 2021-06-24
Implement maintenance logic

### LINKED ISSUE #22: Implement maintenance logic
After #21 is complete, the logic for structure maintenance needs to be added.

For basic maintenance with a Structure still in an operational state, an engineering team takes 1 turn to complete maintenance. This will bring a structure's Integrity back up to 100%.

Structures that are disabled due to low integrity take 2 turns to repair.

Destroyed structures cannot be repaired. These must be demolished and rebuilt.

ldicker83:
Closes #22

### DIFF
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -542,6 +542,12 @@ void MapViewState::checkNewlyBuiltStructures()
 }
 
 
+void MapViewState::updateMaintenance()
+{
+
+}
+
+
 void MapViewState::updateOverlays()
 {
 	checkCommRangeOverlay();
@@ -591,6 +597,7 @@ void MapViewState::nextTurn()
 	countFood();
 	updatePopulation();
 
+	updateMaintenance();
 	updateCommercial();
 	updateBiowasteRecycling();
 	updateMorale();
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -544,9 +544,9 @@ void MapViewState::checkNewlyBuiltStructures()
 
 void MapViewState::updateMaintenance()
 {
-	auto sortLambda = [](const Structure* rhs, const Structure* lhs) -> bool
+	auto sortLambda = [](const Structure* lhs, const Structure* rhs) -> bool
 	{
-		return rhs->integrity() < lhs->integrity();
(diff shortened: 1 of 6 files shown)


OPHD/StructureManager.cpp
```diff
@@ -339,6 +339,20 @@ const StructureList& StructureManager::structureList(Structure::StructureClass s
...
 
+StructureList StructureManager::allStructures()
+{
+	StructureList structuresOut;
+
+	for(auto& structures : mStructureLists)
+	{
+		auto& list = structures.second;
+		std::copy(list.begin(), list.end(), std::back_inserter(structuresOut));
+	}
```
DanRStevens, line 350: Should probably avoid calling the loop variable `structures` if it isn't a collection of structures. It's maybe a bit misleading, since it's actually a `std::pair`, where the `second` element is the list of structures.

We should maybe come up with a convention for naming `std::pair` objects in range-for loops. I don't object to `keyValuePair`, though it is a bit generic. Could maybe name it something more specific, like `structureClassStructuresPair`, but that gets a bit wordy.

The `list` variable could then be named `structures`, which would be a bit more descriptive.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -544,7 +544,20 @@ void MapViewState::checkNewlyBuiltStructures()
 
 void MapViewState::updateMaintenance()
 {
+	auto sortLambda = [](const Structure* rhs, const Structure* lhs) -> bool
+	{
+		return rhs->integrity() < lhs->integrity();
+	};
```
DanRStevens, line 550: These variables are backwards. You have `rhs` on the left hand side, and `lhs` on the right hand side. That makes reading the function implementation a bit confusing.

I assume the intention is to sort structures such that the lowest integrity is listed first.

DanRStevens (approved):
Looks good.

Noted a couple of variable name oddities.

OPHD/States/MapViewStateTurn.cpp
ldicker83: You are not wrong here. It does in fact sort it correctly but I think that's only because of the parameter order. I'll fix that in a commit here shortly.

OPHD/StructureManager.cpp
ldicker83: Fair point. I'll fix this with the below comment as well.

Issue #200, opened 2020-02-09
Decide how to implement difficulty selection


ldicker83:
So I see that @cugone is beginning to work on the difficulty implementation. Thank you for your efforts! It's very much appreciated especially while I'm recovering from illness.

---

So the point of this topic is to decide exactly how to select difficulty. The 'planet select' screen is sort of a stub screen at this point. The difficulty selection that's been presented is very much like the original game, something I'd hoped to avoid.

What I'm envisioning is a game setup screen that present all of these options as a sort of full-screen UI. I'll finish fleshing this out later today...

DanRStevens:
Yeah, I'm finding the new popup slows down the process of creating a new game.

I think it might flow smoother if the difficult selection and the planet selection were done on the same screen. Potentially you could skip selecting a difficulty setting, and just have it use a default value, making the selection optional. That way you still only need one click (on the planet type) to start a new game.

I suppose the initial default should be "easy". If the player changes the difficulty, perhaps it can be remembered for future games.

As for display layout, maybe place a horizontal list of difficulty settings above the list of planets. Since clicking the planet starts the game, it should be below any optional settings.

cugone:
I agree, moving the difficulty options to the `PlanetSelectionState` would be better.

The differences between each difficulty aren't totally explained even in the original. The only thing I've been able to determine is the frequency of disasters. Beginner having none and each difficulty afterwards increasing the frequency.

DanRStevens:
Maybe we can eventually add some sort of mouse-over popup that summarizes what the difficulty setting changes.

ldicker83:
I thought I posted this already but anyway, here goes.

The idea here is to make this as visual as possible with plenty of details, the sort of thing that modern games do. Attached is an archive with a very basic visual layout that I'm considering. Note that color scheme isn't set in stone -- I've been using the black/green as a sort of nod to the original (see the fullscreen UI's) but it's definitely not set in stone.

I didn't note what the images are exactly but basically the first set is some sort of icon that indicates visually a difficulty level (I'm no artist so I leave that up to someone better suited to that to offer ideas), the next set I think is for just the planets? I'm not sure a visual for stars is really necessary.

The final one shows the planet selection rotating.

[...]

PR #995, opened 2021-06-24
Add difficulty to save file

Brett208:
Backend work for #200 

Enables changing difficulty for #911 (Crime) if you manually set it in the saved file.

### DIFF
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -22,6 +22,21 @@
 using namespace NAS2D;
 
 
+std::string difficultyString(Difficulty difficulty)
+{
+	auto it = difficultyTable.value_comp();
+	for (auto difficultyPair : difficultyTable)
+	{
+		if (difficultyPair.second == difficulty)
+		{
+			return difficultyPair.first;
+		}
+	}
+
+	throw std::runtime_error("Provided difficulty does not exist in the difficultyMap");
+}
+
+
 const std::map<StructureState, Color> STRUCTURE_COLOR_TABLE
 {
 	{ StructureState::UnderConstruction, Color{150, 150, 150, 100} },
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -24,8 +24,7 @@ using namespace NAS2D;
 
 std::string difficultyString(Difficulty difficulty)
 {
-	auto it = difficultyTable.value_comp();
-	for (auto difficultyPair : difficultyTable)
+	for (const auto& difficultyPair : difficultyTable)
 	{
 		if (difficultyPair.second == difficulty)
 		{
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -26,7 +26,7 @@ enum class Difficulty
 
 inline const std::map<std::string, Difficulty> difficultyTable
 {
-	{"Beginner", Difficulty::Beginner},
+	{ "Beginner", Difficulty::Beginner },
 	{ "Easy", Difficulty::Easy },
 	{ "Medium", Difficulty::Medium },
 	{ "Hard", Difficulty::Hard }
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -24,7 +24,7 @@ enum class Difficulty
 };
 
 
-inline std::map<std::string, Difficulty> difficultyTable
+inline const std::map<std::string, Difficulty> difficultyTable
 {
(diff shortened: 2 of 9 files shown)


OPHD/States/MapViewStateIO.cpp
```diff
@@ -135,6 +136,19 @@ void MapViewState::save(const std::string& filePath)
 }
 
 
+void MapViewState::serializeProperties(NAS2D::Xml::XmlElement* element, const Planet::Attributes& planetAttributes)
+{
+	XmlElement* properties = new XmlElement("properties");
+	element->linkEndChild(properties);
```
DanRStevens: External concern to this PR, though I wanted to note it may be better to return the new `XmlElement` from the function, and perform the `linkEndChild` at the point of call, rather than passing in an `XmlElement` that is never used for any other purpose.

That may also play better with some of the XML support code in NAS2D, such as `dictionaryToAttributes`, which generates and returns an `XmlElement`, with a `Dictionary` of key/value pairs added as XML attributes.

OPHD/XmlSerializer.h
```diff
@@ -27,6 +28,19 @@ T stringToEnum(const std::unordered_map<std::string, T>& table, std::string valu
...
+T stringToEnum(const std::map<std::string, T>& table, std::string value)
+{
+	auto it = table.find(value);
+	if (it != table.end())
+	{
+		return it->second;
+	}
+
+	throw std::runtime_error("Unable to parse enum with value of " + value);
+}
```
DanRStevens, line 41: Slightly irrelevant observation, but this basically implements [`std::map::at`](https://en.cppreference.com/w/cpp/container/map/at), but with a better exception message. I don't see any way to meaningfully improve this though, nor use `std::map::at` and still get a better exception message.

OPHD/Common.h
```diff
@@ -24,6 +24,18 @@ enum class Difficulty
 };
 
 
+inline std::map<std::string, Difficulty> difficultyTable
+{
+	{"Beginner", Difficulty::Beginner},
+	{ "Easy", Difficulty::Easy },
+	{ "Medium", Difficulty::Medium },
+	{ "Hard", Difficulty::Hard }
+};
```
DanRStevens, line 33: Whitespace is a bit inconsistent between the first entry and the last 3 entries.

----

Should probably mark this as `constexpr` (or at least `const`). Using `constexpr` could replace the use of `inline`.

OPHD/Common.cpp
```diff
@@ -22,6 +22,21 @@
 using namespace NAS2D;
 
 
+std::string difficultyString(Difficulty difficulty)
+{
+	auto it = difficultyTable.value_comp();
```
DanRStevens: Was going to comment, this doesn't appear to be used anywhere, though it seems the Linux builds caught that already.

Also the name seems misleading as to what [`std::map::value_comp`](https://en.cppreference.com/w/cpp/container/map/value_comp) does. In particular, it returns a comparator, but the variable name would seem to imply an iterator. At any rate, I just learned about `value_comp`, which I didn't know existed before.

DanRStevens (approved):
Looks good. Thank you for taking care of the serialization refactor. The updated code makes a bit more sense now.

OPHD/Common.h
DanRStevens: It seems `constexpr` is not possible here, since the type is considered non-literal. Not sure if that's because of the `std::map` or because of the `std::string`. I doubt it's because of the `enum`.

> OPHD/States/../Common.h:27:45: error: constexpr variable cannot have non-literal type 'const std::map<std::string, Difficulty>' (aka 'const map<basic_string<char>, Difficulty>')
> constexpr std::map<std::string, Difficulty> difficultyTable

I went with `inline const` instead.

Brett208: Yeah, std::string is a non-literal. Not sure about std::map, but you definetly cannot declare any container of std::strings as constexpr from experience.

OPHD/Common.cpp
Brett208: Yeah, good call on the name. I was trying to use it as an iterator originally and didn't think about the name being off after I got it working. This function drove me to std::map instead of std::unordered_map since value_comp doesn't exist on unordered_map. I'll plan to fix the variable name in a different PR.

Brett208:
Thank you for pushing the fixes. Agree about all the comments.

Issue #996, opened 2021-06-24
Allow Removing population manually


Brett208:
Currently the Population module only allows adding population outside of standard turn calculations. 

In order to enable various disasters and criminal activity, it would be nice to add a public `removePopulation` function.

The private function series below are difficult to read through. I'm worried about adding `removePopulation` some cleanup work is done to make it easier to understand the logic in these functions among maybe some other cleanup in population.cpp. Particularly how deaths are calculated and reported on.

```C++
void kill_children(int morale, int nurseries);
void kill_students(int morale, int hospitals);
void kill_adults(Population::PersonRole role, int morale, int hospitals); 
```

This is a partial block for #911

DanRStevens:
Oops, accidentally tagged the wrong issue from my PR.

DanRStevens:
Tagging PR #1076 as related. See the `addPopulation` and `operator+=` methods.

Also, more generally, work related to #1020 may be related.

belgianguy:
@DanRStevens:  I think the new Cheat Menu #1226  could be a good location for this functionality as it also allows to add specific population demographics. PopulationTable has += and -= overrides so should be pretty straightforward, although I don't know if the -= override should prevent negative population or whether that's polluting the -= override function expectations.

![image](https://user-images.githubusercontent.com/3091512/188013851-7a9eaeee-906d-4af3-a42d-b0409b3e705f.png)

vs

![image](https://user-images.githubusercontent.com/3091512/188014504-0daae156-4650-477e-9d9f-6030f9e9fbe1.png)

DanRStevens:
The resource handling code ran into a similar issue. Take a look at `StorableResources::cap`. Effectively you can do `min`/`max`/`clamp` operations on the compound object before passing it to `operator-=`.

Actually, I'm just noticing, it seems `PopulationTable` already has support for this operation:
```cpp
const auto deaths = (mPopulationDeath / divisor).cap(mPopulation);
mPopulationDeath = mPopulationDeath % divisor;

mPopulation -= deaths;
mDeathCount += deaths.size();
```

Notice how deaths are capped by the current population (`mPopulation`) before subtracting from it.

Anyway, I would much prefer an approach like that then to pollute `operator-=` with `clamp`.

PR #998, opened 2021-06-25
Inverse some of the if blocks to reduce nesting

ldicker83:
Reduces a bit of nesting in the population class.

### DIFF
--- a/OPHD/Population/Population.cpp
+++ b/OPHD/Population/Population.cpp
@@ -90,6 +90,7 @@ int Population::adults() const
 	return mPopulation[PersonRole::ROLE_STUDENT] + mPopulation[PersonRole::ROLE_WORKER] + mPopulation[PersonRole::ROLE_SCIENTIST] + mPopulation[PersonRole::ROLE_RETIRED];
 }
 
+
 /**
  * Population check for new children.
  * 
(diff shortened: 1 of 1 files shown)


OPHD/Population/Population.cpp
```diff
@@ -98,156 +99,146 @@ int Population::adults() const
...
+	if (mPopulation[PersonRole::ROLE_SCIENTIST] + mPopulation[PersonRole::ROLE_WORKER] < 1) { return; }
 
-	// This should be adjusted to maybe two or three kids per couple to allow for higher growth rates.
-	if (mPopulation[PersonRole::ROLE_SCIENTIST] + mPopulation[PersonRole::ROLE_WORKER] > mPopulation[PersonRole::ROLE_CHILD])
-	{
-		mPopulationGrowth[PersonRole::ROLE_CHILD] += mPopulation[PersonRole::ROLE_SCIENTIST] / 4 + mPopulation[PersonRole::ROLE_WORKER] / 2;
+	mPopulationGrowth[PersonRole::ROLE_CHILD] += mPopulation[PersonRole::ROLE_SCIENTIST] / 4 + mPopulation[PersonRole::ROLE_WORKER] / 2;
 
-		int divisor = moraleModifierTable[moraleIndex(morale)].fertilityRate;
+	int divisor = moraleModifierTable[moraleIndex(morale)].fertilityRate;
```
DanRStevens, line 106: Curious, if values from that table are always used as divisors, why are any of the entries 0? Would it make more sense to set a minimum value of 1? If all values were non-zero, you might not need to guard expressions with `if` statements.

OPHD/Population/Population.cpp
```diff
@@ -98,156 +99,146 @@ int Population::adults() const
 void Population::spawn_children(int morale, int residences, int nurseries)
 {
 	if (residences < 1 && nurseries < 1) { return; }
+	if (mPopulation[PersonRole::ROLE_SCIENTIST] + mPopulation[PersonRole::ROLE_WORKER] < 1) { return; }
```
DanRStevens, line 102: I would probably use `<= 0`, rather than `< 1`, as it kind of gets more to the heart of the matter. This applies multiple times to the guard clauses of all population functions.

OPHD/Population/Population.cpp
```diff
@@ -98,156 +99,146 @@ int Population::adults() const
...
 
-	// This should be adjusted to maybe two or three kids per couple to allow for higher growth rates.
-	if (mPopulation[PersonRole::ROLE_SCIENTIST] + mPopulation[PersonRole::ROLE_WORKER] > mPopulation[PersonRole::ROLE_CHILD])
```
DanRStevens, line 103: It seems this check was modified, rather than simply inverted to reduce nesting. Not sure if that was intentional or not. The commit comment would suggest the change wasn't intentional, though the comment being removed here maybe suggests perhaps it was?

OPHD/Population/Population.cpp
```diff
@@ -98,156 +99,146 @@ int Population::adults() const
...
-			mPopulationGrowth[PersonRole::ROLE_STUDENT] = 0;
-		}
+	mPopulation[PersonRole::ROLE_CHILD] -= deaths;
+	mDeathCount += deaths;
+
+	if (mPopulation[PersonRole::ROLE_CHILD] <= 0)
+	{
+		mPopulationDeath[PersonRole::ROLE_CHILD] = 0;
+		mPopulationGrowth[PersonRole::ROLE_STUDENT] = 0;
 	}
```
DanRStevens, line 200: This block will only run for deaths calculated here. If deaths are the result of some external cause, this block won't run due to the earlier check at the start of the function. That may or may not be a problem. In particular, if the crime module allows for murder, you might get odd effects. In particular, a new child could be near instantly transformed to a student.

This concern applies to multiple blocks from various population functions.

DanRStevens (commented):
Looks like you had a functional change regarding a check for new children. It's not clear if that was intentional or not.

Brett208:
Affects issue #996

OPHD/Population/Population.cpp
Brett208: Flagged this PR with 996 to track as possible issue.

OPHD/Population/Population.cpp
ldicker83: It was intentional. I noticed that it was actually looking at the wrong values so I took the opportunity to fix it. Should have made it a separate commit.

DanRStevens: Not too sure what you mean by looking at the wrong values.

I believe the original code roughly matched how Outpost 2 worked. You couldn't raise new kids if they already exceeded the number of adults. I guess only so much time to work, look after kids, and make kids? Of course, those kids will soon turn into adults, and that happens faster than adults die, so population growth is of course still quite possible.

PR #999, opened 2021-06-26
Fix misreported warehouse usage

### LINKED ISSUE #960: Warehouse Storage misrepresented in full-screen UI
Storage space in the Warehouse fullscreen report is misrepresented in the visual representation of the bars.

ldicker83:
Closes #960

### DIFF
--- a/OPHD/UI/ProductListBox.cpp
+++ b/OPHD/UI/ProductListBox.cpp
@@ -26,9 +26,9 @@ void ProductListBox::productPool(ProductPool& pool)
 {
 	clear();
 
-	const auto productCount = static_cast<std::size_t>(ProductType::PRODUCT_COUNT);
+	const auto totalProducts = static_cast<std::size_t>(ProductType::PRODUCT_COUNT);
 
-	for (std::size_t product_type = 0; product_type < productCount; ++product_type)
+	for (std::size_t product_type = 0; product_type < totalProducts; ++product_type)
 	{
 		const auto productType = static_cast<ProductType>(product_type);
 		const auto productCount = pool.count(productType);
--- a/OPHD/UI/ProductListBox.cpp
+++ b/OPHD/UI/ProductListBox.cpp
@@ -26,15 +26,19 @@ void ProductListBox::productPool(ProductPool& pool)
 {
 	clear();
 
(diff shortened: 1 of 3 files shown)


OPHD/UI/ProductListBox.cpp
```diff
@@ -26,15 +26,19 @@ void ProductListBox::productPool(ProductPool& pool)
...
 	clear();
 
-	for (std::size_t product_type = 0; product_type < static_cast<std::size_t>(ProductType::PRODUCT_COUNT); ++product_type)
+	const auto totalProducts = static_cast<std::size_t>(ProductType::PRODUCT_COUNT);
+
+	for (std::size_t product_type = 0; product_type < totalProducts; ++product_type)
 	{
-		const auto productCount = pool.count(static_cast<ProductType>(product_type));
+		const auto productType = static_cast<ProductType>(product_type);
+		const auto productCount = pool.count(productType);
```
DanRStevens, line 34: Variable names could probably be improved here.

I find `totalProducts` to be a bit misleading of a name. It's basically the number of product types. Maybe `productTypeCount`?

There is both a `product_type` and a `productType`. I would be tempted to rename `product_type` to `productTypeIndex`.

OPHD/UI/ProductListBox.cpp
```diff
@@ -26,15 +26,19 @@ void ProductListBox::productPool(ProductPool& pool)
...
+
 		if (productCount > 0)
 		{
 			ProductListBoxItem* item = new ProductListBoxItem();
-			item->text = productDescription(static_cast<ProductType>(product_type));
+			item->text = productDescription(productType);
 			item->count = productCount;
-			item->usage = static_cast<float>(productCount) / static_cast<float>(pool.capacity());
+			item->usage = static_cast<float>(productCount * pool.productStorageRequirement(productType)) / static_cast<float>(pool.capacity());
```
DanRStevens, line 41: If you want to avoid `float`, you can pre-multiply before the divide:
```cpp
const auto percent = value * 100 / capacity;
```
That would allow `int` values to be used just as effectively.

OPHD/UI/ProductListBox.cpp
ldicker83: I agree. I didn't want to spend too much time on it but I noticed that naming was a little weird here.

PR #1008, opened 2021-06-26
Refactor population

Brett208:
This branch creates a `PopulationTable`, which is very similar to the `ResourcePool` class to separate some of the logic out of population.h. As a side affect, it allowed turning the `enum PopulationRole` into an `enum class` that I thought worked out really well. 

Looks like I have some wicked merge conflicts with #998. Will need some help untangling.

Also, before merging, I did not mean to commit the change to nas2D submodule in commit: _Rename PopulationTable::Role and members to improve readability and style_. I am having trouble figuring out a simple way to remove it from the commit.

### DIFF
--- a/OPHD/Population/Population.cpp
+++ b/OPHD/Population/Population.cpp
@@ -257,15 +257,15 @@ int Population::consume_food(int food)
 
 	for (int i = 0; i < population_to_kill; /**/ )
 	{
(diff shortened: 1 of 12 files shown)


DanRStevens:
I recently updated NAS2D to latest, so a rebase might solve that part for you.

PR #1015, opened 2021-06-26
Refactor Mine::pull

DanRStevens:
Refactor `Mine::pull`, using range-for, and shortening the code by using `std::min`.

Noticed this while working on #1014.


Brett208:
I think the changes make it easier to read. Also like the update to the variable naming conventions.

PR #1027, opened 2021-06-29
Fix copy/paste error

ldicker83:
Fixes what appears to be a copy/paste error in load code that references the wrong XML element.


DanRStevens:
Gah, I was afraid I may have done something like that during the refactoring. Thank you for fixing it.

PR #1033, opened 2021-07-06
Randomize amount stolen

Brett208:
Related to #911 

Add some variety for amount stolen and weight stealing based on selected difficulty.

Learned when you make a std::map const, you must use `std::map::at` instead of `std::map[]` to pull values.


DanRStevens (approved):

DanRStevens:
Ahh yes, `std::map[]` can have some somewhat surprising side effects when the key isn't found.

PR #1034, opened 2021-07-06
Add foley text when resources or food are stolen

Brett208:
I would like to look at preventing the foley text about rebel and separatist factions only appear if morale is low enough in the near future.

Stealing1

Stealing2

Related to #911


DanRStevens (approved):
Hmm, interesting idea. I like it.

----

And because I had to look this up:

> In filmmaking, Foley is the reproduction of everyday sound effects that are added to films, videos, and other media in post-production to enhance audio quality.

Wondering if there might be other definitions.

ldicker83:
I tend to call that flavor text but I'm game for whatever we want to call it.

I like the idea of variations of it especially if we can add thresholds, e.g., if morale hits X level choose from this set of flavor text, if it's at Y level, use this set, etc. etc.

PR #1035, opened 2021-07-08
Morale function grammar

Brett208:


DanRStevens (approved):
Well that makes more sense. I never even noticed that.

I suppose we should probably wrap the `static` functions in an unnamed namespace too, and remove the `static`.

PR #1041, opened 2021-07-17
Add .clang-format

DanRStevens:
Add `.clang-format`, copied from NAS2D.

Reference: https://github.com/lairworks/nas2d-core/issues/893

Having this helps migrate the code base over to a certain style. Though I'm uncertain if this should be merged before the code matches that style. I'm not certain how Visual Studio may interpret and use the file, which may be a pain if it automatically applies sweeping changes to the code.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,82 @@
+Language: Cpp
+AccessModifierOffset: -4
+AlignTrailingComments: false
+AllowShortBlocksOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: true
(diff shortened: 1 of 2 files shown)


Brett208:
Just played with this a bit in one source code file. It seems to be enforcing code changes in areas that I alter only and not holistically to the code base. I didn't test exhaustively, but I think you are safe to merge without forcing large change sets for VS users.

Clang Format

DanRStevens:
Ok, that's good news. Sounds like it can be ignored if programmers disagree with the style file.

Thanks for reporting back on the experience Brett.

PR #1043, opened 2021-07-17
Use ternary to simplify retirement logic

DanRStevens:
Flatten branch structure by pulling out variable using ternary, and using a single code block to handle the variable case.

### DIFF
--- a/OPHD/Population/Population.cpp
+++ b/OPHD/Population/Population.cpp
@@ -146,8 +146,8 @@ void Population::spawnRetiree()
 	mPopulation[PopulationTable::Role::Retired] += retiree;
 
 	/** Workers retire earlier than scientists. */
-	if (randomNumber.generate(0, 100) <= 45) { if (mPopulation[PopulationTable::Role::Scientist] > 0) { mPopulation[PopulationTable::Role::Scientist] -= retiree; } }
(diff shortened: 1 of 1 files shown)


OPHD/Population/Population.cpp
```diff
@@ -146,8 +146,8 @@ void Population::spawnRetiree()
 	mPopulation[PopulationTable::Role::Retired] += retiree;
 
 	/** Workers retire earlier than scientists. */
-	if (randomNumber.generate(0, 100) <= 45) { if (mPopulation[PopulationTable::Role::Scientist] > 0) { mPopulation[PopulationTable::Role::Scientist] -= retiree; } }
-	else { if (mPopulation[PopulationTable::Role::Worker] > 0) { mPopulation[PopulationTable::Role::Worker] -= retiree; } }
+	const auto retirePopulationType = randomNumber.generate(0, 100) <= 45 ? PopulationTable::Role::Scientist : PopulationTable::Role::Worker;
+	if (mPopulation[retirePopulationType] > 0) { mPopulation[retirePopulationType] -= retiree; }
```
DanRStevens, line 150: There's actually a pre-existing bug here. I did not address it with this PR.

There's a check for `> 0`, but `retiree` might be an arbitrary positive number. This case is unique to retirees, since there are two groups feeding retirees, and the contribution might not come entirely from either. Consider having 1 worker and 1 scientist, who both retire on the same turn. That gives 2 retirees, but neither of the feeder groups has 2 people to retire.

PR #1062, opened 2021-10-04
Refactor planet XML load

DanRStevens:
Reference: #830

Refactor Planet XML loading code so `for` loops only iterate over `XmlElement` objects, and to make use of `Dictionary`.

### DIFF
--- a/OPHD/States/Planet.cpp
+++ b/OPHD/States/Planet.cpp
@@ -131,10 +131,8 @@ namespace
 	{
 		Planet::Attributes attributes;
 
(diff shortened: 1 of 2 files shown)


Brett208:
Looks much cleaner than how I originally wrote it. I like how the concept of pulling string data from either a element or an attribute is now possible which made reading the description string trivial in code.

DanRStevens:
Technically it uses custom code for Planet data loading, so it only supports sub-elements here, and not attributes. Took me a moment to remember this code was different. I originally tried to write the code assuming it was using attributes, which didn't exactly work. Thankfully it was a fairly small change to fix it from there.

PR #1070, opened 2021-10-10
Refactor build checks

DanRStevens:
Related: PR #1069

Refactor `checkTubeConnection` and `checkStructurePlacement`. I'm sure it's possible to simplify further, though I didn't want to put too much effort into figuring out how to minimize those checks. I suspect there may be another simpler way of doing them.

### DIFF
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -96,9 +96,10 @@ bool checkStructurePlacement(Tile& tile, Direction dir)
 	}
 
 	const auto connectorDirection = structure->connectorDirection();
-	return (dir == Direction::East || dir == Direction::West) ?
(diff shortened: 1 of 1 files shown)


DanRStevens:
Hmm, I suspect commit 36f2070 has a bit. Which means commit ab950f0 is probably suspect too.

Brett208:
@DanRStevens,

What does 'has a bit' mean I'm this context?

Brett

DanRStevens:
Oops, typo. Should have been "has a bug".

DanRStevens:
Ahh, I see now. It's an associativity problem. The `bool` expression moved to the left is now associating with the `bool` expression used for the ternary. Effectively the ternary is being applied to the new full expression, rather than just the part on the right, as intended. Extra parenthesis will fix this.

PR #1087, opened 2021-10-19
Save game dialog show path

### LINKED ISSUE #929: Display save folder in saved game dialog
There's a common question of where saved games are stored on disk. It's a complicated looking path, and hard to remember. Maybe we should display the path as part of the saved game dialog. Then people would know where to find them. In particular, it makes it easier for people to upload saved games demonstrating a problem.

DanRStevens:
Closes #929

Would be helpful if people could test the "Open Folder" button on Windows or Mac. It works on Linux. It's based on the Main Menu code to open the Help topic in the browser, so hopefully it also works for the system file browser.


DanRStevens:
Thank you for the test!

I've updated the window size, adding 200 pixels to the width. That should be enough for at least another dozen characters. It's also not so big that it exceeds the minimum size of the overall game window.

I'm guessing we're probably not going to have anyone to test on MacOS.

Issue #650, opened 2020-08-20
Split MapViewState into multiple objects


DanRStevens:
The `MapViewState` class is rather large, and handles many responsibilities. I believe it should be split into component controls:
- Info bar (along the top, showing ore and such)
- Map view (detail pane)
- Mini map
- Structure build list
- Navigation controls
- Robot summary display

----

For the navigation controls, we may or may not have that include the surface/underground level display. I'm thinking no, since the current position and surface/underground level is more a property of the map view, rather than the navigation controls. Additionally, the current code ties the movement bounds checking more closely with the map than the navigation buttons.

For the minimap view, a couple of buttons on the left affect minimap display, however the next turn button doesn't really fit. Perhaps leave the map view control buttons separate for now.

More long term, should we consider moving the next turn button somewhere else? It might fit well on the top info bar. Not sure if that's where we'd want it, or if there are historical reasons for putting it where it is now. Actually, it probably makes sense to put it near the turn display.

----

Related to the above, the current `TileMap` code implements some of the map view code. We may need to clean up how those responsibilities are split. I figure `TileMap` should be a pure data structure, with no view responsibilities. We can have a separate `Control` class which takes a reference to a `TileMap` and provides the view of it. That may make layout changes easier in the future.

DanRStevens:
It looks most of the suggestions above were extracted, except for the `Structure` build list. Keeping that in mind, there are a few other areas that could be extracted:
- Structure build list
- Toggle button section
- The whole bottom section that includes the map object palette, the toggle buttons, and the minimap
- The system menu area

As a side note, as components are extracted from `MapViewState`, it may be possible to reduce the repetition in the constructor init-lists of the two constructors on `MapViewState`.

DanRStevens:
In terms of functionality to extract, there is:
- Placing `MapObject` instances
  - Structure
  - Robot
  - Lander
- Generating overlays
- Updating colony state information
- Translating hotkey key codes to game functions
- Translating clicks to either inspect or placement actions
- Turn processing logic
- Mining and refining
- Cheat effects

PR #1126, opened 2021-10-31
Split NavControl from MapViewState

DanRStevens:
Reference: #650

Extract `NavControl` from `MapViewState`, and use the new control for drawing.

Currently the event handling code to handle moving the view is still in `MapViewState`. That means a bit of duplication of code/state tracking where those buttons are on screen. That will be addressed later. Currently the focus is on the draw code.


Brett208:
Thanks for starting to separate out the MapViewState into smaller modules. Its large size is a determent to the overall codebase readability. Its large size also lended to continuing to bloat it instead of taking the more difficult path of separating it out when creating new features.

ldicker83:
> Thanks for starting to separate out the MapViewState into smaller modules. Its large size is a determent to the overall codebase readability. Its large size also lended to continuing to bloat it instead of taking the more difficult path of separating it out when creating new features.

Uh... heh... yeah. That's a whoops on my part. I got so stuck in the "Just make it work and move forward, I'll make it pretty later" that it sort of grew into a monolithic monster that was getting harder and harder to manage. This will make future development a lot easier I think.

So thank you! :D

PR #1129, opened 2021-11-05
Move MiniMap event handling code

DanRStevens:
Reference: #650, PR #1118

Move event handling code for mini map clicks from `MapViewState` to `MiniMap`.


DanRStevens:
Looks like there's a `vcpkg` error:
> Computing installation plan...
> Error: yasm-tool[core] is only supported on 'native & !uwp'

I'm guessing there's a bug in a recent `vcpkg` update.

PR #1132, opened 2021-11-07
Decouple Sprite::update and Sprite::draw

DanRStevens:
Reference: https://github.com/lairworks/nas2d-core/pull/995

Updated NAS2D submodule for decoupling `Sprite::update` and `Sprite::draw`.


Brett208:
For what it is worth, this change makes the codebase follow a similar architecture to Monogame where drawable objects have separate draw and update routines. Not sure if other frameworks follow a similar paradigm.

PR #1134, opened 2021-11-07
TileMap 1D array

DanRStevens:
Using a 1D array means fewer allocations, and less fragmented memory.

Data is organized so adjacent X coordinates are next to each other in memory. That roughly matches the previous order. It also means access is most efficient when iterating though x in inner loops. Adjacent Z data is furthest from each other, which makes sense since you only view a single plane at a time.


Brett208:
I've found learning to work with a 1D array and having to translate into 2D coordinates is easier than dealing with multi-dimensional arrays. Although admittedly I don't have experience with the z dimension.

ldicker83:
This was ultimately the goal. At the time I used a multi-dimensional array to make sure that the map structure was correct.

Thanks for taking care of this!

PR #1135, opened 2021-11-07
Update Mine spawn code

DanRStevens:
I was looking at the `Mine` spawn code and thought it was maybe a bit hard to read. I eventually decided to try re-writing it. I used a different approach here, so the results might not be quite the same.

The yield is now randomly chosen per mine, so the counts of each yield might not be exactly known, though for a large number of mines, as it typical, it should end up being pretty close to the expected value with high probability. I think this may also allow for a little bit of variation between plays, which may be desirable.

It is now more likely to generate the specified number of mines. Previously it could be less if any of the randomly chosen spawn locations were too close to each other. Now it will try to generate a new point. There is a limit on how often it retries, to prevent possible infinite loops for high mine density with impossible placement constraints.

When testing with a large number of mines, the new code seems to respect constraints better than the old code. I'm not too sure why that is. When increasing the number of mines by a factor of `1000`, the old code would place mines next to each other. The new code will still maintain strict spacing requirements.

I speed tested the code up to a factor of `100000` more mines than normal, and it runs with only a small pause of a second or two. Of course most of those generated points won't result in new mines due to placement constraints.

I opted to use functions in an unnamed namespace in the .cpp file, rather than private member functions. I was a little on the fence about one of them, which seemed appropriate to make a member, but ended up leaving in the anonymous namespace next to the other mine generation code. As a slight bonus, it means any updates to that function signature won't trigger recompiles of other modules like a private method declaration in a header file would.

The "percent" values in `HostilityMineYieldTable` are not strictly required to be percents now. They are proportional values. If they add up to 100, then they will represent percents.


DanRStevens:
In case anyone is curious what the higher mine density tests looked like:
Screenshot from 2021-11-07 03-20-49

No two mines next to each other, and nothing with 5 tiles of the edges.

DanRStevens:
Here's what the **old code** generated when increasing the number of mines by a factor of `1000`:
Screenshot from 2021-11-07 04-46-30

Noticed that many mines were placed directly next to each other.

PR #1151, opened 2021-11-14
Improved planet animation

ldicker83:


OPHD/Constants/Numbers.h
```diff
@@ -8,7 +8,7 @@
  */
 namespace constants
 {
-	inline constexpr int PlanetAnimationSpeed{50};
+	inline constexpr int PlanetAnimationSpeed{35};
```
DanRStevens, line 11: I guess technically we should call this a "delay" rather than a "speed". We decrease the delay to increase the speed.

OPHD/States/Planet.cpp
```diff
@@ -69,7 +69,7 @@ void Planet::update()
 	}
 
 	auto& renderer = NAS2D::Utility<NAS2D::Renderer>::get();
-	const auto spriteFrameOffset = NAS2D::Point{mTick % 8 * PlanetSize.x, ((mTick % 64) / 8) * PlanetSize.y};
+	const auto spriteFrameOffset = NAS2D::Point{mTick % 16 * PlanetSize.x, ((mTick % 256) / 16) * PlanetSize.y};
```
DanRStevens, line 72: Should probably be updating the submodule in the same commit as this, since things break when they are updated separately. Not that it really matters that much for back to back commits.

DanRStevens:
Oh, very cool! I rather like the new shadow effect. There's also something rather pleasing about the slower rotation rate, like it's somehow easier to appreciate the detail of the planet surface.

I noticed there were more planets added. Curious when those might be used. Also curious where the graphics came from.

ldicker83:
The surface maps are either public domain from NASA or from an artist that I purchased them from.

PR #1154, opened 2021-11-15
Add check to ensure that an invalid index can't be used.

ldicker83:
Noticed that the notification area can throw an out of bounds exception because the index provided is never checked for validity.

### DIFF
--- a/OPHD/UI/NotificationArea.cpp
+++ b/OPHD/UI/NotificationArea.cpp
@@ -107,7 +107,7 @@ void NotificationArea::onMouseDown(EventHandler::MouseButton button, int x, int 
 	}
 
 	const auto index = notificationIndex({x, y});
-	if (index != NoSelection)
+	if (index != NoSelection && index < mNotificationList.size())
 	{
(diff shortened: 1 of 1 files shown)


DanRStevens:
I was debating how to go about potentially implementing bounds checks in:
```cpp
std::size_t NotificationArea::notificationIndex(NAS2D::Point<int> pixelPosition)
{
	const auto estimatedIndex = static_cast<std::size_t>((mRect.endPoint().y - pixelPosition.y) / IconPaddedSize.y);
	// Icon is clickable, but padding area around icon is not clickable
	if (notificationRect(estimatedIndex).contains(pixelPosition))
	{
		return estimatedIndex;
	}
	return NoSelection;
}
```

I tried running the code, and didn't get any errors. It wasn't clear to me if bounds checks would be needed. Though I was focused on the mouse tracking for the popup text, not the click handler. I can see it being more relevant for the click handler.

I'm thinking I might move that check to `notificationIndex`, which it might make more logical sense. Either return a valid index, or return `NoSelection`.

PR #1155, opened 2021-11-15
Validate map view resize

ldicker83:
Forces validation of MapView tile origin during resize.

### DIFF
--- a/OPHD/UI/DetailMap.cpp
+++ b/OPHD/UI/DetailMap.cpp
@@ -66,6 +66,7 @@ void DetailMap::resize(NAS2D::Vector<int> size)
 	// Set up map draw position
 	const auto sizeInTiles = size.skewInverseBy(TileSize);
 	mMapView.viewSize(std::min(sizeInTiles.x, sizeInTiles.y));
+	mMapView.validateViewLocation();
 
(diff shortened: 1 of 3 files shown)


OPHD/UI/DetailMap.cpp
```diff
@@ -66,6 +66,7 @@ void DetailMap::resize(NAS2D::Vector<int> size)
 	// Set up map draw position
 	const auto sizeInTiles = size.skewInverseBy(TileSize);
 	mMapView.viewSize(std::min(sizeInTiles.x, sizeInTiles.y));
+	mMapView.validateViewLocation();
```
DanRStevens, line 69: I think the check should be incorportated into the `viewSize` method. When you change the view size, you may need to adjust the position to keep the view from going off the bottom or right edge of the map.

The original code:
```cpp
void MapView::viewSize(int sizeInTiles)
{
	mEdgeLength = std::max(3, sizeInTiles);
}
```

I think after the `mEdgeLength` value is updated, a call should be added:
```
mapViewLocation(mOriginTilePosition);
```
That will cause the map view position to be re-clipped, taking into account the new `mEdgeLength`.

PR #1164, opened 2021-11-17
Fix routing issue

### LINKED ISSUE #1161: Second Mine to Smelter trucking route missing
I established a second mine and smelter in my colony. No raw materials seem to leave the mine. When I click on trucking routes, it does not show a valid route. However, when I click on the mine, it shows 1 truck assigned. I'm assuming mines always come with a free truck since I didn't build a truck at the factory.

Screenshot showing second mine without a mining route:
image

Notice in the image, the new mine facility reports 1 truck assigned. Truck route overlays are highlighted. Notice the truck route shows for the original mine. 

The original mine is out of resources and is currently digging a new level so I was going to rely on the new mine production.

This might be my lack of understanding of how to play... I thought the mining route would somehow auto-establish.

Code Pull: 16NOV21
OS: WIN64
Build: Debug 64 VS2019

ldicker83:
Closes #1161

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -206,7 +206,7 @@ void MapViewState::initialize()
 	MAIN_FONT = &fontCache.load(constants::FONT_PRIMARY, constants::FontPrimaryNormal);
 
 	delete mPathSolver;
-	mPathSolver = new micropather::MicroPather(mTileMap);
+	mPathSolver = new micropather::MicroPather(mTileMap, 250, 6, false);
 }
 
 
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -195,7 +195,7 @@ void MapViewState::load(const std::string& filePath)
 	mNavControl = std::make_unique<NavControl>(*mMapView, *mTileMap);
 
 	delete mPathSolver;
-	mPathSolver = new micropather::MicroPather(mTileMap);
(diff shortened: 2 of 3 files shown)


OPHD/States/MapViewState.cpp
```diff
@@ -206,7 +206,7 @@ void MapViewState::initialize()
 	MAIN_FONT = &fontCache.load(constants::FONT_PRIMARY, constants::FontPrimaryNormal);
 
 	delete mPathSolver;
-	mPathSolver = new micropather::MicroPather(mTileMap);
+	mPathSolver = new micropather::MicroPather(mTileMap, 250, 6, false);
```
DanRStevens, line 209: Hmm, that's not very self documenting at the point of call. I needed to look up what those values meant:
```cpp
MicroPather(Graph* graph, std::size_t allocate = 250, std::size_t typicalAdjacent = 6, bool cache = true);
```

Considering MicroPather is an external library (which I don't much want to maintain or have our local copy diverge from the original), I'm not sure there's really much to do about it. Though I do wonder if we could somehow wrap or cordon off uses of the library so we don't have multiple calls like this with non-obvious parameter values.

ldicker83: There isn't. I just took the default values right out of the pathing libary.

I'm really not worried about wrapping things to hide details. That just makes things more complex than necessary just to make the code less obtuse. A comment here stating that these are the library provided defaults may be more useful though the pather is only instantiated during loading and when starting a new game... I don't see where we'd need additional uses of it.

PR #1166, opened 2021-11-17
Default to Debug builds on Linux and MacOS

### LINKED ISSUE #428: Add `debug` build target to `makefile`
Add `debug` build target to `makefile`.

Currently debug builds can be done by setting `makefile` variables from the command line. Unfortunately this isn't very obvious, and can be a bit technical and error prone. By creating an explicit `debug` build target it would become much more user friendly to do debug builds from Linux.

It should also be possible to keep separate intermediate folders for debug builds, which means the whole project won't need to be rebuilt when switching between builds.

Current example debug build command, using `makefile` variables specified at command line:
```
make CXXFLAGS_EXTRA="-g -O0" BUILDDIR=".build/debug/"
```

Related: #427
Inspired by: https://forum.outpost2.net/index.php/topic,6321.msg88997.html#msg88997

DanRStevens:
Closes #428
Reference: https://github.com/lairworks/nas2d-core/issues/867

Add **debug** and **release** config flags. Set **debug** flags as default.

To build a release, run:
```
make CONFIG=release
```

----

Explicitly ignore the return value to `std::system`, as otherwise we get warnings when compiling in debug mode.

Reference: #1165 (Collect calls to `std::system` to one place)

Documentation:
[`std::system`](https://en.cppreference.com/w/cpp/utility/program/system)
> Return value
> Implementation-defined value. If command is a null pointer, returns a nonzero value if and only if the command processor exists.

[`std::ignore`](https://en.cppreference.com/w/cpp/utility/tuple/ignore)
>An object of unspecified type such that any value can be assigned to it with no effect.
> ...
> It can also be used to avoid warnings from unused return values of [[nodiscard]] functions.

[CppCore Guidelines: ES.48: Avoid casts: Note](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#notes-4)
> Never cast to (void) to ignore a [[nodiscard]]return value. If you deliberately want to discard such a result, first think hard about whether that is really a good idea (there is usually a good reason the author of the function or of the return type used [[nodiscard]] in the first place). If you still think it's appropriate and your code reviewer agrees, use std::ignore = to turn off the warning which is simple, portable, and easy to grep.

Since the return value is implementation defined, there isn't much we can assume about it, so it may be best to just ignore it. The consequences of ignoring errors for these calls are pretty low considering the context.

### DIFF
--- a/OPHD/ShellOpenPath.cpp
+++ b/OPHD/ShellOpenPath.cpp
@@ -2,6 +2,7 @@
 
 #include <string_view>
 #include <cstdlib>
+#include <tuple>
 
 
 namespace
@@ -20,5 +21,6 @@ namespace
 
 void shellOpenPath(const std::string& path)
 {
-	std::system((std::string{ShellOpenCommand} + " " + path).c_str());
+	// Explicitly ignore implementation defined return value
+	std::ignore = std::system((std::string{ShellOpenCommand} + " " + path).c_str());
 }
--- a/makefile
+++ b/makefile
@@ -1,5 +1,10 @@
 # Source http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/
 
+CONFIG := debug
+debug_CXX_FLAGS := -Og -g
+release_CXX_FLAGS := -O3
+CONFIG_CXX_FLAGS := $($(CONFIG)_CXX_FLAGS)
+
 SRCDIR := OPHD/
 BUILDDIR := .build/
 OBJDIR := $(BUILDDIR)obj/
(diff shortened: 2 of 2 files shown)


DanRStevens:
If we need to address the use of `std::ignore` to silence warnings about the `[[nodiscard]]` value, then perhaps we should open a new case.

I've considered returning the result of `std::system` from `shellOpenPath`, but then we'd be leaking implementation defined values into the calling code, which might not be prepared to deal with that.

I've also considered converting error return codes to exceptions. However, the documentation doesn't clearly indicate what is considered an error return code, so there's no guarantee it would be correct. At best, we could rely on implied likely meaning, which might be consistent across the platforms we currently support.
> Returns an implementation-defined value (usually the value that the invoked program returns).

Of course, the current code just ignores errors, which given the current context is a fairly reasonably thing to do. Currently we only use the method to either open the browser to the "Help" page, or opening the system file browser to the saved game folder. Both are pretty innocuous if the buttons just do nothing in the event of an error.

PR #1171, opened 2021-11-18
Change default integrity to 100 when loading saved games

DanRStevens:
For old saved games with no integrity value, we should probably assume structures still have full integrity, rather than no integrity. After all, decay wasn't implemented before. Though more practically, it prevents an entire colony from decaying into a non functioning mess when calculating the next turn.

This has kind of been bothering me for a while, when I load old saved games to try and quickly test something.


ldicker83:
Oh for...

I hadn't even thought about this one. Nice catch!

PR #1172, opened 2021-11-19
Refactor `Mine::pull` to use `StorableResources`

DanRStevens:
Refactor `Mine::pull` to work with `StorableResources` objects. It's now possible to transfer all resource amounts in a single call, rather than having to iterate over each component.

### DIFF
--- a/OPHD/Things/Structures/MineFacility.cpp
+++ b/OPHD/Things/Structures/MineFacility.cpp
@@ -55,6 +55,13 @@ void MineFacility::activated()
 }
 
 
+StorableResources MineFacility::maxTransferAmounts()
+{
+	const auto remainingCapacity = MaxCapacity - production();
+	return remainingCapacity.cap(constants::BaseMineProductionRate);
+}
+
+
 void MineFacility::think()
 {
 	if (forceIdle()) { return; }
--- a/OPHD/Things/Structures/MineFacility.cpp
+++ b/OPHD/Things/Structures/MineFacility.cpp
@@ -4,7 +4,10 @@
 #include "../../Constants/Strings.h"
 
 
-const int MineFacilityStorageCapacity = 500;
+namespace
+{
+	const int MineFacilityStorageCapacity = 500;
+}
 
 
 /**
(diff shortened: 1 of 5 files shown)


OPHD/Things/Structures/MineFacility.cpp
```diff
@@ -4,7 +4,10 @@
 #include "../../Constants/Strings.h"
 
 
-const int MineFacilityStorageCapacity = 500;
+namespace
+{
+	const int MineFacilityStorageCapacity = 500;
```
ldicker83, line 9: Shouldn't this just be part of the MineFacility class itself instead of in an unamed namespace?

DanRStevens: Maybe. I was wondering about that. I opted to leave it as a file level constant. The exact details there weren't super relevant to the refactoring I was doing at the time. I did noticed though that there was a mix of code using either dynamic data from member fields, and other code that was based on a compile time constant. We should probably address that at some point.

I guess one question to ask is, can the storage capacity be upgraded as a runtime property? Perhaps it can be something unlocked by tech research.

DanRStevens: Actually, another point is we could allow for varying storage levels. Currently we take some fixed quantity, and then divide it by the number of resource types. We could instead go the other direct. We could define a `StorableResources` constant specifying capacities, and then use the `total()` method to determine the total storage space for all. That might allow space or transfer amounts to vary based on the resource type.

ldicker83: > I guess one question to ask is, can the storage capacity be upgraded as a runtime property? Perhaps it can be something unlocked by tech research.

This is a good question.

My thinking is that storage capacity is determined by the actual structure itself. E.g., MineFacility Mk 1, Mk 2, etc. Same with other storage facilities like the Storage Tanks Mk 1 and Mk 2. We could apply a modifier through research... e.g., Facility 1 has a constant 500 units of storage but with research into, I dunno, Computer Aided Storage Management, or something like that, a modifier like 5% or 10% could be added in addition to the base storage. This wouldn't really modify the individual structure itself, just apply as a global modifier based on research.

That's my thinking anyway.

PR #1178, opened 2021-11-20
Add technology definition

ldicker83:
Adds a class to load, store and find technologies.


ldicker83:
@DanRStevens The class I developed to actually load the technology definition is called `TechnologyReader` -- I don't really care for the name. I was thinking maybe instead `TechnologyLibrary` or `TechnologyCatalog` to be more inline with some of the other naming schemes but wanted you to weigh in before finalizing this PR.

OPHD/Technology/TechnologyReader.cpp
```diff
@@ -3,13 +3,91 @@
...
+		std::string elementValue = techElement->getText();
+		if (elementName == "name")
+		{
+			tech.name = elementValue;
+		}
+		else if (elementName == "description")
+		{
+			tech.description = elementValue;
+		}
+		else if (elementName == "requires")
```
DanRStevens: You may want to take a look at `parsePlanet`. The code there could perhaps be extracted to a generic XML processing routine:
```cpp
NAS2D::Dictionary dictionary{};
for (const auto* element = xmlNode->firstChildElement(); element; element = element->nextSiblingElement())
{
	dictionary.set(element->value(), element->getText());
}
```

OPHD/Technology/TechnologyReader.cpp
```diff
@@ -104,5 +131,32 @@ void TechnologyReader::readTechnology(NAS2D::Xml::XmlElement& technology, const
 
 void TechnologyReader::readEffects(NAS2D::Xml::XmlElement& effects, Technology& technology)
 {
+	for (auto effectElement = effects.firstChildElement(); effectElement; effectElement->nextSiblingElement())
+	{
+		const std::string effectName = effectElement->value();
+		const std::string effectValue = effectElement->getText();
+		auto effectAttributes = NAS2D::attributesToDictionary(*effectElement);
 
+		if (effectName == "modifier")
```
DanRStevens: Ditto.

OPHD/Technology/TechnologyReader.cpp
```diff
@@ -104,5 +131,32 @@ void TechnologyReader::readTechnology(NAS2D::Xml::XmlElement& technology, const
...
+		const std::string effectValue = effectElement->getText();
+		auto effectAttributes = NAS2D::attributesToDictionary(*effectElement);
 
+		if (effectName == "modifier")
+		{
+			technology.effects.push_back(std::make_unique<Technology::Effect>(
+				Technology::Modifier(
+					StringToModifier.at(effectAttributes.get<std::string>("type")),
+					std::stof(effectValue))
+				));
```
DanRStevens, line 162: I believe `std::make_unique` is intended to be used with the exact derived type, rather than the base class type. The parameters should be constructor parameters, rather than a fully constructed object.
```cpp
technology.effects.push_back(
  std::make_unique<Technology::Modifier>(
    StringToModifier.at(effectAttributes.get<std::string>("type")),
    std::stof(effectValue))
  )
);
```

OPHD/Technology/TechnologyReader.cpp
```diff
@@ -104,5 +131,32 @@ void TechnologyReader::readTechnology(NAS2D::Xml::XmlElement& technology, const
...
+		}
+		else if (effectName == "unlock")
+		{
+			technology.effects.push_back(std::make_unique<Technology::Effect>(
+				Technology::Unlock(
+					StringToUnlock.at(effectAttributes.get<std::string>("type")),
+					effectValue)
+				));
```
DanRStevens, line 170: Ditto.

DanRStevens (commented):
Looks like a pretty good start.

I would probably make a few simplifications to how `Modifer` and `Unlock` are stored. That seemed like the key point. The rest of the suggestions were considerably more minor.

OPHD/Technology/Technology.h
```diff
@@ -0,0 +1,18 @@
...
+struct Effect;
+
+struct Technology
+{
+	std::string name;
+	std::string description;
+	
+	unsigned int id{0};
+	unsigned int labType{0};
+	unsigned int cost{0};
```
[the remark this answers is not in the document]
ldicker83: Yeah, I ultimately decided to force initialization through instantiation versus using default initialization.

OPHD/Technology/Technology.h
```diff
@@ -1,18 +1,90 @@
...
-	unsigned int cost{0};
-	std::vector<unsigned int> requiredTechnologies;
-	std::vector<Effect> effects;
+	int id{0};
+	int labType{0};
+	int cost{0};
+
+	std::vector<int> requiredTechnologies;
+
+	std::vector<std::unique_ptr<Effect>> effects;
```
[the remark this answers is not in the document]
ldicker83: This makes sense to me. I got used to C#'s boxing/unboxing semantics and sort of emulated that here. And since we're not adding/removing effects during runtime except when loading these can be separated and managed more easily.

OPHD/Technology/TechnologyReader.h
```diff
@@ -0,0 +1,18 @@
...
+#include <map>
+#include <string>
+
+class TechnologyReader
+{
+public:
+	TechnologyReader() = delete;
+	TechnologyReader(const std::string& techFile);
+
+	std::map<int, Technology> technologies;
```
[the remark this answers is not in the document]
ldicker83: See that's what I was thinking -- I originally called it TechnologyTree but it didn't feel quite right since it's not stored or handled as a tree. So I though perhaps TechnologyCatalog (to match other naming conventions) or TechnologyLibrary.

OPHD/Technology/Technology.h
```diff
@@ -77,12 +77,20 @@ struct Technology
 		std::string value;
 	};
 
+
+	Technology() = delete;
+	Technology(int id, int labType, int cost) :
+		id(id),
+		labType(labType),
+		cost(cost)
+	{}
```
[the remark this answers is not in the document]
ldicker83: Sure, I was going for aggregate initialization here, I just wanted to avoid having these instantiated without initialization. Again, coming a bit from C# and forgetting they're not the same language. Anyway I'm game for changing it. Will make updates.

OPHD/Technology/TechnologyReader.cpp
```diff
@@ -3,13 +3,91 @@
...
+
 TechnologyReader::TechnologyReader(const std::string& techFile)
 {
 	Xml::XmlDocument xmlDocument = openXmlFile(techFile, "technology");
+
+	auto root = xmlDocument.firstChildElement("technology");
+
+	auto firstCategory = root->firstChildElement("category");
+	if (!firstCategory) { return; }
+	readCategories(*firstCategory);
```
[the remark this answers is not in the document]
ldicker83: I structured it in the way that we think about things naturally -- technologies are grouped by category. So they are grouped together in the definitions by category. Very easy and natural way to think about them. Would it be more efficient to reference a category index? Sure. But I'm less worried about that and more concerned about things being easy to understand.

That stated, I've already written the tool and the loading functions like this so I plan to keep it that way for now. :D

OPHD/Technology/TechnologyReader.cpp
ldicker83: I'm game, though I'll probably do that as a refactor versus this PR so that this can get merged sooner rather than later. Plus I think extracting common code is not relevant to this particular PR.

OPHD/Technology/TechnologyReader.cpp
ldicker83: I think you're right since I had to do this in a sort of round about way. The above discussion about making Modifier and Unlock separate classes will resolve this.

OPHD/Technology/TechnologyReader.h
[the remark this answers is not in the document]
DanRStevens: I think "tech tree" is pretty standard terminology. Besides, it's not really about how it's stored. It's more about the graph of dependencies between techs, which forms a tree. In particular, there are no cycles in the dependency graph, which is an important part of what makes the graph a tree.

[Wikipedia: Tree (graph theory)](https://en.wikipedia.org/wiki/Tree_(graph_theory))
> In graph theory, a tree is an undirected graph in which any two vertices are connected by exactly one path, or equivalently a connected acyclic undirected graph. A forest is an undirected graph in which any two vertices are connected by at most one path, or equivalently an acyclic undirected graph, or equivalently a disjoint union of trees.

I guess you could call it a `TechForest`, just in case not all techs are connected by a dependency path. :wink:

DanRStevens (approved):
That was a very clear and easy to read set of changes.

I noticed a few other largely minor potential improvements. I think the most valuable one was about pre-checking for required/option fields when parsing `Technology`, and converting to aggregate initialization of `Technology`. But that's just refactoring.

Anyway, I think this is suitable to merge as is.

OPHD/Technology/TechnologyCatalog.h
```diff
@@ -16,11 +16,14 @@ class TechnologyCatalog
...
 
+	const std::vector<std::string>& categories() const { return mCategorNames; }
+
 	const Technology& technologyFromId(int id) const;
 
 	const std::vector<Technology> technologiesInCategory(const std::string& categoryName) const;
 
 private:
+	std::vector<std::string> mCategorNames;
 	std::map<std::string, std::vector<Technology>> mCategories;
```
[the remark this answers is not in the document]
ldicker83: I opted to have this cached for the UI functions that are going to come with the Research UI. Might as well have those cached so that we can avoid doing any sort of lookups besides basic linear vector traversal.

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+	auto root = xmlDocument.firstChildElement("technology");
+
+	auto firstCategory = root->firstChildElement("category");
+	if (!firstCategory) { return; }
+	readCategories(*firstCategory);
+}
+
+
+const Technology& TechnologyCatalog::technologyFromId(int id) const
```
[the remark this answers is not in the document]
ldicker83: I was thinking the same thing and it was my original plan. I may still go that route for this lookup using pointers returned as references. E.g., keep the original `std::map<category name, technology>` and then a separate `std::map<tech id, technology*>` after the first map is set up. I'll see how it works out as we start to really use this class.

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -0,0 +1,185 @@
...
+							attributes.get<int>("lab_type"),
+							attributes.get<int>("cost")});
+
+	auto& tech = technologies.back();
+	
+	for (auto techElement = technology.firstChildElement(); techElement; techElement = techElement->nextSiblingElement())
+	{
+		const std::string elementName = techElement->value();
+		std::string elementValue = techElement->getText();
+		if (elementName == "name")
```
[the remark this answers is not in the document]
ldicker83: I looked over it and I definitely want to incorporate it. I'm going to leave that for a refactor though as I want to move onto the next part of the research implementation.

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -0,0 +1,185 @@
...
+	{
+		const std::string effectName = effectElement->value();
+		const std::string effectValue = effectElement->getText();
+		auto effectAttributes = NAS2D::attributesToDictionary(*effectElement);
+
+		if (effectName == "modifier")
+		{
+			technology.modifiers.push_back(
+				{StringToModifier.at(effectAttributes.get<std::string>("type")),
+				 std::stof(effectValue)});
```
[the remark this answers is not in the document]
ldicker83: Did not know that. That's a good adjustment to make before merging.

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -98,30 +98,25 @@ void TechnologyCatalog::readTechnologiesInCategory(const std::string& categoryNa
...
-		const auto it = std::find_if(technologies.begin(), technologies.end(), [id](const Technology& tech) { return tech.id == id; });
+		const auto it = std::find_if(technologies.begin(), technologies.end(), [tech](const Technology& technology) { return technology.id == tech.id; });
 		if (it != technologies.end())
 		{
-			throw std::runtime_error("TechnologyReader: Technology ID redefinition '" + std::to_string(id) +
+			throw std::runtime_error("TechnologyReader: Technology ID redefinition '" + std::to_string(tech.id) +
 				"' at (" + std::to_string(technology->row()) + ", " + std::to_string(technology->column()) + ")");
 		}
-
-		readTechnology(categoryName, *technology, attributes);
```
DanRStevens: Looks like a small oversight with this change. The `tech` value isn't being stored anywhere after reading it. Previously that was being done inside of `readTechnology`, which was modified to instead return the read value.

Aside from that, this nicely decouples the code.

Edit: I see that was corrected in a later commit.

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -36,6 +37,71 @@ namespace
 		{"structure", Technology::Unlock::Unlocks::Structure},
 		{"vehicle", Technology::Unlock::Unlocks::Vehicle}
 	};
+
+
+	void readEffects(NAS2D::Xml::XmlElement& effects, Technology& technology)
```
DanRStevens, line 42: You can probably do similar with this method too. Instead of taking a `Technology&` as a parameter, and then modifying it, you could instead return the parsed effects arrays.

Here the code is complicated by having two arrays. I suppose you could return a `struct`/`tuple` of the two arrays, though you could also split the function into two, where each function only processes are particular element type. I believe the `for` loop iteration can use a parameter to `firstChildElement` and `nextSiblingElement` so you only process elements with a given name. That should make it relatively easy to split the function into two, each one returning a `std::vector` for that element type.

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -36,6 +37,71 @@ namespace
...
+		}
+	}
+
+
+	Technology readTechnology(NAS2D::Xml::XmlElement& technology)
+	{
+		const auto attributes = NAS2D::attributesToDictionary(technology);
+		Technology tech = {attributes.get<int>("id"),
+						   attributes.get<int>("lab_type"),
+						   attributes.get<int>("cost")};
```
DanRStevens, line 71: Indentation

OPHD/Technology/TechnologyCatalog.h
```diff
@@ -28,6 +28,4 @@ class TechnologyCatalog
 
 	void readCategories(NAS2D::Xml::XmlElement& node);
 	void readTechnologiesInCategory(const std::string& categoryName, NAS2D::Xml::XmlElement& category);
-	Technology readTechnology(NAS2D::Xml::XmlElement& technology);
-	void readEffects(NAS2D::Xml::XmlElement& effects, Technology& technology);
```
DanRStevens: I like the structure change from moving the helper methods to an unnamed namespace.

I wonder if something similar could be done for the remaining two methods. It seems like they should be rather similar. Potentially you could remove includes for the XML code from the header file, which means fewer transitive includes where `TechnologyCatalog` is used.

DanRStevens (commented):
I'm rather happy with these changes.

OPHD/Technology/TechnologyCatalog.cpp
ldicker83: Yeah, I didn't actually test the code after compiling and realized when testing something else that the technologies themselves weren't getting saved. Wherps.

OPHD/Technology/TechnologyCatalog.h
ldicker83: I was thinking the same thing. Ultimately that's going to be the plan, pull the other two out and moved into the unnamed namespace and keep the class definition very simple and clean.

OPHD/Technology/TechnologyCatalog.cpp
ldicker83: I was thinking about that and how to do it. Initially I was going to have it return an effect but since they are no longer derived I need to figure out a different way to do it. It shouldn't be difficult, just need to come up with a clean way to do it that doesn't also pollute `readTechnology`.

DanRStevens: For reference, maybe look at the code for:
```cpp
for (const auto* node = element->firstChildElement("imagesheet"); node; node = node->nextSiblingElement("imagesheet"))
```

Or:
```cpp
for (auto* mineElement = element->firstChildElement("mines")->firstChildElement("mine"); mineElement; mineElement = mineElement->nextSiblingElement())
```

There are a few places in the code that use a parameter to restrict iteration to only `XmlElement`s of a certain type.

PR #1184, opened 2021-11-22
Add GitHub Actions config for Windows builds

DanRStevens:
GitHub Actions provides runners for Windows, Mac, and Linux. I was experimenting with the Windows builds, since currently that's the slowest for us. Our existing Windows builds on AppVeyor were running serially, taking about 4 minutes per configuration, with 2 configurations, so about 8 minutes per build. The equivalent pipeline using GitHub Actions runs the configurations in parallel, so both will complete in under 4 minutes.

Here are some details on the service offering:
[Usage limits, billing, and administration](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration)
> GitHub plan | Total concurrent jobs | Maximum concurrent macOS jobs
> -- | -- | --
> Free | 20 | 5
> Pro | 40 | 5
> Team | 60 | 5
> Enterprise | 180 | 50

We're of course on the Free plan. Having up to 20 concurrent jobs sounds amazing. We could potentially do builds for all supported platforms and configurations in parallel from a single CI provider.

This may even help with release packaging if we do all builds on a single provider.

As a bonus, it seems there is a preview for VS2022 available.

### DIFF
--- a/.github/workflows/msbuild.yml
+++ b/.github/workflows/msbuild.yml
@@ -1,0 +1,32 @@
+name: MSBuild
+
+on: [push]
+
+env:
+  # Path to the solution file relative to the root of the project.
+  SOLUTION_FILE_PATH: .
+
+  # Configuration type to build.
+  # You can convert this to a build matrix if you need coverage of multiple configuration types.
+  # https://docs.github.com/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+  BUILD_CONFIGURATION: Release
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


DanRStevens:
Small note: Previously I was looking into doing Windows builds on CircleCI, which would also have allowed jobs to run in parallel. Though the Windows environment didn't have `vcpkg` pre-installed, which made the task a bit more difficult. The GitHub Actions Windows environment comes with `vcpkg` pre-installed.

PR #1185, opened 2021-11-23
Fix incorrect structures in menu

### LINKED ISSUE #1156: User able to manipulate 'Take me there' button to build UG structures above ground
A user can build underground structures on the surface if they exploit a bug in the `take me there` process.

**Steps to reproduce:**
1. Get a surface event to occur that has a `Take me there` button
2. Go underground
3. Click on the event and the `Take me there` button.
4. Notice that while the screen centered on the event on the surface, the UI does not clear underground structures
5. Start building underground structures
6. Clicking the next turn button will reset the structure menu back to surface.

Screen shot of being able to build UG structures on the surface:
OPHD Screenshot

Screen shot of a residence I built on the surface after exploiting the bug
image

Version: Master branch pulled on 15NOV.
OS: WIN64
Build: Debug 64 bit Via VS 2019

Well, I made good on my promise to try the game again and found a bug as promised as well.

ldicker83:
Closes #1156

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -287,6 +287,7 @@ void MapViewState::focusOnStructure(Structure* structure)
 {
 	if (!structure) { return; }
 	mMapView->centerOn(NAS2D::Utility<StructureManager>::get().tileFromStructure(structure).xyz());
+	populateStructureMenu();
 }
 
 
--- a/OPHD/States/MapViewStateUi.cpp
+++ b/OPHD/States/MapViewStateUi.cpp
@@ -690,6 +690,7 @@ void MapViewState::onFileIoAction(const std::string& filePath, FileIo::FileOpera
 void MapViewState::onNotificationWindowTakeMeThere(const MapCoordinate& position)
 {
 	mMapView->centerOn(position);
+	populateStructureMenu();
 }
 
 


OPHD/States/MapViewState.cpp
```diff
@@ -287,6 +287,7 @@ void MapViewState::focusOnStructure(Structure* structure)
 {
 	if (!structure) { return; }
 	mMapView->centerOn(NAS2D::Utility<StructureManager>::get().tileFromStructure(structure).xyz());
+	populateStructureMenu();
```
DanRStevens, line 290: Potentially you could make use of `void MapViewState::onNotificationWindowTakeMeThere(const MapCoordinate& position)` to avoid duplication:
```cpp
onNotificationWindowTakeMeThere(NAS2D::Utility<StructureManager>::get().tileFromStructure(structure).xyz());
```

Though I would want to rename it to `onTakeMeThere` to better represent the more general usage.

DanRStevens (approved):
Looks like that should fix the problem.

OPHD/States/MapViewState.cpp
ldicker83: I was thinking about the best way to do that. I'd want to rename that function as you suggested to `onTakeMeThere` and just call it from the `focusOnStructure()` function. I'll open an issue for it.

PR #1195, opened 2021-12-12
Add research tracking

ldicker83:
Adds research tracking.

### DIFF
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -592,3 +592,25 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
 
 	return research;
 }
+
+
+void readResearch(NAS2D::Xml::XmlElement* element, ResearchTracker& tracker)
+{
+	const std::vector<std::string> researchList = NAS2D::split(element->attribute("completed"));
+
+	for (auto& item : researchList)
+	{
+		tracker.addCompletedResearch(std::stoi(item));
+	}
+
+	for (auto currentResearch = element->firstChildElement();
+		currentResearch != nullptr;
+		currentResearch = currentResearch->nextSiblingElement())
+	{
+		const auto dictionary = NAS2D::attributesToDictionary(*currentResearch);
+
+		tracker.startResearch(dictionary.get<int>("tech_id"),
+							  dictionary.get<int>("progress"),
+							  dictionary.get<int>("assigned"));
+	}
+}
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -578,14 +578,14 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
 
 	research->attribute("completed_techs", completedResearch);
 
-	for (auto& item : tracker.currentResearch())
+	for (const auto& [techId, values] : tracker.currentResearch())
 	{
 		research->linkEndChild(NAS2D::dictionaryToAttributes(
 			"current",
 			{{
-				{"tech_id", item.first},
-				{"progress", std::get<0>(item.second)},
-				{"assigned", std::get<1>(item.second)},
+				{"tech_id", techId},
+				{"progress", std::get<0>(values)},
+				{"assigned", std::get<1>(values)},
 			}}
 		));
 	}
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -557,3 +557,13 @@ NAS2D::Xml::XmlElement* writeRobots(RobotPool& robotPool, RobotTileTable& robotM
 
 	return robots;
 }
+
+
+NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker&)
+{
+	auto* research = new NAS2D::Xml::XmlElement("research");
+
+
+
+	return research;
+}
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -596,7 +596,9 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
 
 void readResearch(NAS2D::Xml::XmlElement* element, ResearchTracker& tracker)
 {
-	const std::vector<std::string> researchList = NAS2D::split(element->attribute("completed"));
+	if (!element) { return; }
+	
+	const std::vector<std::string> researchList = NAS2D::split(element->attribute("completed_techs"));
 
 	for (auto& item : researchList)
 	{
@@ -604,8 +606,8 @@ void readResearch(NAS2D::Xml::XmlElement* element, ResearchTracker& tracker)
 	}
 
 	for (auto currentResearch = element->firstChildElement();
-		currentResearch != nullptr;
-		currentResearch = currentResearch->nextSiblingElement())
+		 currentResearch != nullptr;
+		 currentResearch = currentResearch->nextSiblingElement())
 	{
 		const auto dictionary = NAS2D::attributesToDictionary(*currentResearch);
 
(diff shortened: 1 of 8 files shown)


OPHD/States/MapViewStateHelper.cpp
```diff
@@ -559,11 +561,34 @@ NAS2D::Xml::XmlElement* writeRobots(RobotPool& robotPool, RobotTileTable& robotM
...
-NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker&)
+NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
 {
 	auto* research = new NAS2D::Xml::XmlElement("research");
 
+	std::string completedResearch;
+	for (auto techId : tracker.completedResearch())
+	{
+		completedResearch += std::to_string(techId) + ",";
+	}
```
DanRStevens, line 572:
```cpp
const auto completedResearch = NAS2D::join(tracker.completedResearch(), ",");
```

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -559,11 +561,34 @@ NAS2D::Xml::XmlElement* writeRobots(RobotPool& robotPool, RobotTileTable& robotM
...
+
+	if (!completedResearch.empty())
+	{
+		completedResearch.pop_back();
+	}
```
DanRStevens, line 577: Not need for this fixup for trailing "," if you use `NAS2D::join`.

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -592,3 +592,25 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
...
+
+	for (auto currentResearch = element->firstChildElement();
+		currentResearch != nullptr;
+		currentResearch = currentResearch->nextSiblingElement())
+	{
+		const auto dictionary = NAS2D::attributesToDictionary(*currentResearch);
+
+		tracker.startResearch(dictionary.get<int>("tech_id"),
+							  dictionary.get<int>("progress"),
+							  dictionary.get<int>("assigned"));
```
DanRStevens, line 616: I would suggest spilling all parameters for a multi-line call, so that all parameters line up the same (which remains true regardless of tab/space settings).

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -596,16 +596,18 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
...
 	for (auto& item : researchList)
 	{
 		tracker.addCompletedResearch(std::stoi(item));
 	}
 
 	for (auto currentResearch = element->firstChildElement();
-		currentResearch != nullptr;
-		currentResearch = currentResearch->nextSiblingElement())
+		 currentResearch != nullptr;
+		 currentResearch = currentResearch->nextSiblingElement())
```
DanRStevens, line 610: The indentation here looks suspect. I think it may be mixing tabs and spaces.

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -596,16 +596,18 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
 
 void readResearch(NAS2D::Xml::XmlElement* element, ResearchTracker& tracker)
 {
-	const std::vector<std::string> researchList = NAS2D::split(element->attribute("completed"));
+	if (!element) { return; }
+	
+	const std::vector<std::string> researchList = NAS2D::split(element->attribute("completed_techs"));
```
DanRStevens, line 601: Could use `auto` here.

OPHD/Technology/ResearchTracker.h
```diff
@@ -37,6 +37,8 @@ class ResearchTracker
 	void updateResearch(int techId, int progress, int assigned);
 	const std::tuple<int, int>& researchProgress(int techId) const;
 
+	void clear();
```
DanRStevens, line 40: It might be better to re-initialize an object, rather than use a `clear()` method. If you add fields to a class, then you must update `clear()` or you may end up with subtle bugs. It's generally safer to re-initialize.
```cpp
researchTracker = ResearchTracker{}; // Construct new, and assign over top of old instance
```

From another perspective, the constructor already is a method to fully initialize a new blank object.

As a bonus, it may encourage using a non-default constructor to actually initialize the object, rather than just create a new empty object. For instance, when loading a saved game, if the list of completed and in progress techs is already parsed, you could initialize with:
```cpp
researchTracker = ResearchTracker{completedTechs, inProgressTechs};
```

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -578,14 +578,14 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
...
 	{
 		research->linkEndChild(NAS2D::dictionaryToAttributes(
 			"current",
 			{{
-				{"tech_id", item.first},
-				{"progress", std::get<0>(item.second)},
-				{"assigned", std::get<1>(item.second)},
+				{"tech_id", techId},
+				{"progress", std::get<0>(values)},
+				{"assigned", std::get<1>(values)},
```
DanRStevens, line 588: I wonder if you can nest structured bindings to get rid of the `std::get<0>` and `std::get<1>`. Or, just use a `struct` rather than a `std::tuple`.

OPHD/States/MapViewStateHelper.cpp
DanRStevens: Was just thinking `int` needed to be converted to `std::string` before the call to `join`:
```cpp
NAS2D::mapToVector(tracker.completedResearch(), [](const auto& x){return std::to_string(x);})
```

All together, that would be:
```cpp
const auto completedResearch = NAS2D::join(NAS2D::mapToVector(tracker.completedResearch(), [](const auto& x){return std::to_string(x);}), ",");
```

DanRStevens: Ideally, `mapToVector` should work with a function reference, though I've noticed the compiler fails to deduce the template parameter in that case. Perhaps I need to look into this. Ideally it should be possible to convert using:
```cpp
NAS2D::mapToVector(tracker.completedResearch(), std::to_string); // Doesn't currently work, due to failed template deduction
```

OPHD/States/MapViewStateHelper.cpp
ldicker83: I have a feeling this is mostly how GitHub handles tabs vs Visual Studio since for me it lines up properly. I can still make that change anyway since it's problematic outside of my environment.

OPHD/States/MapViewStateHelper.cpp
ldicker83: Interesting. Will investigate.

OPHD/States/MapViewStateHelper.cpp
ldicker83: I tried, it yelled at me. I can go back to a struct, I just didn't think I'd ever use more than the two and tend to prefer using already available structures/classes versus providing my own.

Granted I agree `get<0>` and `get<1>` is _way_ less intuitive than named fields in a custom `struct`... so perhaps it's worth it.

OPHD/Technology/ResearchTracker.cpp
```diff
@@ -0,0 +1,38 @@
...
+	if (std::find(mCompleted.begin(), mCompleted.end(), techId) != mCompleted.end())
+	{
+		throw std::runtime_error("Can't update a technology that's already researched.");
+	}
+
+	auto emplacedItem = mTechnologiesBeingResearched.emplace(techId, std::make_tuple(progress, assigned));
+	if (!emplacedItem.second)
+	{
+		(*emplacedItem.first).second = {progress, assigned};
+	}
```
DanRStevens, line 19: I just noticed this. You can use [`std::map::insert_or_assign`](https://en.cppreference.com/w/cpp/container/map/insert_or_assign) instead of [`std::map::emplace`](https://en.cppreference.com/w/cpp/container/map/emplace). That would eliminate the need for the `if` block to manually write the data when the key already exists.

It was tempting to use `operator[]`, though that would default construct an object before assignment when the key isn't found. Using `insert_or_assign` avoid the needless default construction.

PR #1196, opened 2021-12-15
Refactor `ResearchTracker`

DanRStevens:
Implements the changes suggested in #1195.

### DIFF
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -578,8 +578,8 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
 			"current",
 			{{
 				{"tech_id", techId},
-				{"progress", std::get<0>(values)},
-				{"assigned", std::get<1>(values)},
+				{"progress", values.progress},
+				{"assigned", values.scientistsAssigned},
 			}}
 		));
 	}
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -611,8 +611,10 @@ void readResearch(NAS2D::Xml::XmlElement* element, ResearchTracker& tracker)
 	{
 		const auto dictionary = NAS2D::attributesToDictionary(*currentResearch);
 
-		tracker.startResearch(dictionary.get<int>("tech_id"),
-							  dictionary.get<int>("progress"),
-							  dictionary.get<int>("assigned"));
+		tracker.startResearch(
+			dictionary.get<int>("tech_id"),
+			dictionary.get<int>("progress"),
+			dictionary.get<int>("assigned")
+		);
 	}
 }
--- a/OPHD/States/MapViewStateHelper.cpp
+++ b/OPHD/States/MapViewStateHelper.cpp
@@ -598,7 +598,7 @@ void readResearch(NAS2D::Xml::XmlElement* element, ResearchTracker& tracker)
 {
 	if (!element) { return; }
 	
-	const std::vector<std::string> researchList = NAS2D::split(element->attribute("completed_techs"));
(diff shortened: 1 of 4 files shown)


ldicker83 (commented):
Benefit of reinitialize vs clear()?

Not opposed but I tend to use clear() to maintain consistent syntax with STL containers.

OPHD/States/MapViewStateHelper.cpp
```diff
@@ -578,8 +578,8 @@ NAS2D::Xml::XmlElement* writeResearch(const ResearchTracker& tracker)
 			"current",
 			{{
 				{"tech_id", techId},
-				{"progress", std::get<0>(values)},
-				{"assigned", std::get<1>(values)},
+				{"progress", values.progress},
```
ldicker83, line 581: You know what... I see your point here. It's been what, a month since I wrote this code? And looking at it here in this commit after being off my feet for two weeks with COVID-19 makes it so, so, so easy to understand what this is supposed to mean.

OPHD/Technology/ResearchTracker.cpp
```diff
@@ -12,12 +12,7 @@ void ResearchTracker::startResearch(int techId, int progress, int assigned)
 		throw std::runtime_error("Can't update a technology that's already researched.");
 	}
 
-	auto emplacedItem = mTechnologiesBeingResearched.emplace(techId, ResearchProgress{progress, assigned});
-	if (!emplacedItem.second)
-	{
-		(*emplacedItem.first).second = {progress, assigned};
-	}
-
+	mTechnologiesBeingResearched.insert_or_assign(techId, ResearchProgress{progress, assigned});
```
ldicker83, line 15: Didn't even know this was available... because it exactly solves the original code.

Doesn't help that cppreference is _really_ hard to read.

DanRStevens: Admittedly it does take a bit of effort to learn how to read CppReference. They can be rather pedantic, though very detailed and accurate. They do sometimes have a bit of a high level explanation of stuff, though mostly they're going into the minute details, which are not always easy to read or understand. It's not quite as bad as the legalese you'd find in the C++ Standard, but it's a good way there. I find it's a good intermediate source of documentation.

DanRStevens:
> Benefit of reinitialize vs clear()?
> 
> Not opposed but I tend to use clear() to maintain consistent syntax with STL containers.

I think this aspect of the STL is rather old. Probably long predates the C++11 standard, where move constructors would become a thing. I think copy and move constructors are probably important for re-initializing in an efficient manner.

As for benefits, it largely comes down to maintenance. If you add new fields to a class, you may need to update the constructor, destructor, and any `clear` method, with corresponding changes. If you remove `clear` and instead re-initialize, it saves you from having to write and maintain a `clear` method. Imagine adding a pointer to dynamically allocated memory to a class, and forgetting to update `clear`. That would be very easy to do. It's much harder to forget about the constructor and destructor.

ldicker83:
Hmm. I hadn't thought of that. The cost of reinitializing really isn't an issue in these cases. I guess I'm still sorta stuck in my ways :D 

But yeah, that makes sense. We should probably consider removing 'clear' type functions from all of our classes that use it. The only exceptions I can think of are things like List Boxes where we may want a `dropAllItems` type of function versus `clear`.

DanRStevens:
I think I may have already removed many of them, though I'm sure there's plenty left.

It may make sense to retain `clear` like functionality when you're only doing a partial clear. Such as removing all items from a list box, but maintaining state on the sort column or sort order. Though the current list box doesn't support that, so I don't know that we really need to support some kind of `clear`.

Another thought I had regarding list box was to store the data separately from it, and then just have it be a view on externally managed data. That's a discussion for the list box cleanup issue though.

PR #1200, opened 2022-02-01
Account for IconIndex in updated research definition

ldicker83:


OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -115,7 +115,7 @@ namespace
 
 		try
 		{
-			reportMissingOrUnexpected(dictionary.keys(), {"id", "lab_type", "cost", "name"}, {"description", "requires", "effects"});
+			reportMissingOrUnexpected(dictionary.keys(), {"id", "lab_type", "cost", "name", "icon_index"}, {"description", "requires", "effects", "icon_index"});
```
DanRStevens, line 118: You added `icon_index` to both the required and optional lists. It should only be in one of them.

If you're trying to maintain compatibility with old files, then it should be optional. If this is tied in with a static data file update, then it should be required.

ldicker83: Ah, got it. I didn't realize one was optional (leave it to me to RTFM). Will fix in a different change set.

PR #1201, opened 2022-05-10
Implement structure availability

ldicker83:
Checks for unlocked structure technologies and simplifies `IconGrid` population. This doesn't close any issues but does add structure unlocks through research.

There are a few changes that could make this even cleaner, specifically with the `IconGrid::addItem()` function -- can use a basic struct like `StructureTracker::StructureItemList`, then remove the struct from `StructureTracker` and change `IconGrid::addItem()` to use the struct. Beyond the scope of this change set.

Something that occurred to me while I was working on this is structures superseding older designs and how to handle that. This can be done in a few ways but for me I think the simplest is to add a 'supersedes' parameter to the unlock type. Would mean modifying the tech editor (it needs some fixes anyway) and accepting that as an optional attribute in the reader function.

### DIFF
--- a/OPHD/States/MapViewStateUi.cpp
+++ b/OPHD/States/MapViewStateUi.cpp
@@ -28,6 +28,18 @@
 extern NAS2D::Point<int> MOUSE_COORDS;
 
 
+namespace
+{
+	void fillList(IconGrid& grid, const StructureTracker::StructureItemList& itemList)
+	{
+		for (const auto& item : itemList)
+		{
+			grid.addItem(item.name, item.sheetIndex, item.id);
+		}
+	}
+};
+
+
 /**
  * Sets up the user interface elements
  * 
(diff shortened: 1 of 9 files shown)


OPHD/States/StructureTracker.cpp
```diff
@@ -0,0 +1,62 @@
...
+	mDefaultUndergroundStructures.push_back({constants::Park, 75, StructureID::SID_PARK});
+	mDefaultUndergroundStructures.push_back({constants::UndergroundPolice, 61, StructureID::SID_UNDERGROUND_POLICE});
+	mDefaultUndergroundStructures.push_back({constants::RecreationCenter, 73, StructureID::SID_RECREATION_CENTER});
+	mDefaultUndergroundStructures.push_back({constants::Residence, 55, StructureID::SID_RESIDENCE});
+	mDefaultUndergroundStructures.push_back({constants::UndergroundFactory, 69, StructureID::SID_UNDERGROUND_FACTORY});
+	mDefaultUndergroundStructures.push_back({constants::MedicalCenter, 62, StructureID::SID_MEDICAL_CENTER});
+	mDefaultUndergroundStructures.push_back({constants::Nursery, 77, StructureID::SID_NURSERY});
+	mDefaultUndergroundStructures.push_back({constants::Commercial, 66, StructureID::SID_COMMERCIAL});
+	mDefaultUndergroundStructures.push_back({constants::RedLightDistrict, 76, StructureID::SID_RED_LIGHT_DISTRICT});
+	mDefaultUndergroundStructures.push_back({constants::University, 63, StructureID::SID_UNIVERSITY});
```
DanRStevens, line 49: Rather than use repeated `push_back` in a constructor, you can instead directly initialize the array with a list of items.

```cpp
mDefaultUndergroundStructures = {
  {constants::Laboratory, 58, StructureID::SID_LABORATORY},
  {constants::Park, 75, StructureID::SID_PARK},
  ...
};
```

Using direct initialization instead of repeated mutation with `push_back` means you can also potentially mark the data as `static const`. That would seem logical for the initial default lists, which I wouldn't expect to change.

OPHD/States/MapViewStateUi.cpp
```diff
@@ -334,17 +332,8 @@ void MapViewState::populateStructureMenu()
...
-		mStructures.addItem(constants::RecreationCenter, 73, StructureID::SID_RECREATION_CENTER);
-		mStructures.addItem(constants::Residence, 55, StructureID::SID_RESIDENCE);
-		mStructures.addItem(constants::UndergroundFactory, 69, StructureID::SID_UNDERGROUND_FACTORY);
-		mStructures.addItem(constants::MedicalCenter, 62, StructureID::SID_MEDICAL_CENTER);
-		mStructures.addItem(constants::Nursery, 77, StructureID::SID_NURSERY);
-		mStructures.addItem(constants::Commercial, 66, StructureID::SID_COMMERCIAL);
-		mStructures.addItem(constants::RedLightDistrict, 76, StructureID::SID_RED_LIGHT_DISTRICT);
-		mStructures.addItem(constants::University, 63, StructureID::SID_UNIVERSITY);
+		fillList(mStructures, mStructureTracker.defaultUndergroundStructures());
+		fillList(mStructures, mStructureTracker.unlockedUndergroundStructures());
```
DanRStevens, line 336: I would assume anything in the "default" list is also "unlocked". Maybe it makes sense to initialize the unlocked list to the default list, and then add to the unlocked list from there. Then you'd only need to add elements from a single "unlocked" list here.

OPHD/States/StructureTracker.cpp
ldicker83: I'm an idiot... I knew that too 😩

Will make that a separate change set since this one's already merged.

OPHD/States/MapViewStateUi.cpp
ldicker83: I wanted to keep them separate so that a list of structures that have been unlocked due to research can be maintained.

Then again, that can also be done using the `ResearchTracker` and the `StructureItemFromString` table (expanded later) to do the same thing when needed. I have a near compulsive need to cache things like this.

I think I'll go with that. Same as the last comment, will make that a separate change set.

OPHD/States/StructureTracker.cpp
ldicker83: So tried that, and MSVC crashes and GCC says no. Gonna leave this one alone for now.

PR #1206, opened 2022-05-11
Icon grid new add

### LINKED ISSUE #1203: Add IconGrid::addItem(IconGrid::Item)
Related to #1202

After the IconGrid::Item is added, the IconGrid::addItem should be overloaded to use the new type.

ldicker83:
Closes #1203

### DIFF
--- a/OPHD/UI/IconGrid.cpp
+++ b/OPHD/UI/IconGrid.cpp
@@ -156,7 +156,12 @@ void IconGrid::addItemSorted(const Item& item)
 
 void IconGrid::addItem(const Item& item)
 {
-	addItem(item.name, item.sheetId, item.meta);
+	mIconItemList.push_back(item);
+
+	const int x_pos = (item.sheetId % (mIconSheet.size().x / mIconSize)) * mIconSize;
+	const int y_pos = (item.sheetId / (mIconSheet.size().x / mIconSize)) * mIconSize;
+
+	mIconItemList.back().pos = {x_pos, y_pos};
 }
 
 
--- a/OPHD/UI/IconGrid.cpp
+++ b/OPHD/UI/IconGrid.cpp
@@ -174,8 +174,10 @@ void IconGrid::addItem(const Item& item)
  */
 void IconGrid::addItemSorted(const std::string& name, int sheetIndex, int meta)
 {
+	/*
 	addItem(name, sheetIndex, meta);
 	sort();
+	*/
 }
 
 
@@ -186,6 +188,7 @@ void IconGrid::addItemSorted(const std::string& name, int sheetIndex, int meta)
  *
  * \note	This function does no sanity checking.
  */
+/*
 void IconGrid::addItem(const std::string& name, int sheetIndex, int meta)
 {
 	int x_pos = (sheetIndex % (mIconSheet.size().x / mIconSize)) * mIconSize;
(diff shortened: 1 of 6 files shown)


OPHD/UI/IconGrid.h
```diff
@@ -26,18 +26,14 @@ class IconGrid : public Control
 	/**
 	 * Item used within the IconGrid.
 	 */
-	class Item
+	struct Item
 	{
 	public:
```
DanRStevens, line 31: A `struct` defaults to `public` access, so the `public:` access specifier can be removed with this change.

OPHD/UI/IconGrid.cpp
```diff
@@ -156,7 +156,12 @@ void IconGrid::addItemSorted(const Item& item)
 
 void IconGrid::addItem(const Item& item)
 {
-	addItem(item.name, item.sheetId, item.meta);
+	mIconItemList.push_back(item);
+
+	const int x_pos = (item.sheetId % (mIconSheet.size().x / mIconSize)) * mIconSize;
+	const int y_pos = (item.sheetId / (mIconSheet.size().x / mIconSize)) * mIconSize;
+
+	mIconItemList.back().pos = {x_pos, y_pos};
```
DanRStevens, line 158: No issues here, but I thought it might be interesting to bring up this thread:
[C++ Best way to get integer division and remainder](https://stackoverflow.com/questions/7070346/c-best-way-to-get-integer-division-and-remainder)

Mostly the optimization answer says not to worry about it, the compiler can figure out how to do it efficiently.

Of course the later answer about `divmod` is interesting, in that it could be used to give grid position, which could be scaled by `mIconSize` to get pixel offsets. This has me thinking maybe a `Vector` can be formed from the result of `divmod`, and then scalar multiplication can be used to get pixel offsets.

Of course, maybe `IconGrid` should calculate positions on the fly, rather than store pre-calculated values.

Maybe something to consider for future refactoring.

DanRStevens (commented):
I really like how you simplified the `IconGrid` interface to have a single `addItem` method. Very methodical how that code was transformed.

OPHD/UI/IconGrid.h
ldicker83: Wherps, forgot to remove that. That's why I changed it to a struct over a class.

OPHD/UI/IconGrid.cpp
ldicker83: Could be, but I like to have positional information pre-calculated so that it doesn't need to be redone every frame which can be very expensive especially with multiplication/division/modulus operations. Granted an optimizing compiler can make great work in making this better.

When considering refactoring we should do some benchmarking to see if there's a difference and, assuming there is, if it's worth unrolling it.

DanRStevens: I think the appeal of calculating on the fly, is that if all items are being drawn, then you can use a cheap incremental approach to the positioning. Whereas if you positioning just one at a time, then the multiplication and division becomes a necessity for each item.

At any rate, it's not really a now thing. It might make more sense to change it after potentially refactoring the drawing code.

ldicker83: So that's fair, I hadn't considered that. I tend to think in simpler terms and I have a weird need to pre-cache things. I guess it's a form of premature optimization.

PR #1213, opened 2022-05-24
Updated NAS2D submodule

DanRStevens:
Nothing major, just noticed OPHD was still depending on a version of NAS2D that was somewhat behind current. And I was looking for an excuse to update the default branch of OPHD since it was renamed to `main`. Apparently GitHub pages are unpublished after renaming the default branch, and you need a new push to the default branch to restore the pages.


DanRStevens:
Looks like the GitHub Pages site was successfully restored after the default branch rename to `main`.
https://outpostuniverse.github.io/OPHD/

Issue #1215, opened 2022-05-30
Refactor GameOptionsDialog constructor


matt-handy:
A few minor concerns with this implementation:
1) String constants are in file instead of in a string config file
2) Window size is statically defined instead of dynamically calculated based on window content
3) Button position has to manually enumerated, could be dynamically generated.

Assign to me, I'll take a crack at this.

ldicker83:
Agreed on all points. Happy to have you take a look.

matt-handy:
Closing out as the PR is merged

PR #1218, opened 2022-06-21
1215 revised

matt-handy:
Revised code changes for #1215

### DIFF
--- a/OPHD/UI/GameOptionsDialog.h
+++ b/OPHD/UI/GameOptionsDialog.h
@@ -20,6 +20,12 @@ public:
 	ClickSignal::Source& returnToMainMenu() { return mSignalClose; }
 
 private:
+	const int buttonHeight = 25;
+	const int buttonWidth = 200;
+	const int buttonHorizontalMargin = 5;
+	const int buttonCloseVerticalMargin = 3;
+	const int buttonLargeVerticalMargin = 13;
+
 	void onLoad();
 	void onSave();
 	void onHelp();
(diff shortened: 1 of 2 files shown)


DanRStevens (approved):

DanRStevens:
Huh, I must have missed this one earlier.

OPHD/UI/GameOptionsDialog.h
```diff
@@ -20,6 +20,11 @@ class GameOptionsDialog : public Window
 	ClickSignal::Source& returnToMainMenu() { return mSignalClose; }
 
 private:
+	const int buttonHeight = 25;
+	const int buttonWidth = 200;
+	const int buttonHorizontalMargin = 5;
+	const int buttonVerticalMargin = 3;
```
DanRStevens, line 26: This code is perfectly fine, though I wanted to bring up an alternative option which I learned over the years and found to be quite useful.

Instead of putting implementation constants in the `private` section of the header file (which if changed, forces recompile of all including files), I instead like to put implementation constants in an unnamed namespace in the implementation (.cpp) file. That way if the constants change, only that one implementation file needs to be re-compiled, rather than all files that included the header.

To add some additional detail, putting things in an unnamed namespace forces internal linkage. That makes it impossible to reference the values from other translation units (.cpp files). You might think of this a bit like pre-linking that code and then not exporting a symbol for that variable/function/class in the compiled output. There is no way to reference the item from another compilation unit when the object files are linked together. Even if a symbol was exported, it's in a namespace, so you'd need to specify a namespace prefix, and the namespace is unnamed, so there is nothing you can specify to uniquely identify the namespace.

Documentation:
https://en.cppreference.com/w/cpp/language/namespace
(See section on "Unnamed namespaces")

For examples:
- [DetailMap.cpp](https://github.com/OutpostUniverse/OPHD/blob/main/OPHD/UI/DetailMap.cpp)
- [TileMap.cpp](https://github.com/OutpostUniverse/OPHD/blob/main/OPHD/Map/TileMap.cpp)

PR #1224, opened 2022-07-28
Fixed build issues introduced by warehouse notifications

Inglonias:
I actually bothered to build the project to ensure it works this time. Screenshots from testing are below.

PR #1226, opened 2022-07-30
Added a proper cheat menu

Inglonias:
Previously, pressing Ctrl + Shift + F10 would add 1K of each mined resource.

Now, it brings up a simple window with a text box. You can type various codes for various effects.

| Cheat Code | Effect |
| ------------- | ------------- |
| "goldrush"  |  Add 1000 of each resource  |
| "orderpizza"  |  Fill your colony's food storage  |
| "waveofbabies"  |  Add ten children to the population  |
| "savedbythebell"  |  Add ten students to the population  |
| "gettowork"  |  Add ten workers to the population  |
| "smartypants"  |  Add ten scientists to the population  |
| "electronicoldmen"  |  Add ten retired colonists to the population  |
| "beepboop"  |  Add a RoboDigger, RoboMiner, and RoboDozer to the robot pool  |

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -452,7 +452,7 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 		case NAS2D::EventHandler::KeyCode::KEY_F10:
 			if (NAS2D::Utility<NAS2D::EventHandler>::get().control(mod) && NAS2D::Utility<NAS2D::EventHandler>::get().shift(mod))
 			{
-				mCheatmenu.show();
+				mCheatMenu.show();
 				//StorableResources resourcesToAdd{1000, 1000, 1000, 1000};
 				//addRefinedResources(resourcesToAdd);
 				//updatePlayerResources();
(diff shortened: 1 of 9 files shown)


Inglonias:
Note that this PR was branched off of #1224 to allow it to actually build. Please merge #1224 before merging this!

Inglonias:
One issue I'm seeing is that if I press the WASD keys when entering cheat codes, the main window camera moves.

ldicker83:
> One issue I'm seeing is that if I press the WASD keys when entering cheat codes, the main window camera moves.

This can be addressed by using a bit of a kludge in the `onKeyDown()` handler starting on line 377: https://github.com/Inglonias/OPHD/blob/hackerman/OPHD/States/MapViewState.cpp#L377

Can add a check via `Control::hasFocus()`. Since you'd probably just want to filter by the movement keys vs other keys, you could add something along the lines of:

```cpp
if(mCheatWindow.visible() && mCheatWindow.hasFocus()) { break; }
```

Not the greatest code ever but does the job

ldicker83:
So I looked closer into this and the problem is actually in the way the textfield's and MapViewState interact with each other and how textinput mode is being set. I'm working on a separate fix that will address this. I would disregard the suggestion of the hacky method mentioned above.

I'm going to implement that as a different change set so that this will just work without needing additional change. Will be reviewing this changeset shortlyl

ldicker83:
Okay, so far this looks pretty good. A few minor changes requested:

- There should be two lines between function/class/namespace blocks vs one.
- I'd suggest putting the CheatMenu enum class into the unnamed namespace in the source file unless you think future plans would have this be used elsewhere in the code
- The window should be centered when it's shown

After these changes I think this will be good to merge

Inglonias:
The enum is used elsewhere in the code - In the MapViewStateUi.cpp file. I can't move it to the source file without breaking things. Or if I can, I don't have enough knowledge of C++ to do it "correctly", whatever that means.

Inglonias:
I have made all other suggested changes.

ldicker83:
Looks good to me! Going to go ahead and merge this into main.

PR #1248, opened 2022-08-20
Use packed params

### LINKED ISSUE #217: Use Point/Vector for `(x, y)` parameters

DanRStevens:
Closes #217.

### DIFF
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -365,15 +365,15 @@ std::vector<std::string> splitString(const std::string& string, char delimiter)
 }
 
 
-void drawBasicProgressBar(int x, int y, int width, int height, float percent, int padding)
+void drawBasicProgressBar(NAS2D::Rectangle<int> rect, float percent, int padding)
 {
 	auto& renderer = Utility<Renderer>::get();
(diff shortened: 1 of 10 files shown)


DanRStevens:
The `StructureListBox` and `FactoryListBox` changes could probably use more thorough vetting. I didn't have a convenient saved game around with more than one entry in the list box.

DanRStevens:
I spent a bit of time building out a colony, and the list box updates seem to be fine.

PR #1252, opened 2022-08-20
Remove `\fixme` or replace with "FIXME:"

DanRStevens:
Reference: #307

Add Clang warning flag `-Wdocumentation-unknown-command`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ jobs:
     docker:
       - image: outpostuniverse/nas2d-clang:1.3
     environment:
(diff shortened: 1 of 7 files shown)


ldicker83:
Those were used with Doxygen .... got used to it with NAS2D but it seems to be unnecessary with OPHD since it's not a library.

DanRStevens:
I figured, but that's not actually a Doxygen command:
https://www.doxygen.nl/manual/commands.html

Maybe one that was removed? Maybe a proposed one that was never adopted? I'm not really sure what happened there. At any rate, there where already a number of "FIXME:" comments in the code, so I went with that.

ldicker83:
I think I was originally using \todo and then opted to use \fixme for myself. I dunno.

Either way, no objections :D

PR #1257, opened 2022-08-21
Solar distance - check once

DanRStevens:
We should check `meanSolarDistance` against `0` once during initialization, rather than every time it is used.

Perhaps even better would be to `throw` from the constructor if the passed value was ever `0`.

Would be nice if the check could be done without a floating point comparison (#307, `-Wfloat-equal`), though that get's awkward. It might be possible to use [`std::equal_to`](https://en.cppreference.com/w/cpp/utility/functional/equal_to), though that's getting a bit esoteric.

It might also make sense to set the default value to `1.0` instead of `0`. Of course, that doesn't really eliminate the need for the check, since when the value is set, it could be set to `0`.

### DIFF
--- a/OPHD/Things/Structures/SolarPanelArray.h
+++ b/OPHD/Things/Structures/SolarPanelArray.h
@@ -11,11 +11,15 @@ const int SOLAR_PANEL_BASE_PRODUCUCTION = 50;
 class SolarPanelArray : public PowerStructure
 {
 public:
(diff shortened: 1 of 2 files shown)


ldicker83 (commented):
I love this method of handling initialization. Looks a lot like it does in C# and something I wished we could do in C++. Glad to see it's now correct syntax :D

Issue #1269, opened 2022-08-27
Crash when clicking "Load" for saved games


DanRStevens:
The game crashes with a segmentation fault when the "Load" button is clicked, while trying to load a game from the main menu.

The problem is only present when selecting a file with single click, and then clicking on the "Load" button. The load functionality works normally when double clicking on the saved game file, as well as when selecting a file with single click, and then pressing the "Enter" key to load it.

The crash only happens when the game is freshly started up, and has not yet started a new game or loaded a game. The problem does happen if you click to start a new game, but return to the main menu instead of selecting a planet, and proceed as above.

----

The crash appears to happen in event handling code from within `DelegateX::operator()`, which seems to be called on an invalid or corrupted `DelegateX` object with improper `this` or member function pointer.

At the time of the crash, two event handlers are called. The first one appears to be the "Load" button event handler, which runs `FileIo::onFileIo()`, and completes without error. The second event handler is the one that causes the crash.

I suspect there were two registered click listeners (not verified), and the second one is for a screen which shouldn't be responding to input at that time.

ldicker83:
As an aside, this doesn't occur after loading a game once and backing out to the main menu and going through these steps.

ldicker83:
So I'm looking over the code and neither the MapViewState nor the MainMenuState disconnect themselves from the FileIo object... I'm not sure if this is handled automatically anywhere but if not this signal should be explicitly disconnected. Might resolve the issue (and explain the occasional heap corruption I come across).

DanRStevens:
The crash appears to be coming from the `TextControl` signal `mTextChanged`.

DanRStevens:
I think I might know what's going on here. I think it's a memory corruption issue from invalidated iterators.

The `Signal::emit` method iterates over the `delegateList`, and calls the callbacks. Some of those callbacks may install new callbacks on the same signal handler. When they call `connect` on the same signal handler, the underlying `std::vector` may need to re-allocate to add the new `DelegateX` entry. That re-allocate invalidates the previous iterators. So when control returns back to the `Signal::emit` method, and it continues trying to call subsequent iterators, it's doing it over the old de-allocated `std::vector` memory. Since that memory is now freed, things are able to overwrite it, and so it ends up reading corrupted memory.

PR #1271, opened 2022-08-28
Preconstruct `DelegateX` before calling `connect`/`disconnect`

DanRStevens:
This means fewer overloaded methods will be needed on the `SignalSource` object. It also more tightly associates `this` pointers with member function pointers by composing them into an obvious compound object.

One reason to want to reduce `connect` overloads, is to add debugging capabilities to the `connect` method. This is easier when only a single overload needs to be updated.

Reference: Issue #1269


DanRStevens:
I should probably mention, this was accomplished with a global search/replace regex:
Search: `connect\(([^{].*)\)`
Replace: `connect({$1})`

Note that the above updates both `connect` and `disconnect` calls.

Issue #1235, opened 2022-08-11
"Help" button on in-game system menu is not disabled when returning to main menu


ldicker83:
As the title states. This button should be disabled with the others to prevent possible weird/undefined behavior when the game state is changing.

<img width="1199" alt="Screen Shot 2022-08-10 at 11 14 50 PM" src="https://user-images.githubusercontent.com/13858902/184059309-6b9d3876-e344-49b7-86ad-a29307407fe7.png">

Note in screenshot that HELP button is not greyed out and can still be clicked on.

PR #1278, opened 2022-09-01
Disable help button with the others

belgianguy:
Should solve #1235, I also saw the help button was not disconnected in the GameOptionsDialog.cpp destructor, so I also added that, but that was perhaps not necessary? I put them in different commits.


DanRStevens (approved):
Looks good to me.

PR #1279, opened 2022-09-01
Retargets project to VS2022 build tools and C++20

ldicker83:


DanRStevens:
Oh, I didn't even see this PR. I think this is an earlier version of the change done in #1280.

ldicker83:
It is. Going to close it now.

PR #1282, opened 2022-09-02
Add remove population functions to cheat menu

belgianguy:
First attempt at #996, please review. 

I also noticed there is no destructor for CheatMenu, and no "disconnect" for the btnOkay, is that something that should be added or is that handled automagically?

### DIFF
--- a/OPHD/Population/Population.cpp
+++ b/OPHD/Population/Population.cpp
@@ -46,6 +46,11 @@ void Population::addPopulation(const PopulationTable& population)
 	mPopulation += population;
 }
 
+void Population::removePopulation(const PopulationTable& population)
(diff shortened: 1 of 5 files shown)


DanRStevens (approved):

DanRStevens:
The `Button` handler doesn't need to be disconnected in that context. When the `CheatMenu` is destructed, the contained `Button` will be destructed, which causes the contained `Signal` object to be destructed, which will `disconnect` any callbacks. The target of the callbacks is the actual real concern for dangling callbacks. If instead the target of the callback was being destroyed, then you would need to manually call `disconnect` for any `Signal` that would send messages to it.

PR #1284, opened 2022-09-08
Consolidate Structure/ProductCatalog init calls

ldicker83:
Noticed some inefficiencies in how the `ProductCatalogue::init()` and `StructureCatalogue::init()` functions were being called so I consolidated those calls and simplified the conditional check in `MapViewState::initialize()`.

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -228,17 +228,13 @@ void MapViewState::initialize()
 
 	mPopulationPool.population(&mPopulation);
 
+	StructureCatalogue::init(mPlanetAttributes.meanSolarDistance);
+	ProductCatalogue::init("factory_products.xml");
+
 	if (mLoadingExisting)
 	{
 		load(mExistingToLoad);
 	}
-	else
-	{
-		// StructureCatalogue is initialized in load routine if saved game present to load existing structures
-		StructureCatalogue::init(mPlanetAttributes.meanSolarDistance);
-	}
-
-	ProductCatalogue::init("factory_products.xml");
 
 	setupUiPositions(renderer.size());
 	resetPoliceOverlays();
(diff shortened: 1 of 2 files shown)


OPHD/States/MapViewState.cpp
```diff
@@ -228,17 +228,13 @@ void MapViewState::initialize()
...
+	ProductCatalogue::init("factory_products.xml");
+
 	if (mLoadingExisting)
 	{
 		load(mExistingToLoad);
 	}
-	else
-	{
-		// StructureCatalogue is initialized in load routine if saved game present to load existing structures
-		StructureCatalogue::init(mPlanetAttributes.meanSolarDistance);
```
DanRStevens, line 238: I believe `mPlanetAttributes` isn't guaranteed to be initialized until after the call to `load`. There are two constructors to `MapViewState`. One of them initializes `mPlanetAttributes`, while the other sets `mLoadingExisting` and defers initialization to the `load` call.

It's weird, and probably should be changed, but if that hasn't been accounted for, this may be introducing a bug.

OPHD/States/MapViewStateIO.cpp
```diff
@@ -285,8 +285,6 @@ void MapViewState::load(const std::string& filePath)
 
 	difficulty(stringToEnum(difficultyTable, dictionary.get("difficulty", std::string{"Medium"})));
 
-	StructureCatalogue::init(mPlanetAttributes.meanSolarDistance);
-	ProductCatalogue::init("factory_products.xml");
```
DanRStevens, line 289: Look up a few lines, and you'll see where `mPlanetAttributes` is initialized.

ldicker83: Huh, I thought I did a call trace to make sure that wouldn't be an issue. I'll take a closer look tonight when I get home from work.

PR #1287, opened 2022-09-15
Use `const` list-init default available structs

DanRStevens:
Reference: https://github.com/OutpostUniverse/OPHD/pull/1201#discussion_r869874118

Use `std::vector` list initialization rather than repeated `push_back`, and mark the list as `const` data.

This change used to crash MSVC. For example:
https://github.com/OutpostUniverse/OPHD/actions/runs/2898105709/jobs/4610071913
> D:\a\OPHD\OPHD\OPHD\States\StructureTracker.cpp(7): fatal error C1001: Internal compiler error. [D:\a\OPHD\OPHD\OPHD\ophd.vcxproj]
         LINK : fatal error LNK1000: Internal error during IMAGE::BuildImage [D:\a\OPHD\OPHD\OPHD\ophd.vcxproj]

It seems the updated version of MSVC is now able to handle this code. We may want to delay or avoid merging this change in case we still have developers using older versions of MSVC, or want to support older versions of MSVC.

### DIFF
--- a/OPHD/States/StructureTracker.cpp
+++ b/OPHD/States/StructureTracker.cpp
@@ -4,6 +4,37 @@
 
 namespace
 {
+	const std::vector<IconGrid::Item> DefaultAvailableSurfaceStructures = {
+		{constants::Agridome, 5, StructureID::SID_AGRIDOME},
+		{constants::Chap, 3, StructureID::SID_CHAP},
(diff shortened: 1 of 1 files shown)


ldicker83:
Don't want to support older versions. The updates to MSVC are generally automatic. Went ahead and merged this.

PR #1293, opened 2023-03-30
Don't glow resource/info values on first turn

ldicker83:


ldicker83:
Failing builds are because vcpkg broke sdl2-mixer somehow. Not sure it can be fixed on our end.

ldicker83:
Okay, figured it out. vcpkg pushed updates that makes some of the triplicate parameters unnecessary. Updating nas2d-core to fix this.

OPHD/UI/ResourceInfoBar.cpp
```diff
@@ -145,7 +150,7 @@ void ResourceInfoBar::draw() const
 	for (const auto& [imageRect, amount, spacing] : resources)
 	{
 		renderer.drawSubImage(mUiIcons, position, imageRect);
-		const auto color = (amount <= 10) ? glowColor : NAS2D::Color::White;
+		const auto color = (amount <= 10 && !mIgnoreGlow) ? glowColor : NAS2D::Color::White;
```
DanRStevens, line 153: Was just thinking, rather than having a hardcoded resource threshold of "10" to enable glowing, maybe the threshold should be the parameter. The threshold can be set low initially, to prevent glowing, and increased after a turn is taken. For late game, maybe having more resources in storage will be important to keep buildings operational, and so a larger threshold might be desired.

ldicker83: I would agree with this sort of a change. Definitely something to consider moving forward.

PR #1294, opened 2023-03-31
Add xcode project

ldicker83:
Add xcode project to build macOS application bundles


DanRStevens:
Hmm, looks like it imported dependency source files into our repository. That's a bit of an archaic way to handle dependencies. Normally we use `brew` on MacOS. I would assume Xcode would have some support for that sort of a setup. Maybe there's a way to adjust to use `brew`?

Nice to get application bundling working though.

ldicker83:
That's how macOS framework bundles work. Brew is used for building on mac from the terminal. To distribute an app Xcode is needed to produce an app bundle (e.g., OutpostHD.bundle). Frameworks are the dependencies included with the app bundle to eliminate 'dependency hell' on macOS platforms.

Leads to much larger 'binaries' but it works /me shrugs. Apple does what Apple wants.

Ultimately for the end user it's as simple as downloading a disk image (.dmg file), mounting it and then dragging the bundle into the Applications folder. Extremely simple and the general way applications are distributed on mac's outside of the app store.

PR #1296, opened 2023-04-02
Fix Gameover dialog button not going back to main menu

ldicker83:


OPHD/States/MapViewState.cpp
```diff
@@ -306,8 +313,7 @@ NAS2D::State* MapViewState::update()
 		renderer.drawBoxFilled(windowClientRect, NAS2D::Color::Black);
 		mGameOverDialog.update();
 
-		mFade.update();
-		mFade.draw(renderer);
+		updateFade(renderer, mFade);
```
DanRStevens: There was a vague plan to split `update` from `draw` (which can be marked `const`). That would necessitate splitting of the two calls. Recurse on `update` with mutable objects, then recurse on `draw` with `const` objects.

It's probably wise to separate mutable update code from rendering code, which is not expected to update state.

Anyway, it's a future concern. No need to worry about it here.

ldicker83: Also agreed. I've was originally opposed to this but it makes sense to separate those out. Drawing isn't the same as logic updates after all.

PR #1298, opened 2023-04-06
Fix missing header

ldicker83:
Update to Xcode uses strict checking. <string> wasn't included in the header file which caused it to fail the build. Seems to not be an issue on other platforms/compilers but it doesn't hurt to have everything explicitly defined.


ldicker83:
Seems I accidentally pulled in some additional changes that add a Release build for Xcode.

PR #1303, opened 2023-04-11
Fix `meanSolarDistance` init bug

DanRStevens:
Reference:
https://github.com/OutpostUniverse/OPHD/pull/1284#pullrequestreview-1108976743

### DIFF
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -10,6 +10,8 @@
 #include <NAS2D/Xml/XmlDocument.h>
 #include <NAS2D/Xml/XmlElement.h>
 
+#include <stdexcept>
+
 
 using namespace NAS2D;
 
@@ -265,6 +267,10 @@ std::vector<std::string> splitString(const std::string& string, char delimiter)
 
 void setMeanSolarDistance(float newMeanSolarDistance)
 {
+	if (newMeanSolarDistance <= 0)
+	{
+		throw std::runtime_error("Must set a positive value for `meanSolarDistance`: " + std::to_string(newMeanSolarDistance));
+	}
 	meanSolarDistance = newMeanSolarDistance;
 }
 
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -14,6 +14,12 @@
 using namespace NAS2D;
 
 
+namespace
+{
+	float meanSolarDistance = 1;
+}
+
+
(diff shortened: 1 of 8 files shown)


ldicker83 (commented):
So my thinking is to have gameplay attributes in a 'PlanetAttributes' or 'EnvironmentAttributes' sort of struct or class (leaning toward struct) versus having these values floating around as globals.

DanRStevens:
I agree. I was thinking once the `MapViewState` class was refactored and split up, there would be a more obvious class to use for global game state, which could be retrieved from a global function, rather than just one specific value for `meanSolarDistance`.

Also, I was thinking maybe we should change the `meanSolarDistance` value to something like `solarIrradiation`. Basically, have it be a multiplicative factor, rather than a divisor. Plus, irradiation would be a more natural measure of how much solar energy is arriving, rather than distance alone, as power depends on both distance and intensity.

Brett208:
Thanks for fixing what was probably sloppy initialization on my part back when this was added...

I had thought solar distance was a nice inbetween property that would mean something to the less technical player about how effective solar power would be. My general thought was it might be cool to create a baseball card for a given planet that gave some major figures such as solar distance but never really discussed.

Solar Irradiance as calculated at Earth is typically called the Solar Constant. You could maybe define a Solar Constant for each planet. This doesn't take into effect atmospheric effects and the day/night cycle... you could just boil it down to a solar power effectiveness rating that is low, medium, high and forget the rest of the physics would be easier and even ditch the mean solar distance. Although I think it could be fun to define the properties for a given sun/planet if it fits into the plan.

DanRStevens:
No problems with the original code. The bug was introduced recently during some refactoring in the referenced PR.

Might be nice to include more planetary stats when selecting a planet.

ldicker83:
It's about time to start getting that stuff added into the game.

Properties of the planet and host star are definitely intended to be part of the game (see this thread: https://forum.outpost2.net/index.php/topic,5990.0.html). Basically the game setup would have the player start by choosing a star type, then a planet type. There would be some default setups that come with the game and then the option to 'build your own' by choosing attributes from a set of available options for both star and planet. These would affect things like solar radiation (including whether or not it's variable), gravity, scarcity of metals/minerals, and so on.

These options aren't well defined yet, could use a bit of help with that. Though that would naturally be in a different discussion :D

PR #1304, opened 2023-04-12
Update progress bar drawing code

DanRStevens:
Pass integers for `value` and `max` instead of a float for `percent` (which wasn't technically a percent, since there was no scale factor of 100). This eliminates the need to cast to and from `float` values, both at the call site, and internally within the function.

Add internal check to prevent division by 0. This no longer needs to be done at the call site.

Add internal clipping to prevent ever drawing a progress bar outside of the surrounding border.

Eliminate strange offsets to the inner rect position.


ldicker83:
This looks good to me. It's a lot cleaner too.

PR #1305, opened 2023-04-14
Use win32 API ShellExecute for opening explorer window

ldicker83:
The original method of win32 shell execute used `"open"` which didn't work well (or at all in the case of path's with spaces in the name), with a terminal window opening to execute the command and then closing.

This change set replaces the Windows code with win32 `shellapi` calls according to API documentation provided by Microsoft ([see documentation here](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecutea)). Ensures that the correct behavior is achieved and that it's seamless for the user.

### DIFF
--- a/OPHD/ShellOpenPath.cpp
+++ b/OPHD/ShellOpenPath.cpp
@@ -8,7 +8,7 @@
 namespace
 {
 	#if defined(_WIN32)
-		#include <Windows.h>
+		#include <windows.h>
 		#include <shellapi.h>
 	#elif defined(__APPLE__)
 		constexpr const std::string_view ShellOpenCommand{"open"};
--- a/OPHD/ShellOpenPath.cpp
+++ b/OPHD/ShellOpenPath.cpp
@@ -23,7 +23,7 @@ namespace
 void shellOpenPath(const std::string& path)
 {
 	#if defined(_WIN32)
-		std::ignore = ShellExecute(NULL, L"explore", std::wstring(path.begin(), path.end()).c_str(), NULL, NULL, SW_NORMAL);
+		std::ignore = ShellExecuteA(NULL, "explore", path.c_str(), NULL, NULL, SW_NORMAL);
 	#else
 		// Explicitly ignore implementation defined return value
 		std::ignore = std::system((std::string{ShellOpenCommand} + " " + path).c_str());
--- a/OPHD/ShellOpenPath.cpp
+++ b/OPHD/ShellOpenPath.cpp
@@ -8,7 +8,8 @@
 namespace
 {
 	#if defined(_WIN32)
-		constexpr const std::string_view ShellOpenCommand{"start"};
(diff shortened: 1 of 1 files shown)


ldicker83:
Okay so unsurprisingly linux-mingw fails as the Windows SDK is likely not installed or available in linux environments. Not sure how to get around this as building windows software on Linux still doesn't make too much sense to me.

The previous code did allow this code to compile on linux via mingw which is fine but I plan to adjust the code for macOS as well to use its API as it also fails to open a folder via the shell open command.

OPHD/ShellOpenPath.cpp
```diff
@@ -8,7 +8,8 @@
 namespace
 {
 	#if defined(_WIN32)
-		constexpr const std::string_view ShellOpenCommand{"start"};
+		#include <Windows.h>
+		#include <shellapi.h>
```
DanRStevens, line 12: > OPHD/ShellOpenPath.cpp:11:12: fatal error: Windows.h: No such file or directory
   11 |   #include <Windows.h>

I assume it's because the filesystem is case sensitive on Linux. Maybe try a lowercase `w`.

OPHD/ShellOpenPath.cpp
```diff
@@ -21,6 +22,10 @@ namespace
 
 void shellOpenPath(const std::string& path)
 {
-	// Explicitly ignore implementation defined return value
-	std::ignore = std::system((std::string{ShellOpenCommand} + " " + path).c_str());
+	#if defined(_WIN32)
+		std::ignore = ShellExecuteA(NULL, "explore", path.c_str(), NULL, NULL, SW_NORMAL);
```
DanRStevens, line 26: This could probably be improved so it's not hardcoded to use `ShellExecuteA`. Though we do have a hardcoded `MessageBoxA` already.

You may want to test this, as I'm not certain the encoding being passed will match what's expected on Windows.

DanRStevens:
In a related matter, we may have a couple of other places in the code that uses the wrong case for `#include <windows.h>`. It's possible the MingW build doesn't compile the alternate code path.

OPHD/ShellOpenPath.cpp
ldicker83: ShellExecute (at least in visual studio) is used as an alias between ShellExecuteA and ShellExecuteW -- on my system ShellExecuteW is what it expands to in order to handle unicode paths. It appears to work on this end, and is used as described in MS's documentation. Someone on a non-US English rig would provide a good test case.

ldicker83:
Based on my comment in [NAS2D PR 1088](https://github.com/lairworks/nas2d-core/pull/1088), we should be using `_WIN32` for checking Windows compiles. Mingw and MSVC both defined these internally which is why this hasn't been triggered until now when I introduced the new code.

PR #1306, opened 2023-04-14
Allow zero-max progress bars to draw as empty bars

ldicker83:
This allows progress bars with a zero-value 'max' component to draw as just an empty bar. Fixes an exception in the Warehouse Report.

@DanRStevens, want your opinion on this one. Another option is to just not draw the bar at all in the case that a zero-value for max is found which would require some logic elsewhere in the code to be modified. If that's the option we choose to go with this PR can be deleted.

### DIFF
--- a/OPHD/Common.cpp
+++ b/OPHD/Common.cpp
@@ -284,18 +284,16 @@ float getMeanSolarDistance()
 
 void drawProgressBar(int value, int max, NAS2D::Rectangle<int> rect, int padding)
 {
-	if (max == 0)
-	{
-		throw std::runtime_error("Progress bar must have non-zero max value: " + std::to_string(max));
-	}
-
 	const auto clippedValue = std::clamp(value, 0, max);
-	auto innerRect = rect.inset(padding);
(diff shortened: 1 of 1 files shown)


DanRStevens:
Oh, I forgot to comment on this one.

I have some reservations about the change. It does seem like an error to attempt to draw a scroll bar with a `max` of 0. In that sense, I might prefer an exception. Though I assume in a typical uninitialized case, if the `max` is 0, the `value` is also 0.

At any rate, I don't really object to swallowing the exception here, and making scroll bar drawing a little safer and thought free. Though it may make sense to still at least draw the outer border, and perhaps treat this case like a 0% progress case.

Anyway, I don't really have a saved game setup with a warehouse to test the change. I suppose this is where unit tests and such could have come in handy.

ldicker83:
In this case I came across the exception being thrown with a new game that didn't have a warehouse. I hit F1 to bring up the fullscreen UI and went clicking through the different panels.

Don't need to use F1 to replicate, can double click on a factory or mine to do the same thing. Still would like to find a better way to do this.

PR #1308, opened 2023-04-17
Update git ignore xcode

ldicker83:


.gitignore
```diff
@@ -22,3 +22,5 @@ x64/
 *.DS_Store
 /Build/
 /xcode/OutpostHD.xcodeproj/xcuserdata/
+/xcode/OutpostHD.xcodeproj/project.xcworkspace/xcuserdata/leeordicker.xcuserdatad
```
DanRStevens, line 25: This looks slightly too specific for version control. I would recommend setting the ignore to stop at the level just before your username, assuming the whole folder should be ignored, or set the ignore for files with that extension. The username portion is expected to be variable, based on who builds the project, so a proper solution should work for anyone.

ldicker83: I'm not sure why I hadn't thought of that. Will do that in a coming changeset.

PR #1309, opened 2023-04-17
Enquote path value to all open command to properly handle paths with spaces

ldicker83:
This fixes finder windows not opening with the 'open' shell command. I have not tested this on linux so I'd want your input @DanRStevens before merging this change.


DanRStevens (approved):
Confirmed this does indeed fix the function for paths with spaces.

Kind of had to hack the game a bit to get it to attempt to open a path with spaces.

ldicker83:
Thanks for checking!

At least now the game is properly opening folders in all tested OS's. Will need to get that pushed into a release asap, it's a pretty useful function.

I did note some text overflow on macOS X (has a long filepath). Not sure how useful having the file path is so I'm thinking of just removing the filepath text and adjusting the location of the button. Different changeset, naturally.

PR #1312, opened 2023-04-27
Tidy file io

### LINKED ISSUE #1310: Tidy FileIO Window
I mentioned in a comment in #1309 that I was seeing text overflow in the filepath for the savegames. It occurred to me that the text really doesn't need to be there, with the open savegame folder button, it becomes sort of a moot point.

So basically I want to remove the text, move the button to the bottom of the window on the left side. Going to play a bit with the layout and see what looks/feels the best here.

ldicker83:
Closes #1310

### DIFF
--- a/OPHD/UI/FileIo.cpp
+++ b/OPHD/UI/FileIo.cpp
@@ -32,8 +32,6 @@ FileIo::FileIo() :
 
 	size({700, 350});
 
-	add(mLabelFilePath, {5, 25});
-	mLabelFilePath.size({600, 20});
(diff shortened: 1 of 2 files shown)


ldicker83:
So I'm seeing that our dependency installer script is breaking in nas2d-core _once again_. Not sure why this time.

ldicker83:
Turns out the latest version of vcpkg expects all library names to be in lowercase. Last commit in this change set updates nas2d-core to the latest version with a correct script.

PR #1313, opened 2023-04-28
Update `.gitignore` to remove user specific part

DanRStevens:
Reference:
https://github.com/OutpostUniverse/OPHD/pull/1308#discussion_r1169005756

Assuming this is correct, and still ignores everything that needs to be ignored.


ldicker83:
Looks good to me.

PR #1319, opened 2023-04-30
Update Truck movement cost algorithm for Mines and Smelters

DanRStevens:
Instead of checking for path start/end points (which will always be a mine and smelter), instead check for tiles with Mines or Smelters. This makes the cost at each tile depend only on tile info. It does change the algorithm slightly, as a route from a mine can now pass through another mine. Previously only other roads were passable, aside from the start and end points.

(Note: No paths passing through another smelter would be a shortest path to a smelter.)


ldicker83 (approved):
This makes sense to me. Easier to understandl

PR #1320, opened 2023-05-01
Refactor path finding cost calculation

DanRStevens:
Related to #1319, as that changes enables additional simplification in this set of changes.


ldicker83:
Definitely like this change. There were a few things that threw me off because of GitHub's diff-ing engine but looks good to me!

PR #1323, opened 2023-05-01
Update keyboard movement

ldicker83:
Changes keyboard bindings so that movement in any direction is more intuitive.

### DIFF
--- a/OPHD/Common.h
+++ b/OPHD/Common.h
@@ -46,7 +46,11 @@ enum class Direction
 	East,
 	West,
 	North,
-	South
+	South,
+    NorthWest,
+    NorthEast,
+    SouthWest,
+    SouthEast
 };
 
 
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -405,22 +405,22 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 	{
 		case NAS2D::EventHandler::KeyCode::KEY_w:
 		case NAS2D::EventHandler::KeyCode::KEY_UP:
-			mMapView->moveView(Direction::North);
+			mMapView->moveView(Direction::NorthWest);
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_s:
 		case NAS2D::EventHandler::KeyCode::KEY_DOWN:
-			mMapView->moveView(Direction::South);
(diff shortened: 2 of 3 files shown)


OPHD/Common.h
```diff
@@ -46,7 +46,11 @@ enum class Direction
 	East,
 	West,
 	North,
-	South
+	South,
+    NorthWest,
+    NorthEast,
+    SouthWest,
+    SouthEast
```
DanRStevens, line 53: Looks like spaces were used to indent here.

OPHD/States/MapViewState.cpp
```diff
@@ -405,22 +405,22 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 	{
 		case NAS2D::EventHandler::KeyCode::KEY_w:
 		case NAS2D::EventHandler::KeyCode::KEY_UP:
-			mMapView->moveView(Direction::North);
+			mMapView->moveView(Direction::NorthWest);
```
DanRStevens, line 408: You know, if we updated the parameter type here, we could change this from:
```cpp
mMapView->moveView(Direction::NorthWest);
```
To:
```cpp
mMapView->moveView(DirectionNorthWest);
```

Same readability, but without needing an enum to offset lookup.

DanRStevens (commented):
Yes! This is so much more intuitive. Oddly, I wouldn't have really thought to move diagonally like that, so that it appears to move straight. This is so much better.

OPHD/States/MapViewState.cpp
DanRStevens: Ahh, but there's also the up/down buttons handled by that function.

ldicker83:
After not having played the game for awhile it threw me off a bit and then I remembered you talking about how it wasn't very intuitive. So I figured I'd modify it a bit :D Means it ultimately moves '2 tiles' versus just one but it seems to make more sense. Since we both had that experience others must have as well.

OPHD/Common.h
ldicker83: Ooof, yeah, that's a change in VS2022's latest things, I've been using the spaces over tabs lately in new projects. Will have to update the text editor to keep this consistent.

OPHD/States/MapViewState.cpp
ldicker83: True, but I kinda like the indirection a bit. Kinda like namespaces but without the namespace

PR #1362, opened 2023-05-16
Use on-demand check for `Mine::exhausted()`

### LINKED ISSUE #1333: Remove `Mine::checkExhausted` from `public` interface
This should really be a private implementation detail. In particular, it makes more sense to check this when ore is removed from a mine. Or maybe just check on demand in `exhausted()`, since the check should be cheaper with mine veins removed (PR #1330).

DanRStevens:
Rather than maintaining separate state to know if a mine is exhausted, just check if it has any resources left. This is now a simpler cheaper check since mine veins were removed. It also eliminates the potential for mismatch between resource counts and the exhausted flag, in case they are not both updated when they need to be.

Effectively, we want to use the actual resource counts as the single source of truth, rather than maintain separate state that must be kept in sync.

Closes #1333

### DIFF
--- a/OPHD/Mine.cpp
+++ b/OPHD/Mine.cpp
@@ -34,8 +34,8 @@ namespace
 	};
 
 
-	// [Exhausted, Active, 4x miningEnabled]
-	const std::bitset<6> DefaultFlags{"001111"};
+	// [Active, 4x miningEnabled]
+	const std::bitset<5> DefaultFlags{"01111"};
 }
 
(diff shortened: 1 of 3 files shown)


DanRStevens:
I maybe have some mild concern this one way cause backwards compatibility issues with any saved game files that may have exhausted mines. I'm not sure how the loading code would handle the old exhausted flag, now that the flag no longer exists.

PR #1368, opened 2023-05-17
Update tech catalog

ldicker83:
Adds additional information needed to render category icons for the research interface.

Also refactors some of the code to simplify implementation.

### DIFF
--- a/OPHD/Technology/TechnologyCatalog.cpp
+++ b/OPHD/Technology/TechnologyCatalog.cpp
@@ -172,7 +172,7 @@ namespace
 		for (auto category = &node; category; category = category->nextSiblingElement())
 		{
 			const auto attributes = NAS2D::attributesToDictionary(*category);
-			TechnologyCatalog::Category newCategory{std::stoi(attributes.get("icon_index")), attributes.get("name")};
+			TechnologyCatalog::Category newCategory{std::stoi(attributes.get("icon_index")), attributes.get("name"), {}};
 
 			for (const auto& cat : categories)
 			{
--- a/OPHD/Technology/TechnologyCatalog.cpp
+++ b/OPHD/Technology/TechnologyCatalog.cpp
@@ -207,6 +207,12 @@ TechnologyCatalog::TechnologyCatalog(const std::string& techFile)
 }
 
 
+const std::vector<std::string>& TechnologyCatalog::categories()
+{
+	return mCategorNames;
+}
+
+
 const Technology& TechnologyCatalog::technologyFromId(int id) const
 {
 	for (const auto& category : mCategories)
--- a/OPHD/Technology/TechnologyCatalog.cpp
+++ b/OPHD/Technology/TechnologyCatalog.cpp
@@ -172,7 +172,7 @@ namespace
 		for (auto category = &node; category; category = category->nextSiblingElement())
 		{
 			const auto attributes = NAS2D::attributesToDictionary(*category);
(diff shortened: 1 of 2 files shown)


OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -166,21 +166,25 @@ namespace
...
-	std::map<std::string, std::vector<Technology>> readCategories(NAS2D::Xml::XmlElement& node)
+	std::vector<TechnologyCatalog::Category> readCategories(NAS2D::Xml::XmlElement& node)
 	{
-		std::map<std::string, std::vector<Technology>> categories;
+		std::vector<TechnologyCatalog::Category> categories;
 		for (auto category = &node; category; category = category->nextSiblingElement())
 		{
 			const auto attributes = NAS2D::attributesToDictionary(*category);
-			const auto name = attributes.get("name");
+			TechnologyCatalog::Category newCategory{std::stoi(attributes.get("icon_index")), attributes.get("name"), {}};
```
DanRStevens, line 175: `attributes.get<int>("icon_index")`

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -166,21 +166,25 @@ namespace
...
+			for (const auto& cat : categories)
 			{
-				throw std::runtime_error("TechnologyReader: Category redefinition '" + name + "'" + nodeAtString(*category));
+				if (cat.name == newCategory.name)
+				{
+					throw std::runtime_error("TechnologyReader: Category redefinition '" + newCategory.name + "'" + nodeAtString(*category));
+				}
 			}
-			categories[name] = readTechnologiesInCategory(*category);
+			newCategory.technologies = readTechnologiesInCategory(*category);
```
DanRStevens, line 184: Potentially, you could delay construction of `newCategory` to this point, which avoids the need to initialize the `std::vector<Technology>` and then overwriting it with a new `vector`. Though the category `name` is needed for the redefinition check, so if that is read ahead of time you'd need to `std::move` it into the `Catagory` to avoid copy/re-init.

DanRStevens (approved):
Looks good.

The `icon_index` field part can be simplified a bit.


I still have doubts about splitting the technologies by category. That would seem to force a two stage lookup of techs, first by category, then by name or index. I think it would be simpler to work with if the category was an attribute of each tech, rather than a structural organization to the data.

OPHD/Technology/TechnologyCatalog.cpp
```diff
@@ -172,16 +172,17 @@ namespace
...
+				if (cat.name == name)
 				{
-					throw std::runtime_error("TechnologyReader: Category redefinition '" + newCategory.name + "'" + nodeAtString(*category));
+					throw std::runtime_error("TechnologyReader: Category redefinition '" + name + "'" + nodeAtString(*category));
 				}
 			}
-			newCategory.technologies = readTechnologiesInCategory(*category);
+			
+			const TechnologyCatalog::Category newCategory{attributes.get<int>("icon_index"), name, readTechnologiesInCategory(*category)};
 			categories.push_back(newCategory);
```
DanRStevens: Hmm, could even use `emplace_back` here, so the `Category` is constructed in-place in the `std::vector`:
```cpp
categories.emplace_back(attributes.get<int>("icon_index"), name, readTechnologiesInCategory(*category));
```

DanRStevens: Also, the use of `const` here takes away from the possibility of using `std::move` (or rather, you would just get a copy if you tried). Though `emplace_back` is better, since then there's no need to move/copy.

PR #1370, opened 2023-05-18
Improve `StructureManager` `const`-correctness

DanRStevens:
Related to: PR #1369 

Mark methods as `const`, so they can be called on const objects and references. Mark parameters as `const`, so they can accept `const` objects.

Pre-populate `std::map` of `StructureClass` so key lookup never fails for a valid `StructureClass`. This allows a `const` lookup method to be used, as we no longer end up modifying the `std::map` to insert blank `std::vector` values when a new key is used.

### DIFF
--- a/OPHD/MapObjects/Structure.cpp
+++ b/OPHD/MapObjects/Structure.cpp
@@ -107,6 +107,16 @@ std::string StructureName(StructureID id)
 }
 
 
+std::vector<Structure::StructureClass> allStructureClasses()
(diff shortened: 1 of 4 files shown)


DanRStevens:
I've wanted to improve the `const`-correctness on that one for ages, but in the past got hung up on the `std::map` not having keys pre-populated, so a simple conversion led to exceptions from `map::at` due to failed key lookups.

PR #1380, opened 2023-05-18
Remove `mine()` check from `walkGraph`

DanRStevens:
Remove the `mine()` check from `walkGraph` that prohibited connections through any `Tile` with a `Mine` on it.

The `MiningFacility` doesn't need a Command Center connection, but should be allowed to transmit a connection, if it's adjacent to another `Structure`.

Technically this change isn't specific to the `MiningFacility`, but rather to any structure on the same `Tile` as a `Mine`. Normally we would only expect a `MiningFacility` on such a `Tile`.

Related to:
- PR #1377


DanRStevens:
Just noticed that even with this change, a `MiningFacility` doesn't appear connected to the Command Center when the connection overlay is turned on.

Issue #1324, opened 2023-05-01
Research tree won't populate 


acf280:
Opening the mines/factories/satellite dialog box, the research tree just shows a blank slate. Honestly not sure if this is a bug with the build or with my installation, so I hope this is the right place to raise the issue - sorry in advance if not!

ldicker83:
This is expected - it's not complete at this point. I should have the panels at least indicate that they're not implemented, so I'll put that on the to-do list. Thanks!

Issue #27, opened 2018-04-28
Build Research UI


ldicker83:
Need a way to tell laboratories what to go do with themselves.

Related to #23 and #24.

ldicker83:
This is effectively going to be a two part issue -- one will require a dialog type window like the `FactoryProduction` Window, the other will be the fullscreen UI which hasn't yet been discussed or designed.

PR #1384, opened 2023-05-22
Add research report

ldicker83:
Starts to implement #27 and addresses #1324

This is mostly a visual implementation but the branch is getting really large so I want to end the branch here and implement the desired functionality into smaller change sets.

### DIFF
--- a/OPHD/GraphWalker.cpp
+++ b/OPHD/GraphWalker.cpp
@@ -71,22 +71,19 @@ static bool validConnection(Structure* src, Structure* dst, Direction direction)
 }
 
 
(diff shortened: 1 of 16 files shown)


ldicker83:
So this is an unintentionally colossal change set. There are a lot of ways I could/should have broken this down into smaller chunks that are more managable.

I chose to end this branch here with the visual parts of the research interface set up. There are a few changes that need to be done to other parts of the code to allow the research interface to be productive, and I wanted to spin those off into smaller PR's.

So yeah, that's the justification of creating the PR here. It's an unfinished feature but this will at least set the groundwork for going forward with finishing up the research interface.

PR #1395, opened 2023-05-29
Avoid `Rectangle` internal fields

DanRStevens:
Similar work in NAS2D:
- https://github.com/lairworks/nas2d-core/pull/1123

This is to support the transition in NAS2D:
- https://github.com/lairworks/nas2d-core/issues/704


DanRStevens:
I skipped modifying code that assigns to internal fields, rather than simply reading internal fields. The modification code will be easier to address after the internal changes, and is a much smaller bulk of code to deal with.

PR #1432, opened 2024-01-25
Feature/shift move

oscar139:
Shift + <Dir>  moves the map by 5 tiles.


OPHD/States/MapViewState.cpp
```diff
@@ -411,27 +411,34 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 		mReportsUiSignal();
 		return;
 	}
-
+	int repeat = 1;
+	if ((uint16_t)mod & (uint16_t)NAS2D::EventHandler::KeyModifier::Shift)
```
DanRStevens, line 415: Bit of a style issue here, we avoid C-style casts, and instead use C++ style casts. For example `static_cast<uint16_t>(mod)`.

However, in this particular case, I suspect an explicit cast may not be necessary. There are overloaded logical operators for this type, so it should be possible to do the logical `&` without needing to cast. See `EventHandler.h` and `EventHandler.cpp`.

OPHD/States/MapViewState.cpp
```diff
@@ -411,27 +411,34 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
...
+		repeat = 5;
 	switch(key)
 	{
+		
 		case NAS2D::EventHandler::KeyCode::KEY_w:
 		case NAS2D::EventHandler::KeyCode::KEY_UP:
-			mMapView->moveView(MapOffsetNorthWest);
+			for (int i = 0; i < repeat; i++)
+				mMapView->moveView(MapOffsetNorthWest);
```
DanRStevens: You can probably do this without the loop. If you check the definition, you'll find:
```cpp
constexpr auto MapOffsetNorthWest = MapOffset{DirectionNorthWest, 0};
```

Following definitions further, you'll find:
```cpp
constexpr auto DirectionNorthWest = NAS2D::Vector{-1, -1};
```

The `NAS2D::Vector` type has an overloaded `operator*`. Hence you could write something like:
```cpp
mMapView->moveView(MapOffset{DirectionNorthWest * repeat, 0});
```

Of course, at that point it may make sense to rename `repeat` to something else. Maybe `amount`, or `scale` or something similar.

oscar139:
I am getting tripped up on git with the sub-modules. My last commit relies on EventHandler to NAS2D::EventHandler::shift() being a static method. Should they be static or how else could I access them with out them being static?

DanRStevens:
I think you're right in that `shift` should be `static`. Actually, I'm wondering if that should be moved to `enum KeyModifier`.

I'm thinking I may need to go adjust that now.

OPHD/States/MapViewState.cpp
```diff
@@ -412,26 +412,33 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 		return;
 	}
 
+	int shiftMoveScalar = 1;
+	if (NAS2D::EventHandler::shift(mod))
+	{
+		shiftMoveScalar = 5;
+	}
```
DanRStevens, line 419: This can be shortened with a ternary operator, which also allows the value to be marked `const`:
```cpp
const int shiftMoveScalar = NAS2D::EventHandler::shift(mod) ? 5 : 1;
```

PR #1433, opened 2024-01-25
Changed connnect EventHandler to disconnect.

oscar139:
I get this exception when running the debugger. After starting a game and returning to the main menu and exiting. 

Also after starting a game returning to the main menu and starting a new game.

Exception thrown at 0x00007FF6CE99CD7B in OPHD.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF.

Investigating the call stack I made my way to this destructor 

 `ResearchReport::~ResearchReport()
{
	NAS2D::Utility<NAS2D::EventHandler>::get().mouseButtonDown().connect({this, &ResearchReport::onMouseDown});
}` 

I suspect .connect is a typo. Changing it to disconnect gets rid of the exception.

### DIFF
--- a/OPHD/UI/Reports/ResearchReport.cpp
+++ b/OPHD/UI/Reports/ResearchReport.cpp
@@ -89,7 +89,7 @@ ResearchReport::ResearchReport() :
 
 ResearchReport::~ResearchReport()
 {
-	NAS2D::Utility<NAS2D::EventHandler>::get().mouseButtonDown().connect({this, &ResearchReport::onMouseDown});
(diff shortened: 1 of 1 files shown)


DanRStevens (approved):
Huh, that does appear to be a bug.

Strangely, I don't get any crashes on my system. Maybe it's something specific to a Windows debug build that highlights the problem.

----

I see there are a lot of whitespace errors in the code. Also a lot of the new research code uses spaces instead of tabs for indentation. @ldicker83, you may want to adjust your editor settings. Either that, or we adjust the project indentation convention.

Thinking we might need a pass over the code to fix the odd whitespace. I usually try to keep whitespace changes separate from code changes so as to reduce the noise in commit diffs. Though in this case it's simple enough to not be too distracting.

PR #1434, opened 2024-01-26
Feature/worker scientist availability

oscar139:
I changed the population panel to indicate Number and Availability of Workers and Scientists and resized the panel accordingly. 
Workers: Number : Availability
Scientists:  Number : Availability

### DIFF
--- a/OPHD/UI/PopulationPanel.cpp
+++ b/OPHD/UI/PopulationPanel.cpp
@@ -6,6 +6,7 @@
 #include "../Constants/UiConstants.h"
 
 #include <libOPHD/Population/Population.h>
+#include <libOPHD/Population/PopulationPool.h>
 
 #include <NAS2D/Utility.h>
 #include <NAS2D/Resource/Font.h>
@@ -65,7 +66,7 @@ PopulationPanel::PopulationPanel() :
 		imageCache.load("ui/skin/window_bottom_right.png")
 	}
 {
-	constexpr int linesOfText = 14;
+	constexpr int linesOfText = 16;
 	constexpr int edgeBuffer = constants::Margin * 2;
 	const int windowHeight = mFontBold.height() + (mFont.height() * linesOfText) + (edgeBuffer * 2 /* Times two to account for both the edge and the divider line. */);
 
@@ -88,12 +89,15 @@ PopulationPanel::PopulationPanel() :
 	size({windowWidth, windowHeight});
 }
 
-
 void PopulationPanel::population(Population* pop)
 {
 	mPopulation = pop;
 }
 
+void PopulationPanel::populationPool(PopulationPool* popPool)
+{
+	mPopulationPool = popPool;
+}
 
 void PopulationPanel::addMoraleReason(const std::string& str, int val)
 {
@@ -112,7 +116,7 @@ void PopulationPanel::update()
 
 	auto position = NAS2D::Point{positionX() + constants::Margin, positionY() + constants::Margin};
 
-	// POPULATION
+	// POPULATION Statistics
 	renderer.drawText(mFontBold, constants::PopulationBreakdown, position);
 	const auto& population = mPopulation->getPopulations();
 	const std::array populationData
@@ -126,19 +130,40 @@ void PopulationPanel::update()
 
 	position.y += fontBoldHeight + constants::Margin;
 	const auto textOffset = NAS2D::Vector{IconSize + constants::Margin, (IconSize / 2) - (fontHeight / 2)};
+
 	for (const auto& [imageRect, personCount, personRole] : populationData)
 	{
 		renderer.drawSubImage(mIcons, position, imageRect);
 
-		const auto roleCount = std::to_string(personCount);
 		renderer.drawText(mFont, personRole + ": ", position + textOffset);
 
-		const NAS2D::Point<int> labelPosition = {positionX() + mPopulationPanelWidth - mFont.width(roleCount) - constants::Margin , position.y + textOffset.y};
-		renderer.drawText(mFont, roleCount, labelPosition);
+		const auto personCountString = std::to_string(personCount);
+		const NAS2D::Point<int> labelPosition = {positionX() + mPopulationPanelWidth - mFont.width(personCountString) - constants::Margin, position.y + textOffset.y};
+		renderer.drawText(mFont, personCountString, labelPosition);
+
 		position.y += IconSize + constants::Margin;
 	}
 
-	// DIVIDER LINE
+	// DIVIDER LINE Between population statistics and population availability statistics
+	renderer.drawLine(position, position + NAS2D::Vector<int>{mFont.width(constants::PopulationBreakdown) + constants::Margin, 0}, Color::DarkGray);
+	position.y += constants::Margin;
+
+	const std::array populationAvailablitiyStatistics{
+		std::tuple{"Available Workers: ", mPopulationPool->availableWorkers()},
+		std::tuple{"Available Scientists: ", mPopulationPool->availableScientists()},
+	};
+
+	for (const auto& [statisticLabel, personCount] : populationAvailablitiyStatistics)
+	{
+		const auto personCountString = std::to_string(personCount);
+		const Color statusColor = personCount <= 0 ? Color::Red : Color::White;
+		renderer.drawText(mFont, statisticLabel, position, Color::White);
+		const NAS2D::Point<int> labelPosition = {positionX() + mPopulationPanelWidth - mFont.width(personCountString) - constants::Margin, position.y};
+		renderer.drawText(mFont, personCountString, labelPosition, statusColor);
+		position.y += fontBoldHeight + constants::Margin;
+	}
+
+	// DIVIDER LINE Between population statistics and morale statistics
 	position = NAS2D::Point{positionX() + mPopulationPanelWidth, positionY() + constants::Margin};
 	renderer.drawLine(position, position + NAS2D::Vector<int>{0, rect().size.y - 10}, Color::DarkGray);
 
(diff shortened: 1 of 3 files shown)


OPHD/UI/PopulationPanel.h
```diff
@@ -4,6 +4,8 @@
 
 #include <NAS2D/Renderer/RectangleSkin.h>
 
+#include <libOPHD/Population/PopulationPool.h>
```
DanRStevens, line 7: Consider using a forward declare for `PopulationPool` rather than a `#include`. Similar to `Population`, we don't need a complete type in this file. Using a forward declare here, and moving the `#include` to the `.cpp` will decrease transitive includes, which helps lower compile times.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -81,7 +81,7 @@ PopulationPanel::PopulationPanel() :
 
 	largestStringLength += edgeBuffer + mFont.width("999999");
 
-	mPopulationPanelWidth = mFontBold.width(constants::PopulationBreakdown) + edgeBuffer;
+	mPopulationPanelWidth = mFontBold.width(constants::PopulationBreakdown) + edgeBuffer + mFont.width(" : 999");//mFont.width(" : 999") Accounts for the space needed for worker/scientist avialibility.
 	const int windowWidth = largestStringLength + edgeBuffer + mPopulationPanelWidth;
```
DanRStevens, line 85: This might need some adjustment. Although the values are currently aligned with the end of the title string, it's a bit odd to be adding the length of values to the length of the title. There's an implied assumption about that alignment, which could potentially change in the future if the display is changed.

It would seem more intuitive to be adjusting the `largestStringLength` field instead. Though I may need to look at this a bit more closely to know if that's appropriate.

Perhaps a deeper question is if the colon separated double value will be clear. Maybe we need a new line with descriptive text to distinguish between total people and available people.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -132,9 +138,13 @@ void PopulationPanel::update()
 
 		const auto roleCount = std::to_string(personCount);
 		renderer.drawText(mFont, personRole + ": ", position + textOffset);
+		std::string roleCountString = roleCount;
+		if (personRole == "Workers")
+			roleCountString = roleCount + " : " + std::to_string(mPopulationPool->availableWorkers());//Add population of workers Available
```
DanRStevens: I'm somewhat bothered by the use of an `if` statement like this inside of a loop. Maybe we should look at adjusting the loop data, rather than adding `if` statements to the loop body. Maybe we add the extra info to the fields in the defined array, or add more rows to the array. Or maybe we add a new section to that part of the UI.

OPHD/UI/PopulationPanel.cpp
oscar139: I had to do a double take when you pointed this out. Why did I add it to mPopulationPanelWidth instead of largestStringLength? 

largestStringLength controls the size of the moral breakdown side of the population panel making room for the largest moral issue string and adding to it doesn't help make space for the : and worker usage availability on the population breakdown side. Adding to largestStringLength just makes the moral breakdown side larger. 

`const int windowWidth = largestStringLength + edgeBuffer + mPopulationPanelWidth;`

I didn't know why windowWidth isn't something like mMoralPanelWidth + edgeBuffer + mPopulationPanelWidth

For the deeper question I suppose we need to decide on what we want to indicate what the other value is. I just wanted some other number that i could intuit what the number of remaining workers and scientists instead of running out unexpectedly.

OPHD/UI/PopulationPanel.cpp
oscar139: Commit cb642f8 moved the appended portion of the string to the array of tuples eliminating the if statements in the loop.

OPHD/UI/PopulationPanel.cpp
DanRStevens: Ahh, it seems I didn't examine the code closely enough. It seems you are right that's for the other side of the panel.

You might guess that I didn't design the UI, so I sometimes miss some of the fine details like that. Most of the UI code I've touched has been for low level refactoring, without much consideration for high level design.

I'm open to suggestions to rename things for better clarity.

Looking at the game UI, I'm thinking maybe we should add another section below the "Population Breakdown" section for "Available Population" (or some similar description), which shows the available number of workers and scientists. The other population components don't really contribute work, so they wouldn't need to be listed there. That might provide more clarity than a ":" separate set of two values.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -112,33 +116,53 @@ void PopulationPanel::update()
...
 		renderer.drawSubImage(mIcons, position, imageRect);
 
-		const auto roleCount = std::to_string(personCount);
 		renderer.drawText(mFont, personRole + ": ", position + textOffset);
 
-		const NAS2D::Point<int> labelPosition = {positionX() + mPopulationPanelWidth - mFont.width(roleCount) - constants::Margin , position.y + textOffset.y};
-		renderer.drawText(mFont, roleCount, labelPosition);
+		const NAS2D::Point<int> labelPosition = {positionX() + mPopulationPanelWidth - mFont.width(std::to_string(personCount)) - constants::Margin, position.y + textOffset.y};
+		renderer.drawText(mFont, std::to_string(personCount), labelPosition);
+
```
DanRStevens, line 143: This makes multiple calls to `std::to_string` with the same argument. We should probably cache the result to avoid having to call it twice. Notably, the `std::to_string` function will likely cause memory allocation (assuming no short string optimization), and we should probably try to avoid multiple short allocations when reasonable to avoid them.

It shouldn't cost anything to extend the lifetime of a temporary to the body of the loop.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -112,33 +116,53 @@ void PopulationPanel::update()
...
+	const std::array populationAvailablitiyStatistics{
+		std::tuple{"Available Workers: ", mPopulationPool->availableWorkers()},
+		std::tuple{"Available Scientists: ", mPopulationPool->availableScientists()},
+	};
+
+	for (const auto& [statisticLable, personCount] : populationAvailablitiyStatistics)
+	{
+		const Color statusColor = personCount <= 0 ? Color::Red : Color::White;
+		renderer.drawText(mFont, statisticLable, position, Color::White);
```
DanRStevens, line 159: Spelling of `statisticLable`

DanRStevens (commented):
That looks pretty good.

image

----

Most of my feedback is pretty minor on this one.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -65,7 +66,7 @@ PopulationPanel::PopulationPanel() :
 		imageCache.load("ui/skin/window_bottom_right.png")
 	}
 {
-	constexpr int linesOfText = 14;
+	constexpr int linesOfText = 16;
```
oscar139, line 69: PopPanel

I do not know if this helps. But I made a quick visualizer for the 16 lines of text in red.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -126,19 +130,41 @@ void PopulationPanel::update()
 
 	position.y += fontBoldHeight + constants::Margin;
 	const auto textOffset = NAS2D::Vector{IconSize + constants::Margin, (IconSize / 2) - (fontHeight / 2)};
+	const int populationStatisticsWidth = mFont.width(constants::PopulationBreakdown) + constants::Margin;
```
DanRStevens, line 133: Very minor point here, though I was thinking this line doesn't really relate to the surrounding code. Printing of the mentioned strings happens before the block of code above, and use of this value happens below the block of code below. It seems like this line should be moved to one of those two places, probably after the block below, where it's used.

Or potentially, the value could be inlined, since it's only used once. Though that might make the line where it's used a bit long.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -126,19 +130,41 @@ void PopulationPanel::update()
...
 
-		const auto roleCount = std::to_string(personCount);
 		renderer.drawText(mFont, personRole + ": ", position + textOffset);
 
-		const NAS2D::Point<int> labelPosition = {positionX() + mPopulationPanelWidth - mFont.width(roleCount) - constants::Margin , position.y + textOffset.y};
-		renderer.drawText(mFont, roleCount, labelPosition);
+		const auto personCountString = std::to_string(personCount);
+		const NAS2D::Point<int> labelPosition = {positionX() + mPopulationPanelWidth - mFont.width(personCountString) - constants::Margin, position.y + textOffset.y};
+		renderer.drawText(mFont, personCountString, labelPosition);
+
```
DanRStevens, line 144: Very minor point here, though I was just thinking this refactoring is a bit independent of the new feature you're adding. That could make it a good candidate to do this change as a separate commit, before the changes to add your new feature.

No changes required here. Just noting it for future reference as a way of potentially communicating a change is a cleanup to old code versus part of implementing a new feature.

OPHD/UI/PopulationPanel.h
```diff
@@ -23,6 +24,7 @@ class PopulationPanel : public Control
 	PopulationPanel();
 
 	void population(Population* pop);
+	void populationPool(PopulationPool* popPool);
```
DanRStevens, line 27: Largely unrelated to this PR, though I noticed it during review and wanted to note this construct as an area for potential future work. I opened Issue https://github.com/OutpostUniverse/OPHD/issues/1442 to elaborate on it.

OPHD/UI/PopulationPanel.cpp
oscar139: Inlining made it no longer than the surrounding code, its intent appears obvious. So Ill inline it.

DanRStevens (approved):

PR #1435, opened 2024-01-26
Feature/expand structure icon tool tip

oscar139:
Expanded the Structure icon tool tip to now include the cost to build it and the workers and scientist needed to operate it.

I don't know if this is a too ad hoc solution but I made a temporary Structure Object to gain access to the population requirements.
`const auto& tempStructure = Structure::Structure(Structure::StructureClass::Undefined, (StructureID)mIconItemList[mHighlightIndex].meta);`
`const auto& popNeeds = tempStructure.populationRequirements();`

I would also like to direct your attention to the HEADS UP comment. Again this feels ad hoc.

### DIFF
--- a/OPHD/UI/IconGrid.cpp
+++ b/OPHD/UI/IconGrid.cpp
@@ -151,6 +151,12 @@ void IconGrid::addItem(const Item& item)
 }
 
 
+void IconGrid::addResourceCosts(std::string name, std::vector<std::tuple<std::string, int>>& resourceCost)
+{
+	mResourceCosts[name] = resourceCost;
+}
+
+
 /**
  * Set item availability
  */
--- a/OPHD/UI/IconGrid.cpp
+++ b/OPHD/UI/IconGrid.cpp
@@ -378,10 +378,40 @@ void IconGrid::update()
 		if (mShowTooltip)
 		{
 			const auto& highlightedName = mIconItemList[mHighlightIndex].name;
-			const auto tooltipRect = NAS2D::Rectangle<int>{{position.x, position.y - 15}, {mFont.width(highlightedName) + 4, mFont.height()}};
-			renderer.drawBoxFilled(tooltipRect, NAS2D::Color{245, 245, 245});
+
+			int toolTipBoxWidth = mFont.width(highlightedName);
+			int toolTipBoxHeight = mFont.height();
+
+			auto resourceCostIterator = mResourceCosts.find(mIconItemList[mHighlightIndex].name);
+			if (resourceCostIterator != mResourceCosts.end())
+			{
+				for(auto& resourceCost: resourceCostIterator->second)
+				{
+					toolTipBoxWidth = std::max(toolTipBoxWidth, mFont.width(std::get<0>(resourceCost) + ": " + std::to_string(std::get<1>(resourceCost))));
+				}
+				toolTipBoxHeight = mFont.height() * static_cast<int>(resourceCostIterator->second.size() + 1);
+			}
+
+			const auto tooltipRect = NAS2D::Rectangle<int>{{position.x, position.y - toolTipBoxHeight}, {toolTipBoxWidth + 4, toolTipBoxHeight}};
+			renderer.drawBoxFilled(tooltipRect, NAS2D::Color{225, 225, 225});
+
+			const auto toolTipHighlightRect = NAS2D::Rectangle<int>{{position.x, position.y - mFont.height()}, {toolTipBoxWidth + 4, mFont.height()}};
+			renderer.drawBoxFilled(toolTipHighlightRect, NAS2D::Color{245, 245, 245});
 			renderer.drawBox(tooltipRect, NAS2D::Color{175, 175, 175});
 			renderer.drawText(mFont, highlightedName, position + NAS2D::Vector{2, -15}, NAS2D::Color::Black);
+
+
+			if (resourceCostIterator != mResourceCosts.end())
+			{
+				const int numberOfLines = static_cast<int>(resourceCostIterator->second.size());
+				const int startOffset = static_cast<int>(-mFont.height() * numberOfLines);
+				int currentLine = 0;
+				for(auto& resourceCost: resourceCostIterator->second)
+				{
+					renderer.drawText(mFont, std::get<0>(resourceCost) + ": " + std::to_string(std::get<1>(resourceCost)), position + NAS2D::Vector{2, startOffset + (-15 + mFont.height() * currentLine)}, NAS2D::Color::Black);
+					currentLine++;
+				}
+			}
 		}
 	}
 }
--- a/OPHD/UI/IconGrid.h
+++ b/OPHD/UI/IconGrid.h
@@ -48,6 +48,7 @@ public:
 	bool empty() const { return mIconItemList.empty(); }
 
 	void addItem(const Item&);
+	void addResourceCosts(std::string name, std::vector<std::tuple<std::string, int>>& resourceCost);
 
 	void removeItem(const std::string& item);
 	bool itemExists(const std::string& item);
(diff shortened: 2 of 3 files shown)


oscar139:
I was getting error Sprite::play called on undefined action: construction (structures/air_shaft.sprite). It was being caused by 

`const auto& tempStructure = Structure::Structure(Structure::StructureClass::Undefined, (StructureID)mIconItemList[mHighlightIndex].meta);`

It would happen whenever you placed your cursor over the Robominer in the FactoryProduction Panel. Which mIconItemList[mHighlightIndex].name produces "RoboMiner" vs "Robominer"  defeating 

`if (toolTipStrings[0] != "Robodigger" && toolTipStrings[0] != "Robodozer" && toolTipStrings[0] != "Robominer")`

OPHD/UI/IconGrid.h
```diff
@@ -15,6 +15,10 @@
 
 #include <algorithm>
 
+#include "../MapObjects/Structure.h"
+#include "../StructureCatalogue.h"
+#include "../Common.h"
+#include "../StorableResources.h"
```
DanRStevens, line 21: Seeing as how these are newly added and not referenced in the header file, it would be better to move new includes to the implementation file instead. That would avoid increasing transitive includes, which affect compile times.

OPHD/UI/IconGrid.cpp
```diff
@@ -371,11 +371,61 @@ void IconGrid::update()
...
-			const auto& highlightedName = mIconItemList[mHighlightIndex].name;
-			const auto tooltipRect = NAS2D::Rectangle<int>{{position.x, position.y - 15}, {mFont.width(highlightedName) + 4, mFont.height()}};
-			renderer.drawBoxFilled(tooltipRect, NAS2D::Color{245, 245, 245});
-			renderer.drawBox(tooltipRect, NAS2D::Color{175, 175, 175});
-			renderer.drawText(mFont, highlightedName, position + NAS2D::Vector{2, -15}, NAS2D::Color::Black);
+			std::vector<std::string> toolTipStrings;
+			toolTipStrings.push_back(mIconItemList[mHighlightIndex].name);
+			/* HEADS UP Feels like there is a better way. Unless more robots are to be added it works.
+			* Without this name check accessing the structure resource and population requirements causes an error because the robots are not structures*/
+			if (toolTipStrings[0] != "Robodigger" && toolTipStrings[0] != "Robodozer" && toolTipStrings[0] != "Robominer" && this->mIconSheet.name() != "ui/factory.png")
```
DanRStevens, line 378: This seems a bit too ad hoc. I would avoid using display strings, which are subject to change, for any kind of logical condition.

There's probably a better way to handle this case, though I may need to circle back to this. Perhaps we need a new field on each data element for use in the tooltip message. It would probably be easier to get this data when setting the grid icon then when trying to display it.

OPHD/UI/IconGrid.cpp
```diff
@@ -371,11 +371,61 @@ void IconGrid::update()
...
+			{
+				const auto& tempStructure = Structure::Structure(Structure::StructureClass::Undefined, (StructureID)mIconItemList[mHighlightIndex].meta);
+				const auto& popNeeds = tempStructure.populationRequirements();
+				const auto& cost = StructureCatalogue::costToBuild((StructureID)mIconItemList[mHighlightIndex].meta);
```
DanRStevens, line 382: Similar to the last comment, this might be easier to lookup when setting the grid data, rather than during display.

DanRStevens (commented):
Yeah, this one seems a bit too ad hoc in its current form.

OPHD/UI/IconGrid.h
oscar139: To clear things up about this comment and the similar one on the other PR. Move needed includes (in this case it may not be needed) to the .cpp.

OPHD/UI/IconGrid.cpp
```diff
@@ -378,10 +378,40 @@ void IconGrid::update()
...
+			int toolTipBoxWidth = mFont.width(highlightedName);
+			int toolTipBoxHeight = mFont.height();
+
+			auto resourceCostIterator = mResourceCosts.find(mIconItemList[mHighlightIndex].name);
+			if (resourceCostIterator != mResourceCosts.end())
+			{
+				for(auto& resourceCost: resourceCostIterator->second)
+				{
+					toolTipBoxWidth = std::max(toolTipBoxWidth, mFont.width(std::get<0>(resourceCost) + ": " + std::to_string(std::get<1>(resourceCost))));
+				}
```
DanRStevens, line 391: You can potentially use [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding) to unpack the tuple in the ranged-for, without having to use `std::get<index>`.

```cpp
for(const auto& [resourceName, resourceCost] : resourceCostIterator->second)
{
	toolTipBoxWidth = std::max(toolTipBoxWidth, mFont.width(resourceName + ": " + std::to_string(resourceCost)));
}
```

OPHD/UI/IconGrid.cpp
```diff
@@ -378,10 +378,40 @@ void IconGrid::update()
...
+			if (resourceCostIterator != mResourceCosts.end())
+			{
+				const int numberOfLines = static_cast<int>(resourceCostIterator->second.size());
+				const int startOffset = static_cast<int>(-mFont.height() * numberOfLines);
+				int currentLine = 0;
+				for(auto& resourceCost: resourceCostIterator->second)
+				{
+					renderer.drawText(mFont, std::get<0>(resourceCost) + ": " + std::to_string(std::get<1>(resourceCost)), position + NAS2D::Vector{2, startOffset + (-15 + mFont.height() * currentLine)}, NAS2D::Color::Black);
+					currentLine++;
+				}
```
DanRStevens, line 413: Same note about structured bindings here.

Additionally, the `std::string` here is being created again, even though it matches the previously generated value. That doubles up on both the dynamic memory allocation, and the string concatenation. Granted it would be awkward to somehow re-use the previously generated value with the current structure.

More thoughts on this below.

OPHD/UI/IconGrid.h
```diff
@@ -108,6 +109,7 @@ class IconGrid : public Control
 	NAS2D::Vector<int> mGridSize; /**< Dimensions of the grid that can be contained in the IconGrid with the current Icon Size and Icon Margin. */
 
 	IconItemList mIconItemList; /**< List of items. */
+	std::map<std::string, std::vector<std::tuple<std::string, int>>> mResourceCosts; /**< List of resources required to build the item. */
```
DanRStevens, line 112: This is an interesting and slightly surprising data structure to accomplish the given task. I can see how this would work, and be fairly general. It also seems a bit heavy in some ways.

The `std::vector` portion contains multiple `std::string`, and those's strings are likely to be identical for every `std::vector` instance, yet there is no sharing. Each `std::string` is a separate copy taking up additional memory.

On the other hand, by storing the `std::string` and the `int` separately means we need to dynamically generate concatenated strings during both sizing and display. If the full string was generated up front, that would eliminate the need for dynamic creation during display. Additionally, the string sizes could potentially be calculated up front as well, so they wouldn't need to be recalculated every frame.

As a further minor note, although there would be just as many strings as before, the combination of the resource name and cost are more likely to be unique, so in that sense there would be less duplication. There's no savings here, rather, there's more justification for the current consumption.

One downside to generating the strings early, is if you wanted to format the resource name differently from the resource cost, it would then be harder to do. Perhaps the names and cost would be in a different color, or font size, or have different alignment.

----

As for why I found the solution surprising, I think I expected the entries of `IconItemList` to contain the extra data, rather than a separate lookup structure. That would simplify the looping, since all data would be stored together in the same structure, so there would be no need for an additional lookup, and no need for error checking against a failed lookup.

Perhaps the other surprising bit was the custom data structure using only standard library types. I was perhaps expecting `StorableResources` and `PopulationRequirements` as the types for additional fields. This idea of course being more natural if they are added to `IconItemList`, rather than the `std::map` structure.

----

Even if the data is stored as separate components (and dynamically generated during display, rather than storing a pre-generated `std::string`), the sizing can still be done up front. The needed size would be known as soon as the `IconItemList` field is set/updated. Here I'm kind of assuming it would be set up front, and not really modified after. I haven't really checked the source for that point. It's probably not actually written with that in mind.

Actually, I'm wondering about the design having an `addItem` function. Maybe there should only be the option to construct `IconGrid` with all items at once. It's harder to optimize these things when the data is mutable, and in many cases, mutability isn't actually needed. The game knows early on what all the options and costs are, with changes not being common.

DanRStevens (commented):
The solution seems reasonable. I had a few thoughts that might be worth considering.

Issue #1442, opened 2024-02-07
`PopulationPanel` should take required arguments as constructor parameters


DanRStevens:
In regards to the following code:
```cpp
	PopulationPanel();

	void population(Population* pop);
	void populationPool(PopulationPool* popPool);
```

When there are required parameters to an object, such as `Population*` and `PopulationPool*`, they should usually be provided as constructor parameters, rather than a separate setter. In particular, much of the code in the class assumes these values are never `nullptr`, which is their default values. It would make sense to remove their default values and force them to be initialized during object construction.

Actually, if the two fields are being initialized in the constructor, it may also make sense to convert them to reference syntax. A reference would need to be initialized by a constructor init list, so by moving initialization there, it makes it possible to use references. In addition to nicer access syntax, references are generally considered non-nullable, so that would also be consistent with assumptions made in the code.

Seems like the values should also be marked as `const`, since this should only be for reading and display purposes. Though I wonder if perhaps they need to make use of a member function that hasn't been marked `const`. Potentially the change isn't as simple as I think.

----

Target design:
```cpp
PopulationPanel(const Population& pop, const PopulationPool& popPool);
```

DanRStevens:
It's been a few weeks, so I thought maybe I'd lookup some of the comments on Discord and summarize them here for reference.

----

PR #1444 dealt with marking the needed support functions in `PopulationPool` as `const`.

----

Remaining steps (suggested commits):

- Add optional constructor parameters, with default values of `nullptr`. This can be done while keeping a single constructor. Keep pointer syntax for now.
  - `PopulationPanel(Population* pop = nullptr, PopulationPool* popPool=nullptr);`
- Update the calling code to use the new constructor arguments.
- Remove the default `nullptr` values, and the setter methods (they shouldn't be used at this point).
  - `PopulationPanel(Population* pop, PopulationPool* popPool);`
- Convert to reference syntax, including the member field, the constructor parameters, and update the constructor call(s).
   - `PopulationPanel(Population& pop, PopulationPool& popPool);`
- Mark the member fields and constructor parameters as `const` (optionally this could possibly be done first)
   - Example: `Population& mPopulation` -> `const Population& mPopulation`

ldicker83:
In agreement here, that's a pattern I've recently been starting to change by deleting default constructors unless truly needed and opting for c'tor's with required parameters instead of setters.

DanRStevens:
Looks like changes that fix this issue have already been merged.

PR #1443, opened 2024-02-08
Refactor/population panel constructor

oscar139:
Initial attempt to address issue #1442.

### DIFF
--- a/OPHD/States/MapViewStateUi.cpp
+++ b/OPHD/States/MapViewStateUi.cpp
@@ -75,8 +75,6 @@ void MapViewState::initUi()
 	mFileIoDialog.hide();
 
 	mPopulationPanel.position({675, constants::ResourceIconSize + 4 + constants::MarginTight});
-	mPopulationPanel.population(&mPopulation);
-	mPopulationPanel.populationPool(&mPopulationPool);
 
 	mResourceBreakdownPanel.position({0, 22});
 	mResourceBreakdownPanel.playerResources(&mResourcesCount);
--- a/OPHD/UI/PopulationPanel.cpp
+++ b/OPHD/UI/PopulationPanel.cpp
@@ -49,7 +49,7 @@ static const std::array moraleStringColor
 };
 
 
-PopulationPanel::PopulationPanel(Population* pop, PopulationPool* popPool) :
+PopulationPanel::PopulationPanel(const Population* pop, const PopulationPool* popPool) :
 	mFont{fontCache.load(constants::FONT_PRIMARY, constants::FontPrimaryNormal)},
 	mFontBold{fontCache.load(constants::FONT_PRIMARY_BOLD, constants::FontPrimaryNormal)},
 	mIcons{imageCache.load("ui/icons.png")},
--- a/OPHD/UI/PopulationPanel.cpp
+++ b/OPHD/UI/PopulationPanel.cpp
@@ -49,7 +49,7 @@ static const std::array moraleStringColor
 };
 
 
-PopulationPanel::PopulationPanel(const Population* pop, const PopulationPool* popPool) :
+PopulationPanel::PopulationPanel(const Population& pop, const PopulationPool& popPool) :
 	mFont{fontCache.load(constants::FONT_PRIMARY, constants::FontPrimaryNormal)},
 	mFontBold{fontCache.load(constants::FONT_PRIMARY_BOLD, constants::FontPrimaryNormal)},
 	mIcons{imageCache.load("ui/icons.png")},
@@ -110,7 +110,7 @@ void PopulationPanel::update()
 
 	// POPULATION Statistics
 	renderer.drawText(mFontBold, constants::PopulationBreakdown, position);
-	const auto& population = mPopulation->getPopulations();
+	const auto& population = mPopulation.getPopulations();
 	const std::array populationData
 	{
 		std::tuple{NAS2D::Rectangle<int>{{0, 96}, {IconSize, IconSize}}, population.child, std::string("Children")},
@@ -141,8 +141,8 @@ void PopulationPanel::update()
 	position.y += constants::Margin;
 
 	const std::array populationAvailablitiyStatistics{
-		std::tuple{"Available Workers: ", mPopulationPool->availableWorkers()},
-		std::tuple{"Available Scientists: ", mPopulationPool->availableScientists()},
+		std::tuple{"Available Workers: ", mPopulationPool.availableWorkers()},
+		std::tuple{"Available Scientists: ", mPopulationPool.availableScientists()},
 	};
 
 	for (const auto& [statisticLabel, personCount] : populationAvailablitiyStatistics)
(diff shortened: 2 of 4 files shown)


libOPHD/Population/PopulationPool.h
```diff
@@ -15,16 +15,16 @@ class PopulationPool
...
 	bool usePopulation(PopulationRequirements populationRequirements);
 
 	void clear();
 
 	int scientistsAsWorkers();
-	int scientistsEmployed();
-	int workersEmployed();
+	const int scientistsEmployed() const;
+	const int workersEmployed() const;
 	int populationEmployed();
```
oscar139, line 28: scientistsAsWorkers, populationEmployed, and size have a similar behavior as the other functions. Does it make sense to change them too. I made the changes locally and it did not appear to create any issues.

DanRStevens: It's often a good idea to marking functions as `const` as a sort of default, and only remove it if the function really needs to mutate state.

A quick note on semantics: I've seen cases where you could technically mark a function as `const`, and it would compile, but that was only because there was a level of indirection to the data being modified. The internal level of indirection is an implementation detail, and the nature of the function in that case really was to mutate state. In such cases, I think it makes the most sense to not declare a method as `const`. In particular, if you were to refactor the internal implementation, you might be required to remove the `const`, which could potentially break calling code if it assumed a method was callable on a `const` object.

DanRStevens: It doesn't really make sense to declare the return values as `const`. They are being copied anyway.

Besides, the caller should decide if the result they're storing should be `const` or not.

OPHD/UI/PopulationPanel.h
```diff
@@ -55,4 +55,6 @@ class PopulationPanel : public Control
 	int mResidentialCapacity{0};
 	int mCrimeRate{0};
 	int mPopulationPanelWidth{0};
+
+	void init();
```
DanRStevens, line 60: I generally recommend against `init` methods, as what they contain should really be in a constructor.

These did make sense in the past, when you wanted to share code between constructors, which I see is the direction you were going here. Since C++11 this approach is no longer necessary, since "Delegating Constructors" were introduced to the language. When you want to re-use constructors, you can call one from the other using the member initialization list. In that case, only that one single "member" can be initialized.

Some details are found about half way down this page:
https://en.cppreference.com/w/cpp/language/constructor

One approach might be for the default constructor to call a constructor taking parameters, and passing default values. However, in that case, it might make sense to set default values for the parameters, making them optional, and allowing that constructor to also function as the default constructor.

In this particular case, that's possible if you're still using pointer syntax, but not reference syntax. You can't quite switch to reference syntax until after the setters are removed, since a reference is basically a `const` pointer (to data that may or may not be `const`).

DanRStevens (commented):
Looks like there are merge conflicts that need to be resolved. Probably from when PR #1434 was merged.

OPHD/UI/PopulationPanel.h
ldicker83: Generally agreed here -- Part of the style that I've been using lately is "Only 5 Lines of Code" per function (including c'tor) and I tend to extract some of that code into stand alone functions that are wrapped in an anonymous namespace.

DanRStevens: My general rule of thumb is to try and keep functions within one page of code, so you never need to scroll to see the full context. There's a lot less you need to remember when you can see it all at once.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -66,6 +66,8 @@ PopulationPanel::PopulationPanel(Population* pop, PopulationPool* popPool) :
 		imageCache.load("ui/skin/window_bottom_right.png")
 	}
 {
+	mPopulation = pop;
+	mPopulationPool = popPool;
```
DanRStevens: These two lines should probably have been part of the previous commit. It might also have been helpful to use the constructor init-list, rather than the constructor body. That would be important for the switch to reference syntax, since the constructor body is too late to initialize references (or `const` values).

Note that a pointer to const data is not itself const. The pointer can still be modified, even if the data is const. A reference is like a const pointer, in that the pointer itself can not be modified after initialization.

OPHD/States/MapViewStateUi.cpp
```diff
@@ -75,8 +75,6 @@ void MapViewState::initUi()
 	mFileIoDialog.hide();
 
 	mPopulationPanel.position({675, constants::ResourceIconSize + 4 + constants::MarginTight});
-	mPopulationPanel.population(&mPopulation);
-	mPopulationPanel.populationPool(&mPopulationPool);
```
DanRStevens, line 79: This could potentially have been grouped into the previous commit. Once initialization is done using constructor parameters, there is no longer a need to call the setter methods. Doing so initializes the values twice.

OPHD/UI/PopulationPanel.cpp
```diff
@@ -66,8 +66,8 @@ PopulationPanel::PopulationPanel(const Population* pop, const PopulationPool* po
 		imageCache.load("ui/skin/window_bottom_right.png")
 	}
 {
-	mPopulation = pop;
-	mPopulationPool = popPool;
+	mPopulation = &pop;
+	mPopulationPool = &popPool;
```
DanRStevens, line 70: Ahh, I see internally pointer syntax is still being used. For a moment I was wondering why this didn't result in a compiler error. If reference syntax had been used internally, then this initialization would have been too late. The constructor init list is needed to initialize references. Once the constructor body is entered, all component objects are already initialized.

Or was that why pointer syntax was still being used internally?

DanRStevens (approved):
Very nicely presented. Very easy to read through and understand.

For the second commit, there were two changes grouped together, which could possibly have been separate commits.
- Changing from calling setter methods to using new constructor parameters
- Marking the fields and parameters as `const`

Though that's a fairly minor observation.

I've pushed the branch to the main repo so the CI build will run. I'll merge this once the builds finish, assuming no problems.

----

Edit: Huh, I see CI build failures, though the logs make it look like it actually passed. I may need to investigate a bit further to see what happened.

DanRStevens:
Looks like constructor init list items don't match the order of the fields in the class. We have a warning flag enabled that checks for that.

Member fields are always initialized in their declaration order, not the order they appear in the init list. Hence putting them out of order in the init list can be misleading. One of the reasons for that is a class can have many constructors, which means it can have many init list orderings. The class definition only appears once though, so the order specified there is what is used.

Initialization order is important in the case of exceptions, since the exception handler must destruct all objects in the reverse order they were constructed (and only for the objects that completed construction). If a constructor fails with an exception part way through, some set of member fields may need to be destructed, depending on how far into construction the class got.

I'm not entirely sure why all constructors need to initialize objects in the same order, though if all constructors initialize objects in the same order, than they can all share the same exception handling code to destruct objects in the same order.

DanRStevens (approved):

PR #1445, opened 2024-03-12
Fix PgUp/PgDown not changing structure palette

### LINKED ISSUE #1364: PageUp/Down key from underground to surface doesn't change structure picker palette

alynsarcana:
The number keys called a local function that handled changing the structure picker. This change adds a 2nd parameter (defaulting to null) to that local function, allowing the event handlers for PageUp/Down to call the same function, rather than call directly to the mMapView object. If the 2nd parameter have a value, then the 1st parameter is ignored and the 2nd is used to change the mMapView z level.

----

Edit:
Closes #1364

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -458,11 +458,11 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_PAGEUP:
-			changeViewDepth(0, MapOffsetUp);
+			moveView( MapOffsetUp);
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_PAGEDOWN:
-			changeViewDepth(0, MapOffsetDown);
+			moveView(MapOffsetDown);
 			break;
 
 
@@ -723,29 +723,44 @@ void MapViewState::onSystemMenu()
 	resetUi();
 }
 
-
 /**
- * Changes the current view depth. If a value is given for the 2nd parameter, the first will be ignored.
- */
-void MapViewState::changeViewDepth(int depth, std::optional<MapOffset> direction)
-{
+* Handle side effects of changing depth view
+*/
+void MapViewState::onChangeDepth(int oldDepth, int newDepth) {
 	if (mBtnTogglePoliceOverlay.isPressed())
 	{
-		changePoliceOverlayDepth(mMapView->currentDepth(), depth);
-	}
-
-	if (direction.has_value())
-	{
-		mMapView->moveView(direction.value());
-	}
-	else
-	{
-		mMapView->currentDepth(depth);
+		changePoliceOverlayDepth(oldDepth, newDepth);
 	}
 
 	if (mInsertMode != InsertMode::Robot) { clearMode(); }
 
 	populateStructureMenu();
+}
+
+void MapViewState::moveView(MapOffset offset) {
+	int oldZLevel = mMapView->currentDepth();
+	
+	mMapView->moveView(offset);
+
+	int newZLevel = mMapView->currentDepth();
+	if (oldZLevel != newZLevel) {
+		onChangeDepth(oldZLevel, newZLevel);
+	}
+}
+
+/**
+ * Changes the current view depth.
+ */
+void MapViewState::changeViewDepth(int depth)
+{
+	int oldZLevel = mMapView->currentDepth();
+
+	mMapView->currentDepth(depth);
+
+	int newZLevel = mMapView->currentDepth();
+	if (oldZLevel != newZLevel) {
+		onChangeDepth(oldZLevel, newZLevel);
+	}
 }
 
 
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -458,11 +458,11 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_PAGEUP:
-			mMapView->moveView(MapOffsetUp);
+			changeViewDepth(0, MapOffsetUp);
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_PAGEDOWN:
-			mMapView->moveView(MapOffsetDown);
+			changeViewDepth(0, MapOffsetDown);
 			break;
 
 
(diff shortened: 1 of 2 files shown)


DanRStevens:
Oh interesting, and unexpected.

I edited the lead comment to link to the issue. Using the lead message with `closes` allows the issue to auto close when the PR is merged. I also adjusted the title. I figure a description is more useful in the title for browsing than an issue number. Plus, as I've just discovered, the issue doesn't get linked when referenced from the title, but does get linked when referenced in comments.

OPHD/States/MapViewState.cpp
```diff
@@ -725,18 +725,26 @@ void MapViewState::onSystemMenu()
 
 
 /**
- * Changes the current view depth.
+ * Changes the current view depth. If a value is given for the 2nd parameter, the first will be ignored.
  */
-void MapViewState::changeViewDepth(int depth)
+void MapViewState::changeViewDepth(int depth, std::optional<MapOffset> direction)
```
DanRStevens, line 730: This seems to complicate the semantics a bit. Though perhaps the semantics of the view change code are already a bit more complicated than I'd like.

I'm wondering if we need a separate method `onDepthChange`, split off from `changeViewDepth`, which can handle the side effects of changing the depth. Right now we have a couple ways of changing the depth, either by offsetting it from the old value, or by setting it directly. Both might potentially clamp the values to the map dimensions. Both might (potentially) need to run side effects of a depth change. By moving the side effects to a separate function, it would be possible to repeat them when either offsetting or setting directly.

I noticed the `changePoliceOverlayDepth` function expects to know both the `oldDepth` and the `newDepth`. I suppose an `onDepthChange` method could take both values when running the event. Though I wonder if `changePoliceOverlayDepth` really needs both values. I haven't looked at it in ages, and it seems odd to me that it should need to know the old depth.

OPHD/States/MapViewState.cpp
```diff
@@ -458,11 +458,11 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_PAGEUP:
-			mMapView->moveView(MapOffsetUp);
+			changeViewDepth(0, MapOffsetUp);
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_PAGEDOWN:
-			mMapView->moveView(MapOffsetDown);
+			changeViewDepth(0, MapOffsetDown);
```
DanRStevens, line 465: Rather than changing the call here, I think maybe we need to add an extra call to the proposed `onChangeDepth` event method.

OPHD/States/MapViewState.cpp
alynsarcana: Another option might be to only use one method of changing the depth. In this case I would suggest setting it directly, as that appears to be the only that clamps the values. The key press event handlers could calculate what the value would be for a PageUp/Down press and pass that into the existing changeViewDepth() function. 

Something like the for the PageUp event for example:
`int depth = mMapView->currentDepth() + MapOffsetUp.z;`
`changeViewDepth(depth);`

DanRStevens: Values are always clamped. That's especially important for relative offsets.

For relative offsets, when calling `MapView::moveView`, it offsets the values and calls `MapView::mapViewLocation`, which handles the clamping. The clamping is direct for `x` and `y`. For the `z` value, there's a further call to `MapView::currentDepth` which does the clamping.

For directly setting the depth, `MapViewState::changeViewDepth` makes a direct call to `MapView::currentDepth`, same as above, to clamp the `z` value.

----

Perhaps the issue here is there is a varying level of abstraction. For offsetting values, there are direct calls to the `MapView` object which handles the offsetting and the clamping. For setting a `z` value directly, there is instead a call on the `MapViewState` object, which then call `MapView` functions. We should perhaps always be calling a method on the same object if we want to maintain a consistent level of abstraction.

The `MapView` class doesn't know about `MapViewState`, so it can't trigger events on `MapViewState` for a changed `z` value. At least, it can't notify it directly. Maybe with a `Signal` or `Delegate` it could trigger something, but I'd like to avoid that kind of complexity.

Maybe what we need is to route both types of updates through methods on `MapViewState`, so it can intercept changes to `z` values and run a method such as `onDepthChange` suggested above. In that way, it can be a simple call when a `z` value change is detected. Perhaps we need a `MapViewState::moveView` method, which delegates to `MapView::moveView`, but also includes a check for changed `z` values.

I think I would check for changes to `z` by recording the old value before doing the update, and checking if it changed after the update. That way, if there's an attempt to change the `z` value, but clamping prevents any real change, we can avoid running code for the side effects. That would be more accurate than checking if there's an attempt to change the `z` value, which might be prevented by clamping.

DanRStevens: I'm maybe a little worried my thoughts above were a bit too abstract. Maybe I should outline kind of what I'm thinking in terms of refactoring I might attempt.

1. Split `MapViewState::changeViewDepth` so side effects are handled by a new method `MapViewState::onChangeDepth`.
2. Add a new method `MapViewState::moveView`, which delegates to `MapView::moveView`
3. Update calls in `MapViewState` to use `MapViewState::moveView` instead of directly calling `MapView::moveView`
4. Update `MapViewState::moveView` to detect `z` changes, and call the new `onChangeDepth` method. This would fix the bug.
5. Optional: Update `MapViewState::changeViewDepth` to detect `z` changes (which might have been prevented from clamping), and only run the `onChangeDepth` method when there's an actual real `z` value change. This would improve the efficiency of already working code.

ldicker83: I'm with Dan on this one. I've recently been reading the book Five Lines of Code and the refactoring suggestions match what Dan outlines above.

alynsarcana: Those changes make sense. Also, thanks for the new book suggestion. I'll have to take a look at that.

OPHD/States/MapViewState.cpp
```diff
@@ -458,11 +458,11 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_PAGEUP:
-			mMapView->moveView(MapOffsetUp);
+			moveView( MapOffsetUp);
```
DanRStevens, line 461: Inconsistent spacing

DanRStevens (dismissed):
Looks good. One minor spacing inconsistency I noticed.

I'll push the branch to the repo so the tests run.

DanRStevens:
Aha, I've discovered how to fetch the changes without having to first add a new remote for the fork. Seems there are some GitHub refs that can be used, which don't normally appear in branch lists.

For example:
```
git fetch origin pull/1445/head:fixPgUpPgDown
git co fixPgUpPgDown
git push -u origin fixPgUpPgDown
```

That's quite a bit more convenient than adding a new remote.

Anyway, tests are running. The Windows build can take somewhere from 5-15 minutes. I won't be sticking around though. Maybe the whitespace thing will be fixed before I revisit.

DanRStevens:
When testing this locally I noticed one oddity, which I haven't really looked into yet. The hotkeys all work with these changes, but the mouse buttons no longer switch the construction menu. Checking against `main`, it seems the mouse buttons did switch the construction menu before.

It's curious because there were no obvious changes to the mouse button handling code. Maybe there were extra calls to handle the level change side effects which are no longer getting called due to the z-level change detection? I know a few places in the game checked certain conditions every frame drawn, not simply when a state change was made.

DanRStevens:
In `MapViewState::onMouseDown`, there is a call to `changeViewDepth`, which should be updated to call `onChangeDepth` instead:
```cpp
		const auto oldDepth = mMapView->currentDepth();
		mNavControl->onClick(MOUSE_COORDS);
		if (oldDepth != mMapView->currentDepth())
		{
			changeViewDepth(mMapView->currentDepth());
		}
```

DanRStevens (approved):
Excellent. I just ran a manual test and this fixes the mouse button issue I mentioned earlier.

Issue #1422, opened 2023-06-06
Split OPHD game mechanics into libOPHD


DanRStevens:
Continuation of:
- Issue #1191

In the previous case, a number of projects were setup to house code for various aspects of OPHD, and it's testing and development. Now we need to start splitting off and moving code into those projects.

In paritcular, this Issue is meant to track moving code relating to core game mechanics into `libOPHD`. This is likely to include:
- Maps
- MapObjects (Robots, Structures)
- Mines (should probably become part of MapObjects)
- Resources (ore/minerals)
- Population models
- Any other similar code

PR #1449, opened 2024-04-15
Feature/map offset operator overloads

oscar139:
Added project file for tests for OPHD. And added the operator overloads for MapOffset for neater scalar multiplication.

### DIFF
--- a/OPHD/Map/MapOffset.h
+++ b/OPHD/Map/MapOffset.h
@@ -8,3 +8,35 @@ struct MapOffset
 	NAS2D::Vector<int> xy;
 	int z;
 };
+
+
+inline bool operator==(const MapOffset& offset1, const MapOffset& offset2)
+{
+	return offset1.xy == offset2.xy && offset1.z == offset2.z;
+}
+
+
+inline bool operator!=(const MapOffset& offset1, const MapOffset& offset2)
+{
+	return !(offset1 == offset2);
+}
+
+
+inline MapOffset& operator*=(MapOffset& offset, int scalar)
+{
+	offset.xy *= scalar;
+	offset.z *= scalar;
+	return offset;
+}
+
+
+inline MapOffset operator*(MapOffset offset, int scalar)
+{
+	return offset *= scalar;
+}
+
+
+inline MapOffset operator*(int scalar, MapOffset offset)
+{
+	return offset *= scalar;
+}
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -741,7 +741,6 @@ void MapViewState::onChangeDepth(int oldDepth, int newDepth) {
 
 void MapViewState::moveView(MapOffset offset) {
 	int oldZLevel = mMapView->currentDepth();
-	
 	mMapView->moveView(offset);
 
 	int newZLevel = mMapView->currentDepth();
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -421,22 +421,22 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
 
 		case NAS2D::EventHandler::KeyCode::KEY_w:
 		case NAS2D::EventHandler::KeyCode::KEY_UP:
-			mMapView->moveView(MapOffset{DirectionNorthWest * shiftMoveScalar, 0});
+			moveView(MapOffset{DirectionNorthWest * shiftMoveScalar, 0});
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_s:
 		case NAS2D::EventHandler::KeyCode::KEY_DOWN:
-			mMapView->moveView(MapOffset{DirectionSouthEast * shiftMoveScalar, 0});
+			moveView(MapOffset{DirectionSouthEast * shiftMoveScalar, 0});
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_a:
 		case NAS2D::EventHandler::KeyCode::KEY_LEFT:
-			mMapView->moveView(MapOffset{DirectionSouthWest * shiftMoveScalar, 0});
+			moveView(MapOffset{DirectionSouthWest * shiftMoveScalar, 0});
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_d:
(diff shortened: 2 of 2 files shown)


OPHD/Map/MapOffset.h
```diff
@@ -8,3 +8,36 @@ struct MapOffset
 	NAS2D::Vector<int> xy;
 	int z;
 };
+
```
ldicker83, line 11: Style guidelines: double spaces between methods and functions. Otherwise I'm not seeing anything blocking here. Would like to see some feedback from @DanRStevens before merging.

oscar139: Ok will fix. I add two more commits to actually use the new operators. Oh and we some how missed updating calls to the new MapViewState:moveView from mMapView::moveView in MapViewState::onKeyDown

OPHD/Map/MapOffset.h
```diff
@@ -8,3 +8,41 @@ struct MapOffset
...
+}
+
+
+inline MapOffset operator*(const MapOffset& offset, int scalar)
+{
+	MapOffset result;
+	result.xy = offset.xy * scalar;
+	result.z = offset.z * scalar;
+	return result;
+}
```
DanRStevens, line 31: When you also support `operator*=`, you can use that to implement `operator*`. Simply remove the `const &` from the parameter passing so it generates a copy, and then use `operator*=` on the copy.

```cpp
inline MapOffset operator*(MapOffset offset, int scalar)
{
	return offset *= scalar;
}
```

OPHD/States/MapViewState.cpp
```diff
@@ -421,22 +421,22 @@ void MapViewState::onKeyDown(NAS2D::EventHandler::KeyCode key, NAS2D::EventHandl
...
 		case NAS2D::EventHandler::KeyCode::KEY_a:
 		case NAS2D::EventHandler::KeyCode::KEY_LEFT:
-			mMapView->moveView(MapOffset{DirectionSouthWest * shiftMoveScalar, 0});
+			mMapView->moveView(MapOffset{DirectionSouthWest * shiftMoveScalar});
 			break;
 
 		case NAS2D::EventHandler::KeyCode::KEY_d:
 		case NAS2D::EventHandler::KeyCode::KEY_RIGHT:
-			mMapView->moveView(MapOffset{DirectionNorthEast * shiftMoveScalar, 0});
+			mMapView->moveView(MapOffset{DirectionNorthEast * shiftMoveScalar});
```
DanRStevens: This doesn't appear to be using the new `operator*`. It's still using the existing `operator*` on the `NAS2D::Vector<int>` class. The only change is there is no explicit initialization value for `z`. Since the `struct` is initialized with `{}`, unspecified fields should be default initialized, which in this case should result in a `0` for `z`.

I assume what you meant to write was:
```cpp
mMapView->moveView(MapOffset{DirectionNorthEast, 0} * shiftMoveScalar);
```

Of course with the new `operator*` that means we can go back to using the predefined `MapOffset` values from `DirectionOffset.h`:
```cpp
mMapView->moveView(MapOffsetNorthEast * shiftMoveScalar);
```

OPHD/States/MapViewState.cpp
```diff
@@ -741,7 +741,7 @@ void MapViewState::onChangeDepth(int oldDepth, int newDepth) {
 
 void MapViewState::moveView(MapOffset offset) {
 	int oldZLevel = mMapView->currentDepth();
-	
+
```
DanRStevens, line 744: I approve of whitespace fixes. Preferably they should be in a separate commit from code changes.

testOPHD/testOPHD.cpp
```diff
@@ -0,0 +1,6 @@
+#include <gtest/gtest.h>
+
+
+TEST(Test, EmptyTest)
+{
+}
```
DanRStevens, line 6: Slightly awkward situation, as we still haven't completed:
- Issue #1422

Ideally, `MapOffset` would be migrated to `libOphd`, and so unit test code for it would go into `testLibOphd`. Unfortunately that hasn't happened yet.

----

In C++ it's a bit more natural to write unit tests against library code. I wouldn't expect there to be a `testOphd` project, since the `Ophd` project is an app rather than a library.

----

The `makefile` doesn't support building the new project. It does support building `testLibOphd`.

testOPHD/MapOffsetTests.cpp
```diff
@@ -0,0 +1,105 @@
+#include "OPHD/Map/MapOffset.h"
```
DanRStevens, line 1: As `MapOffset` is expected to be in library code, and the unit tests for that library are in a separate project, it would be more appropriate for the `#include` to use `<>` rather than `""`.

testOPHD/MapOffsetTests.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+	MapOffset offset1;
+	offset1.xy = { 1, 2 };
+	offset1.z = 3;
+
+	MapOffset offset2;
+	offset2.xy = { 1, 2 };
+	offset2.z = 3;
+
+	EXPECT_TRUE(offset1 == offset2);
```
DanRStevens, line 16: This would be more terse with inline construction:
```cpp
EXPECT_TRUE(MapOffset{{1, 2}, 3} == MapOffset{{1, 2}, 3});
```

----

Additionally, this particular check might use `EXPECT_EQ` instead of `EXPECT_TRUE`. Not that it's much shorter or really changes readability.

```cpp
EXPECT_EQ(MapOffset{{1, 2}, 3}, MapOffset{{1, 2}, 3});
```

----

Similar for other tests too. Use inline construction. Using more specific expectations, such as `EXPECT_NE` (not equal).

testOPHD/MapOffsetTests.cpp
```diff
@@ -0,0 +1,105 @@
...
+	EXPECT_EQ(result3, expectedResult3);
+
+	int zeroScalar = 0;
+	MapOffset result4 = offset * zeroScalar;
+	MapOffset expectedResult4 = {0, 0, 0};
+	EXPECT_EQ(result4, expectedResult4);
+
+	MapOffset result5 = offset *= scalar;
+	MapOffset expectedResult5 = {-2, 4, 6};
+	EXPECT_EQ(result5, expectedResult5);
```
DanRStevens, line 104: Might be wise to split tests for `operator*=` from the tests for `operator*`.

PR #1451, opened 2024-05-16
Add listbox to research report

ldicker83:
Adds ListBox to ResearchReport UI. Addresses #27.

This simply adds the listbox with sorted available research topics. It doesn't do anything else. Opting to break this down into much smaller parts so I don't have a colossal change set.

### DIFF
--- a/OPHD/UI/Reports/ResearchReport.cpp
+++ b/OPHD/UI/Reports/ResearchReport.cpp
@@ -63,8 +63,8 @@ ResearchReport::ResearchReport() :
 		button->toggle(false);
 	}
 
-
 	add(lstResearchTopics, {});
+	lstResearchTopics.selectionChanged().connect({this, &ResearchReport::handleTopicChanged});
 
 	const Point<int> buttonStartPosition{rect().position.x + MarginSize * 3 + CategoryIconSize, rect().position.y + MarginSize * 2 + fontBigBold.height()};
 	const int buttonSpacing = btnAllTopics.size().x + MarginSize;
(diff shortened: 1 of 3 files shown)


OPHD/UI/Reports/ResearchReport.cpp
```diff
@@ -95,37 +100,84 @@ ResearchReport::~ResearchReport()
...
+		panel.selected = false;
+	}
 
-	IconArea = {
-		{rect().position.x + MarginSize * 3 + CategoryIconSize,
-		rect().position.y + fontBigBold.height() + btnAllTopics.size().y + MarginSize * 3},
-		{((rect().size.x / 3) * 2) - (MarginSize * 4) - CategoryIconSize,
-		rect().size.y - MarginSize * 4 - fontBigBold.height() - btnAllTopics.size().y}};
+	mCategoryPanels.front().selected = true;
+	mSelectedCategory = &mCategoryPanels.front();
```
DanRStevens, line 179: If you swap these lines, you can avoid calling `.front()` twice.

OPHD/UI/Reports/ResearchReport.cpp
```diff
@@ -95,37 +100,84 @@ ResearchReport::~ResearchReport()
...
+	setSectionRects();
+
+	lstResearchTopics.area(mResearchTopicArea);
+}
+
+
+void ResearchReport::setSectionRects()
+{
+	mCategoryIconArea =
+	{
```
DanRStevens, line 135: This is a function with both a significantly complex calculation, and a side effect (assigning to member variables). It may be wise to separate the two actions, particularly if you want to add tests for the calculation at some point.

Instead, I would suggest calculation functions that return the resulting value, which would be separate from any function that might call them and assign the result to member variables.

----

If the write is the only access to member variables, than a separated calculation function could be left at module level scope (as opposed to making it a `static` member function). That limits visibility since it no longer needs to be declared in a header file.

ldicker83: Totally agreed, something I plan to do in a later change set.

PR #1452, opened 2024-05-17
Add topic description

ldicker83:


OPHD/UI/Reports/ResearchReport.cpp
```diff
@@ -93,6 +97,12 @@ ResearchReport::ResearchReport() :
 	{
 		buttons[i]->position(buttonStartPosition + Vector<int>{buttonSpacing * static_cast<int>(i), 0});
 	}
+
+	const auto rects = std::array{&mCategoryIconArea, &mResearchTopicArea, &mTopicDetailsArea};
+	for (auto rect : rects)
+	{
+		RectList.push_back(rect);
+	}
```
DanRStevens: Could have just used the `std::array` as defined, and not copied it to a new data type.

Not that it really matters since this was just a temporary thing.

PR #1456, opened 2024-05-18
Check for lab availability

ldicker83:
Small change to check for lab availability. Originally was going to hide the panel and display a message about needing labs to be built before showing any of the research but I think that would unnecessarily burden the player. In the end I'll be adding a button to start research that can be enabled/disabled depending on if necessary labs are available.


OPHD/StructureManager.h
```diff
@@ -121,6 +121,20 @@ class StructureManager
...
+		unsigned int count = 0;
+		for (const auto* structure : getStructures<StructureType>())
+		{
+			if (structure->state() == state)
+			{
+				++count;
+			}
+		}
+		return count;
+	}
```
DanRStevens, line 136: I was right about to comment this is a generic method to count operational structures (of any type), and should be generalized. Glad to see you made that change.

ldicker83: Yup. Noticed that and was like yeah I can do better.

Not sure why I changed tense in the commit notes for this one but eh, it's whatever :D

PR #1464, opened 2024-07-25
Refactor/move map offset to libOPHD

oscar139:
Addressing Issue #1422


libOPHD/libOPHD.vcxproj.filters
```diff
@@ -77,8 +80,8 @@
     <ClInclude Include="XmlSerializer.h">
       <Filter>Header Files</Filter>
     </ClInclude>
-  </ItemGroup>
-  <ItemGroup>
-    <None Include="..\.clang-format" />
+    <ClInclude Include="Map\MapOffset.h">
+      <Filter>Header Files\Map</Filter>
+    </ClInclude>
```
DanRStevens, line 85: Any idea why the reference to `.clang-format` was removed?

Or for that matter, any idea why it was there in the first place? Does having the reference there make it easy to open and edit the file? I haven't used Visual Studio in ages, so I'm not sure what the developer experience is like in that environment.

testLibOPHD/testLibOPHD.vcxproj
```diff
@@ -163,6 +163,7 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="main.cpp" />
+    <ClCompile Include="MapOffset.cpp" />
```
DanRStevens, line 166: In the NAS2D unit test project, test files usually have the ending ".test.cpp". Not sure if that standard is super important, though it does make the filename tabs visually distinct when both the main file and the test file are open in an editor.

----

Test projects need at least one source file. That meant including dummy `main.cpp` files for empty projects. Once at least one valid test source file is added, the dummy `main.cpp` can be removed. (Might be best to handle that as a separate followup commit).

DanRStevens (commented):
This looks like largely the correct update.

libOPHD/libOPHD.vcxproj.filters
DanRStevens: I undid the `.clang-format` change, since it seemed to be unintentional.

I don't entirely understand why that entry is there, or if we need it, though it sounds like that's a decision to make somewhere else.

testLibOPHD/testLibOPHD.vcxproj
```diff
@@ -162,7 +162,7 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
-    <ClCompile Include="main.cpp" />
+    <ClCompile Include="MapOffset.cpp" />
```
DanRStevens, line 165: Just noting that we never renamed this file to `MapOffset.test.cpp`, which would match the convention in other test projects.

DanRStevens (approved):

Issue #1231, opened 2022-07-30
Morale should stay steady for a few turns after landing


ldicker83:
#1230 will make morale steady while colonists are still onboard the starship.

In addition, morale should remain steady for a few turns after the population has been dropped to the surface.

As with #1229, the buffer period should be adjusted based on difficulty level. Default ('hard mode') should be 15 turns with no crime updates from the time that colonists are landed (just over 'one year')

Issue #1230, opened 2022-07-30
Morale starts dropping before colonists have landed


ldicker83:
Noted by @Inglonias in #1229;

Morale should remain steady until colonists have been landed on the surface.

ldicker83:
> Could maybe add a "Just Landed" morale bonus, that counteracts the initial colony conditions. Would make sense, as I'm sure people would be happy to have arrived after such a long journey, and there could be some excitement over starting a new colony.

Not a bad idea. The sort of excitement of starting anew.

> Might also skip the morale calculation code if there are no colonists.

That was the general idea that I had. I created the issue when I noticed it as a reminder to address it.

DanRStevens:
Looks like this was fixed by PR #1466. The head message even states it fixes this issue, though because the "Fixes" word wasn't repeated for each linked issue, it didn't close all of them.

PR #1466, opened 2024-10-01
Added difficulty-dependent "grace period" for crime and morale

### LINKED ISSUE #1229: Crime shouldn't start until some amount of time after landing
Related to #1228, criminal activity should probably not start until some amount of time _after_ the population has landed when the colonists have had some time to settle into their new home.

The idea behind this is that 1) colonists are going to be too busy adjusting to their new life to plan and commit crimes and 2) from a gameplay standpoint we want to give the player at least a few turns to establish themselves before crime is something they need to worry about.

This buffer period should be adjusted based on difficulty level. Default ('hard mode') should be 15 turns with no crime updates from the time that colonists are landed (just over 'one year')

andrewsg:
Skips morale and crime calculation until a difficulty-level-dependent "grace period" of 15-30 turns after human colonists have landed.

Fixes #1229, #1230, #1231

### DIFF
--- a/OPHD/States/MapViewState.h
+++ b/OPHD/States/MapViewState.h
@@ -214,6 +214,7 @@ private:
 	void updateCommercial();
 	void updateMaintenance();
 	void updateMorale();
+	void notifyBirthsAndDeaths();
 	void updateResidentialCapacity();
 	void updateBiowasteRecycling();
 	void updateResources();
@@ -299,12 +300,25 @@ private:
 	TechnologyCatalog mTechnologyReader;
 
 	Planet::Attributes mPlanetAttributes;
-	Difficulty mDifficulty = Difficulty::Medium;
 
 	int mFood{0};
 
+	// DIFFICULTY
+	Difficulty mDifficulty = Difficulty::Medium;
+
+	// Length of "honeymoon period" of no crime/morale updates after landing, in turns
+	std::map<Difficulty, int> gracePeriod
+	{
+		{Difficulty::Beginner, 30},
+		{Difficulty::Easy, 25},
+		{Difficulty::Medium, 20},
+		{Difficulty::Hard, 15}
+	};
+
 	// MISCELLANEOUS
 	int mTurnCount = 0;
+
+	int mTurnNumberOfLanding = std::numeric_limits<int>::max(); /**< First turn that human colonists landed. If never landed, default is int max (representing the future). */
 
 	int mCurrentMorale;
 	int mPreviousMorale;
--- a/OPHD/States/MapViewState.h
+++ b/OPHD/States/MapViewState.h
@@ -11,6 +11,8 @@
 #include "../StorableResources.h"
 #include "../RobotPool.h"
 
+#include "../Constants/Numbers.h"
+
 #include "../Map/MapCoordinate.h"
 
 #include "../MapObjects/Robot.h"
@@ -318,7 +320,7 @@ private:
 	// MISCELLANEOUS
 	int mTurnCount = 0;
 
-	int mTurnNumberOfLanding = std::numeric_limits<int>::max(); /**< First turn that human colonists landed. If never landed, default is int max (representing the future). */
+	int mTurnNumberOfLanding = constants::ColonyShipOrbitTime; /**< First turn that human colonists landed. */
 
 	int mCurrentMorale;
 	int mPreviousMorale;
--- a/OPHD/States/MapViewStateEvent.cpp
+++ b/OPHD/States/MapViewStateEvent.cpp
@@ -89,6 +89,7 @@ void MapViewState::onFactoryProductionComplete(Factory& factory)
  */
 void MapViewState::onDeployColonistLander()
 {
+	mTurnNumberOfLanding = mTurnCount;
 	mPopulation.addPopulation({0, 10, 20, 20, 0});
 }
 
--- a/OPHD/States/MapViewStateEvent.cpp
+++ b/OPHD/States/MapViewStateEvent.cpp
@@ -89,7 +89,9 @@ void MapViewState::onFactoryProductionComplete(Factory& factory)
  */
 void MapViewState::onDeployColonistLander()
 {
-	mTurnNumberOfLanding = mTurnCount;
+	if (mTurnNumberOfLanding > mTurnCount) {
+		mTurnNumberOfLanding = mTurnCount;
+	}
 	mPopulation.addPopulation({0, 10, 20, 20, 0});
 }
 
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -171,6 +171,7 @@ void MapViewState::save(const std::string& filePath)
 			{"prev_morale", mPreviousMorale},
 			{"colonist_landers", mLandersColonist},
 			{"cargo_landers", mLandersCargo},
+			{"turn_number_of_landing", mTurnNumberOfLanding},
 			{"children", population.child},
 			{"students", population.student},
 			{"workers", population.worker},
@@ -565,6 +566,8 @@ void MapViewState::readPopulation(NAS2D::Xml::XmlElement* element)
 
 		mCurrentMorale = dictionary.get<int>("morale");
 		mPreviousMorale = dictionary.get<int>("prev_morale");
+
+		mTurnNumberOfLanding = dictionary.get<int>("turn_number_of_landing");
 
 		const auto meanCrimeRate = dictionary.get<int>("mean_crime", 0);
 
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -567,7 +567,7 @@ void MapViewState::readPopulation(NAS2D::Xml::XmlElement* element)
 		mCurrentMorale = dictionary.get<int>("morale");
 		mPreviousMorale = dictionary.get<int>("prev_morale");
 
-		mTurnNumberOfLanding = dictionary.get<int>("turn_number_of_landing");
+		mTurnNumberOfLanding = dictionary.get<int>("turn_number_of_landing", constants::ColonyShipOrbitTime);
 
 		const auto meanCrimeRate = dictionary.get<int>("mean_crime", 0);
 
--- a/OPHD/States/MapViewStateTurn.cpp
+++ b/OPHD/States/MapViewStateTurn.cpp
@@ -21,7 +21,6 @@
 #include <vector>
 #include <algorithm>
 
-
 namespace
 {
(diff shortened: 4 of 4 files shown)


andrewsg:
Existing behavior may change slightly in that crime won't pause if the population drops to zero after landing. I believe this change will not impact gameplay in a practical sense, but we can easily re-add the zero population check if needed.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -723,7 +732,15 @@ void MapViewState::nextTurn()
 	updateMaintenance();
 	updateCommercial();
 	updateBiowasteRecycling();
-	updateMorale();
+
+	// Morale will not change until at least n turns from landing, depending on difficulty
+	if (turnsSinceLanding > gracePeriod[mDifficulty])
```
andrewsg, line 737: This is separate from the crime check above so as not to change the sequential order of crime update, population update, morale update.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -747,6 +764,12 @@ void MapViewState::nextTurn()
 
 	mMineOperationsWindow.updateTruckAvailability();
 
+	// If this is the first turn with population, then set mTurnNumberOfLanding
+	if (mPopulation.getPopulations().size() > 0 && mTurnCount < mTurnNumberOfLanding)
+	{
+		mTurnNumberOfLanding = mTurnCount;
+	}
```
DanRStevens, line 771: Maybe we should consider setting an additional member field here (which also defaults to `std::numeric_limits<int>::max()`).
```cpp
mMoraleActiveTurn = mTurnNumberOfLanding + gracePeriod[mDifficulty]
```

That means we'd need to do the `gracePeriod` lookup once, when colonists are placed, rather than every turn. 

A new field also means we can simplify checks. Instead of:
```cpp
if (turnsSinceLanding > gracePeriod[mDifficulty])
```
We could instead check:
```cpp
if (mTurnCount > mMoraleActiveTurn)
```
That could allow us to remove the calculation for `turnsSinceLanding`, which was also calculated every turn. It also means you don't need to worry about negative values.

Of course such a new field would also need to be set when loading saved games. It could be calculated in the same way as here, using `mTurnNumberOfLanding`, though only if `mTurnNumberOfLanding` is not `std::numeric_limits<int>::max()`. We would want to avoid overflow in that case.

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -710,7 +716,10 @@ void MapViewState::nextTurn()
 
 	updateResidentialCapacity();
 
-	if (mPopulation.getPopulations().size() > 0)
+	int turnsSinceLanding = mTurnCount - mTurnNumberOfLanding; // If negative, landing has not yet occurred.
+
+	// Colony will not have a crime rate until at least n turns from landing, depending on difficulty
+	if (turnsSinceLanding > gracePeriod[mDifficulty])
```
DanRStevens, line 722: This part relates to the last comment.

OPHD/States/MapViewState.h
```diff
@@ -214,6 +214,7 @@ class MapViewState : public Wrapper
 	void updateCommercial();
 	void updateMaintenance();
 	void updateMorale();
+	void notifyBirthsAndDeaths();
```
DanRStevens, line 217: I kind of like that you moved the notification code to a separate function from the `updateMorale` function.

OPHD/States/MapViewStateIO.cpp
```diff
@@ -566,6 +567,8 @@ void MapViewState::readPopulation(NAS2D::Xml::XmlElement* element)
 		mCurrentMorale = dictionary.get<int>("morale");
 		mPreviousMorale = dictionary.get<int>("prev_morale");
 
+		mTurnNumberOfLanding = dictionary.get<int>("turn_number_of_landing");
```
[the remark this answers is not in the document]
andrewsg: I've added a default of `constants::ColonyShipOrbitTime`. For posterity, here's my explanation from Discord:

> This will work correctly for people who are importing a save that's older than `constants::ColonyShipOrbitTime + gracePeriod`. It will also work correctly for people who are importing a save where they haven't landed yet.
> 
> Saves who have landed, but haven't hit colony deorbit, will start their grace period immediately even if they have already had some morale/crime changes. Saves who have exceeded colony deorbit will start their grace period immediately but it will end at `constants::ColonyShipOrbitTime + gracePeriod`.

OPHD/States/MapViewStateTurn.cpp
andrewsg: After considering this, I think I would suggest leaving it as it is due to the additional code complexity. Alternatively, we could switch to using `constexpr` arrays to store difficulty-dependent data in a way that doesn't involve a runtime lookup at all. If you still prefer I make this change, I'll see what I can do to keep the complexity down.

OPHD/States/MapViewState.h
andrewsg: Thanks!

OPHD/States/MapViewStateIO.cpp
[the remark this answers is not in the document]
DanRStevens: This appears to address the concern about a default value quite nicely.

OPHD/States/MapViewStateTurn.cpp
DanRStevens: Hmm, it's curious to hear that. My suggestion was mainly about reducing complexity, though perhaps I phrased it so it sounded like there was an emphasis on efficiency, which was not the intent.

----

Currently the code has:
```cpp
int turnsSinceLanding = mTurnCount - mTurnNumberOfLanding; // If negative, landing has not yet occurred.
```

This is then used a couple of times in a check:
```cpp
if (turnsSinceLanding > gracePeriod[mDifficulty])
```

If we substitute the expression into that check, we get:
```cpp
if (mTurnCount - mTurnNumberOfLanding > gracePeriod[mDifficulty])
```

Re-arranging, we have:
```cpp
if (mTurnCount > mTurnNumberOfLanding + gracePeriod[mDifficulty])
```

The suggestion I made was to store a field:
```cpp
mMoraleActiveTurn = mTurnNumberOfLanding + gracePeriod[mDifficulty]
```

Substituting that into the re-arranged check we get:
```cpp
if (mTurnCount > mMoraleActiveTurn)
```

This seems simpler, at least for the check.

Actually, we never really use `mTurnNumberOfLanding`, other than in the check, which can be more simply written if we had a field for `mMoraleActiveTurn`. I wrote my suggestion as adding a new field, though potentially we could replace one field with the other.

By adjusting the field we store, we can front-load most of the complexity for the check to the point when colonists are first landed.

Storing a different field might mean we change which field gets written to a saved game file. Though the suggestion for the new field field, with a default of `constants::ColonyShipOrbitTime + gracePeriod[mDifficulty]` should behave identically to the current proposal.

Alternatively, we could translate between the current field and the new field suggestion during save/load. That would add some complexity. It wouldn't be a whole lot of extra complexity, though it could be avoided if we settle on a new field before saved games start proliferating with a slightly different field we might then feel the need to continue supporting.

I guess one question here is are there other reasons we'll need to know `mTurnNumberOfLanding`, or is it likely to only ever be used in this one place? If there are other reasons for needing `mTurnNumberOfLanding`, then it may be best to store it as is. If not, using `mMoraleActiveTurn` may allow for simpler code.

DanRStevens: Given Leeor's comment on Discord about potentially allowing the difficulty setting to be modified during gameplay, perhaps it's best to keep the current design, using the `mTurnNumberOfLanding` as the primary field, and doing the `gracePeriod` lookup each turn.

DanRStevens (commented):
Given the comments on Discord about possibly altering the difficulty level during gameplay, I think this design is what we want.

I made a few minor suggestions here, which either weren't thought of or weren't formalized during the last review.

There is one comment about consistent default values, and a related note about overflow, which could be relevant to future saved game loading if allowed to proliferate. For that reason I'll hold off on giving formal approval so that can be responded to. Otherwise, this looks about ready to be merged.

OPHD/States/MapViewStateTurn.cpp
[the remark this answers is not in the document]
DanRStevens: Was just thinking this suggestion could have gone further, and set a `bool` variable, rather than repeating a boolean expression in each `if` condition.

```cpp
bool isMoraleEnabled = mTurnCount > mTurnNumberOfLanding + gracePeriod[mDifficulty];
```

OPHD/States/MapViewStateEvent.cpp
```diff
@@ -89,6 +89,7 @@ void MapViewState::onFactoryProductionComplete(Factory& factory)
  */
 void MapViewState::onDeployColonistLander()
 {
+	mTurnNumberOfLanding = mTurnCount;
```
DanRStevens, line 92: That's nice and simple.

I suppose I should bring up a section from the last review:
> We would still need to retain the second check, since there can be multiple landers, and we probably want to activate on the first landing, and not reset the value upon a subsequent landing.

I'm fine with resetting `mTurnNumberOfLanding` on a subsequent landing. There wasn't really a clear specification on desired behavior here. Just wanted to draw attention to this in case it was an accidental change.

Or maybe @ldicker83 would like to weigh in on desired behavior in this case. Should the field represent the turn of the first landing, or the turn of the last landing? With multiple landers, the two can be different.

DanRStevens (approved):
This looks good to merge.

I did have one minor comment, about an underspecified area, but that's something easy to change at a later date, and shouldn't have far reaching consequences if we do change it.

OPHD/States/MapViewStateEvent.cpp
andrewsg: Whoops, I did miss this. Thank you for pointing that out. Since you've asked for confirmation I'll wait to amend it until that's answered, but it is unintentional on my part.

DanRStevens: Doesn't seem like Leeor is able to respond right now. Given that change was accidental, I think maybe just go ahead and re-add the check `if (mTurnNumberOfLanding > mTurnCount)` around that line. I think that's probably what everyone is expecting.

Plus, it avoids a possible case where morale can become active due to an early landing, and then gets disabled again because more colonists land, leaving morale paused in a slightly altered state. That might be a bit unexpected.

----

Alternatively, I suppose you could use [`std::min`](https://en.cppreference.com/w/cpp/algorithm/min) instead of an `if` check:
```cpp
mTurnNumberOfLanding = std::min(mTurnNumberOfLanding, mTurnCount);
```

andrewsg: Done.

DanRStevens (approved):

PR #1468, opened 2024-11-19
Feature/colony ship morale multiplier

oscar139:


DanRStevens (approved):

OPHD/States/MapViewStateTurn.cpp
```diff
@@ -717,7 +717,7 @@ void MapViewState::nextTurn()
 	updateResidentialCapacity();
 
 	// Colony will not have morale or crime effects until at least n turns from landing, depending on difficulty
-	bool isMoraleEnabled = mTurnCount > mTurnNumberOfLanding + gracePeriod[mDifficulty];
+	bool isMoraleEnabled = mTurnCount > mTurnNumberOfLanding + GracePeriod.at(mDifficulty);
```
DanRStevens, line 720: Ahh yes. I see you hit the subtle issue where `operator[]` only works on mutable objects, since it inserts a new value when it can't find an existing one.

Probably safer too, since we can now expect an exception in case the difficulty value isn't found in the table.

DanRStevens (approved):

PR #1473, opened 2024-12-10
Feature/multi line tool tip

oscar139:
Changes allow the Tool Tip Class to support tool-tips with multiple lines.

### DIFF
--- a/libControls/ToolTip.cpp
+++ b/libControls/ToolTip.cpp
@@ -4,11 +4,34 @@
 #include <NAS2D/Utility.h>
 #include <NAS2D/Renderer/Renderer.h>
 
+#include <string>
+#include <sstream>
+#include <functional>
 
 namespace
 {
 	constexpr int MarginTight{2};
 	constexpr auto PaddingSize = NAS2D::Vector{MarginTight, MarginTight};
+
+	void forEachLine(const std::string& text, const std::function<void(const std::string&)>& lineHandler)
+	{
+		std::istringstream stream(text);
+		for (std::string line; std::getline(stream, line);)
+		{
+			lineHandler(line);
+		}
+	}
+
+
+	// Returns the maximum width of the largest string from the string stream in pixels.
+	int calculateMaxWidthStringSize(const std::string& text, const NAS2D::Font& font)
+	{
+		int maxWidthStringSize = 0;
+		forEachLine(text, [&font, &maxWidthStringSize](const std::string& lineStr) {
+			maxWidthStringSize = std::max(maxWidthStringSize, font.size(lineStr).x);
+		});
+		return maxWidthStringSize;
+	}
 }
 
 
@@ -42,7 +65,11 @@ void ToolTip::add(Control& c, const std::string& str)
 
 void ToolTip::buildDrawParams(std::pair<Control*, std::string>& item, int mouseX)
 {
-	const auto toolTipSize = mFont.size(item.second) + PaddingSize * 2;
+	const auto toolTipLineHeight = mFont.height();
+	const auto numberOfLines = static_cast<int>(std::count(item.second.begin(), item.second.end(), '\n') + 1);
+	const auto toolTipTextHeight = toolTipLineHeight * numberOfLines;
+	const auto toolTipTextWidth = calculateMaxWidthStringSize(item.second, mFont);
+	const auto toolTipSize = NAS2D::Vector{toolTipTextWidth, toolTipTextHeight} + PaddingSize * 2;
 
 	auto tooltipPosition = item.first->position();
 
@@ -102,6 +129,12 @@ void ToolTip::draw() const
 		auto& renderer = NAS2D::Utility<NAS2D::Renderer>::get();
 		renderer.drawBoxFilled(rect(), NAS2D::Color::DarkGray);
 		renderer.drawBox(rect(), NAS2D::Color::Black);
-		renderer.drawText(mFont, mFocusedControl->second, position() + PaddingSize);
+
+		auto linePosition = position() + PaddingSize;
+
+		forEachLine(mFocusedControl->second, [this, &renderer, &linePosition](const std::string& lineStr) {
+			renderer.drawText(mFont, lineStr, linePosition);
+			linePosition.y += mFont.height();
+		});
 	}
 }


DanRStevens (commented):
Thinking about the impact to the API of this change. Do we want to use a `std::vector` of strings, or do we want to support strings with embedded newlines? I'd be tempted to consider strings with embedded newlines. Seems like a more natural API to use. Particularly since sources of multi-line input may come with naturally embedded newlines.

libControls/ToolTip.cpp
```diff
@@ -42,7 +43,20 @@ void ToolTip::add(Control& c, const std::string& str)
...
+	const auto toolTipHeight = toolTipLineHeight * numberOfLines + PaddingSize.y * 2;
+
+	std::istringstream stream(item.second);
+	std::string line;
+	int maxWidth = 0;
+	while (std::getline(stream, line, '\n'))
+	{
+		maxWidth = std::max(maxWidth, mFont.size(line).x);
+	}
+	const auto toolTipWidth = maxWidth + PaddingSize.x * 2;
```
DanRStevens, line 57: I like how there's a final calculated result which is marked as `const`. Though I wonder if there's a way to calculate it that doesn't "leak" mutable local variables for the temporary values throughout the remainder of the function scope.

My first thought was if there was a ranged-for syntax to process a string by lines. Didn't see anything promising.

My second thought was perhaps a regular `for` loop could be used with multiple variable initialization. That would allow some of the temporary variables to be contained within the `for` block scope. Though it seems multiple variable initialization is only possible when they are all the same type.

My final thought, is perhaps the code could be extracted to a separate function. That would keep the drawing code at a high level, while separating the low level details of finding the max width of a sequence of lines.

libControls/ToolTip.cpp
```diff
@@ -102,6 +116,14 @@ void ToolTip::draw() const
...
-		renderer.drawText(mFont, mFocusedControl->second, position() + PaddingSize);
+
+		std::istringstream stream(mFocusedControl->second);
+		std::string line;
+		auto linePosition = position() + PaddingSize;
+		while (std::getline(stream, line, '\n'))
+		{
+			renderer.drawText(mFont, line, linePosition);
+			linePosition.y += mFont.height();
+		}
```
DanRStevens, line 127: This seems pretty close to duplicated code. Effectively there is a `forEachLine` functionality that has been duplicated, though with different details on what to do with those lines.

DanRStevens:
This might be of interest to a couple of those comments:
[`std::getline`](https://en.cppreference.com/w/cpp/string/basic_string/getline)
```cpp
    // read file line by line
    std::istringstream input;
    input.str("1\n2\n3\n4\n5\n6\n7\n");
    int sum = 0;
    for (std::string line; std::getline(input, line);)
        sum += std::stoi(line);
    std::cout << "\nThe sum is " << sum << ".\n\n";
```

The use of the `for` loop is interesting.

libControls/ToolTip.cpp
```diff
@@ -42,7 +43,21 @@ void ToolTip::add(Control& c, const std::string& str)
...
+		int maxWidth = 0;
+		processLines(text, [this, &maxWidth](const std::string& line) {
+			maxWidth = std::max(maxWidth, mFont.size(line).x);
+		});
+		return maxWidth;
+	};
+
+	const auto toolTipWidth = calculateMaxWidth(item.second) +PaddingSize.x * 2;
+
+	const auto toolTipSize = NAS2D::Vector<int>{toolTipWidth, toolTipHeight};
```
DanRStevens, line 60: Was just thinking, the individual offsets using `PaddingSize.x * 2` and `PaddingSize.y * 2` could have been handled at the end.

```cpp
const auto toolTipSize = NAS2D::Vector{toolTipWidth, toolTipHeight} + PaddingSize * 2;
```

Though at that point, maybe the intermediate values would be better named something like `textWidth` and `textHeight`.

----

I think the type parameter `<int>` can be inferred when both values are `int`.

libControls/ToolTip.cpp
```diff
@@ -102,6 +117,21 @@ void ToolTip::draw() const
...
+
+
+void ToolTip::processLines(const std::string& text, const std::function<void(const std::string&)>& lineProcessor) const
+{
+	std::istringstream stream(text);
+	for (std::string line; std::getline(stream, line);)
+	{
+		lineProcessor(line);
 	}
 }
```
DanRStevens, line 137: This function doesn't make use of any member fields of `ToolTip`, so also shouldn't be a regular member. A `static` member maybe, but perhaps it's best to remove this function from the class entirely, and put it in an unnamed namespace at the start of the .cpp file. The unnamed namespace will give it internal linkage, so it will never conflict with global names from other files during the link stage.

libControls/ToolTip.h
```diff
@@ -9,6 +9,7 @@
 
 #include <utility>
 #include <vector>
+#include <functional>
```
DanRStevens, line 12: Moving the new function declaration to an unnamed namespace also means you can move this include out of the header file. Leaving it in the header means it would be transitively included by anything including the `ToolTip.h` header.

Or, if you want to get rid of this include entirely, you can transform the method into a template, with a template parameter for the function object rather than using `std::function`. If you're not concerned about restricting the template type parameter to ensure it's something useful, the syntax should be fairly simple:
```cpp
template <typename Callback>
void processLines(const std::string& text, const Callback& lineProcessor) const;
```

Of course if you try to use a non-function type for the type parameter, such as an `int`, you may end up with cryptic error messages. There might be a way to restrict the type to a callable, though that would be more work.

libControls/ToolTip.cpp
```diff
@@ -42,7 +43,21 @@ void ToolTip::add(Control& c, const std::string& str)
...
+	const auto numberOfLines = static_cast<int>(std::count(item.second.begin(), item.second.end(), '\n') + 1);
+	const auto toolTipHeight = toolTipLineHeight * numberOfLines + PaddingSize.y * 2;
+
+	auto calculateMaxWidth = [this](const std::string& text) {
+		int maxWidth = 0;
+		processLines(text, [this, &maxWidth](const std::string& line) {
+			maxWidth = std::max(maxWidth, mFont.size(line).x);
+		});
+		return maxWidth;
+	};
```
DanRStevens, line 56: Might make sense to pull this section out as a standalone function. If `processLines` were standalone, perhaps in a unnamed namespace, then this could be made standalone too.

DanRStevens (commented):
I'm kind of liking how `processLines` cleans up some of that earlier duplication.

DanRStevens (approved):
I'm quite happy with this. Had a couple of small optional thoughts.

libControls/ToolTip.cpp
```diff
@@ -102,6 +129,12 @@ void ToolTip::draw() const
...
 		renderer.drawBoxFilled(rect(), NAS2D::Color::DarkGray);
 		renderer.drawBox(rect(), NAS2D::Color::Black);
-		renderer.drawText(mFont, mFocusedControl->second, position() + PaddingSize);
+
+		auto linePosition = position() + PaddingSize;
+
+		forEachLine(mFocusedControl->second, [this, &renderer, &linePosition](const std::string& lineStr) {
+			renderer.drawText(mFont, lineStr, linePosition);
+			linePosition.y += mFont.height();
+		});
```
DanRStevens, line 138: It's also possible to do a named capture, as an alternative to capturing `this`, and everything that has access to:
```cpp
[&font=mFont, &renderer, &linePosition]
```

DanRStevens (approved):

PR #1474, opened 2024-12-13
Fix/police overlay on load from main menu

oscar139:
When loading a game from the main menu toggling the police overlay will not display a police overlay even if there is a working police station. As an additional minor detail the toggle state of the overlay buttons carries over between loads which will not reflect the current display of the overlays in the map.

### DIFF
--- a/OPHD/States/MapViewState.cpp
+++ b/OPHD/States/MapViewState.cpp
@@ -204,6 +204,7 @@ MapViewState::MapViewState(MainReportsUiState& mainReportsState, const Planet::A
 	mRobots{"ui/robots.png", 46, constants::MarginTight},
 	mConnections{"ui/structures.png", 46, constants::MarginTight},
 	mPopulationPanel{mPopulation, mPopulationPool, mMorale},
+	mPoliceOverlays(static_cast<std::vector<Tile*>::size_type>(mTileMap->maxDepth())),
 	mResourceInfoBar{mResourcesCount, mPopulation, mMorale, mFood},
 	mRobotDeploymentSummary{mRobotPool},
 	mMiniMap{std::make_unique<MiniMap>(*mMapView, *mTileMap, mRobotList, planetAttributes.mapImagePath)},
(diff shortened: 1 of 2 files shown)


OPHD/States/MapViewState.cpp
```diff
@@ -1412,18 +1412,24 @@ void MapViewState::updateCommRangeOverlay()
...
 
 void MapViewState::resetPoliceOverlays()
 {
-	const auto adjustedZ = mTileMap->maxDepth() + 1;
-	mPoliceOverlays = std::vector<std::vector<Tile*>>(static_cast<std::size_t>(adjustedZ));
+	if (mTileMap != nullptr)
+	{
+		const auto adjustedZ = mTileMap->maxDepth() + 1;
+		mPoliceOverlays = std::vector<std::vector<Tile*>>(static_cast<std::size_t>(adjustedZ));
+	}
```
DanRStevens, line 1432: It might be preferable to use an early `return`, rather than indent all the mainline code.

----

Perhaps a bigger question is how or why is `mTileMap` not initialized before this is called. It would seem there may be a structural defect with the code, which is being compensated for here, rather than fixed at the source. This check seems to imply bad design on our part somewhere. Perhaps we should identify and fix the design that makes that check necessary.

DanRStevens (approved):

OPHD/States/MapViewState.cpp
```diff
@@ -204,6 +204,7 @@ MapViewState::MapViewState(MainReportsUiState& mainReportsState, const Planet::A
 	mRobots{"ui/robots.png", 46, constants::MarginTight},
 	mConnections{"ui/structures.png", 46, constants::MarginTight},
 	mPopulationPanel{mPopulation, mPopulationPool, mMorale},
+	mPoliceOverlays(static_cast<std::vector<Tile*>::size_type>(mTileMap->maxDepth())),
```
DanRStevens, line 207: Marking this as the off by one bug that was found and fixed in #1481.

PR #1481, opened 2024-12-19
Removed MapViewState::resetPoliceOverlays()

oscar139:
MapViewState::resetPoliceOverlays() is only used in MapViewState::updatePoliceOverlay.

### DIFF
--- a/OPHD/States/MapViewStateIO.cpp
+++ b/OPHD/States/MapViewStateIO.cpp
@@ -333,7 +333,7 @@ void MapViewState::load(const std::string& filePath)
 	}
 
 	updateCommRangeOverlay();
-	mPoliceOverlays.resize(static_cast<std::vector<Tile*>::size_type>(mTileMap->maxDepth()));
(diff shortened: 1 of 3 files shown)


OPHD/States/MapViewStateIO.cpp
```diff
@@ -333,6 +333,7 @@ void MapViewState::load(const std::string& filePath)
 	}
 
 	updateCommRangeOverlay();
+	mPoliceOverlays.resize(static_cast<std::vector<Tile*>::size_type>(mTileMap->maxDepth()));
```
DanRStevens, line 336: Should this be `maxDepth` or `maxDepth + 1`? We use `maxDepth + 1` elsewhere.

The max value is endpoint inclusive, whereas a size value is endpoint exclusive. Surface level gets a police overlay, along with all underground layers.

----

I like that you went out of your way to ensure a correct type for the `static_cast`. A lot of people may have just written `std::size_t`. Like the original code did.

DanRStevens (approved):

Issue #1491, opened 2025-01-08
Fix warning `-Wuninitialized` when `MapViewState` initializes `mCrimeExecution`


DanRStevens:
When updating to a newer version of GCC (packaged with Ubuntu 24.04) the following warning is generated:
> OPHD/States/MapViewState.cpp: In constructor 'MapViewState::MapViewState(MainReportsUiState&, const std::string&)':
> OPHD/States/MapViewState.cpp:179:25: error: member 'MapViewState::mNotificationArea' is used uninitialized [-Werror=uninitialized]
>   179 |         mCrimeExecution(mNotificationArea),
>       |                         ^~~~~~~~~~~~~~~~~

In this case, all that is needed is a reference to `mNotificationArea`, which is stored for later use, and is only used after `mNotificationArea` is fully initialized. In that sense, this code isn't actually a problem. Though it is perhaps a little bit odd to be passing around a reference to an object that hasn't been constructed yet.

----

Potentially we could re-order fields in `MapViewState` so that `mCrimeExecution` is after `mNotificationArea`. That would solve the warning.

One oddity noticed here is that game engine objects are usually initialized before UI objects. In this case, `CrimeExecution` handles game engine logic, though it also needs access to the UI to notify of events. This dependency is reversed from what is normally expected. Potentially we could decouple the two by having `CrimeExecution` raise events, where the event handler can have knowledge of the UI and be responsible for adding notifications. The would remove the need for `CrimeExecution` to have any direct knowledge of `mNotificationArea`. This seems like a cleaner long term solution.

----

Related:
- Issue https://github.com/lairworks/nas2d-core/issues/1155

DanRStevens:
Adding some debug info to reproduce the warning locally. Assuming a Docker image is available for the updated GCC compiler, then it can be used interactively to compile the file with the warning:

```sh
docker run --volume ./:/code --workdir=/code --rm --tty --interactive --user="1000:1000" ghcr.io/lairworks/build-env-nas2d-gcc:1.6 bash
```

From within the docker container, the one problem file can be compiled using:
```sh
make .build/Debug_Linux_OPHD/Intermediate/States/MapViewState.o
```

Or, equivalently:
```sh
[...]

PR #1498, opened 2025-01-10
Refactor crime `Difficulty` setting

DanRStevens:
Set `Difficulty` in constructors of crime related classes. Store a `const&` to the original value, so we don't need to use setters to keep shadow copies up-to-date.

----

Work for:
- #1491

Related to:
- PR #1497

### DIFF
--- a/OPHD/States/CrimeExecution.cpp
+++ b/OPHD/States/CrimeExecution.cpp
@@ -47,7 +47,7 @@ namespace
 }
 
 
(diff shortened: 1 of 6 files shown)


oscar139 (approved):

DanRStevens:
Curiously, the GCC warning mentioned in Issue #1491 seems to have gone away after commit 823aa6d7d8cab27a31aa88a66557440502cf6f9e, despite that commit not attempting to address the warning. Although there were changes to the constructor init-list, it was adding new parameters, not removing the `mNotificationArea` parameter that was responsible for the warning.

Issue #1446, opened 2024-03-12
Improve `const` correctness


DanRStevens:
There are a number of places in the code that can and should be using `const`, but currently aren't. I thought I'd create an issue to track such updates in one place.

One notable example is the use of `getStructures`, and iterating over the corresponding `Structure*` elements.

----

Related to the above, it should be noted that when `auto` deduces a pointer type, applying `const` to it will apply to the pointer, not the underlying type. This is similar to type aliases with `using` and `typedef`. When a type alias refers to a pointer, applying `const` to the alias applies it to the pointer, not the underlying pointed to type.

Example of `auto` deducing `Structure*`:
```cpp
    for (auto structure : storage)
```

Applying `const` to the above will deduce `Structure* const`, which may not be what was intended:
```cpp
    for (const auto structure : storage)
```

Most likely the intended type is `const Structure*`, which can be deduced using `auto*` instead of `auto`.

Example with deduced type `const Structure*`:
```cpp
    for (const auto* structure : storage)
```

DanRStevens:
The above mentioned area have been looked at and fixed. Seeing as how there are no obvious outstanding cases to fix, I think it's time to close this.

With that said, `const` correctness is a continual effort, and more cases may pop up in the future. In that case, we can still reference this issue to tie them together, even if the issue has already been closed.

PR #1505, opened 2025-01-12
Spelling and const

Brett208:
Just dropping in and saying hello. I noted a couple of mistakes when poking around the source code to try out the recent changes.

I did manage to crash the game when loading a save (I think a nullptr). If I reproduce, I'll put up an issue.

Unfortunately, I died by running out of common metals before I got very far. Guess I need to build some more trucks earlier next time. Also, the UI for the above ground factory is a bit tough to navigate. It kept idling itself without explaining itself (maybe because I didn't have enough resources).

~Brett


Brett208:
Force push to squash 3 similar commits

DanRStevens (approved):
Good to hear from you. This is unexpected. More so because GitHub didn't send a notification for your PR. Most puzzling.

That is a shocking number of typos. I'm surprised you found so many.

The large number of comment typos really got me thinking I never really read comments. It maybe drives home the point about how comments can get out-of-date.

Adding a reference to:
- Issue #1446

Nicely presented

PR #1543, opened 2025-01-25
Refactor/game state load

oscar139:


DanRStevens (approved):
Looks good.

I am curious why `mNewMapView` was added as a new field, and why replacing `mMapView` was delayed until `update` was called. Was there was some problem that was being avoided?

PR #1577, opened 2025-02-12
Refactor/FileIoDialog show

oscar139:


DanRStevens (commented):
This turned out better than I had planned, with one small concern.

appOPHD/UI/FileIo.h
```diff
@@ -25,7 +27,8 @@ class FileIo : public Window
 	FileIo();
 	~FileIo() override;
 
-	void setMode(FileOperation fileOp);
+	void showOpen(const std::string& directory = constants::SaveGamePath);
+	void showSave(const std::string& directory = constants::SaveGamePath);
```
DanRStevens, line 31: Kind of a similar concern about the putting `constants::SaveGamePath` as default value in `FileIo.h` as the previous comment.

The current use of this dialog is for saved games. I'd be hesitant to say it's specific to that though.

appOPHD/UI/FileIo.cpp
```diff
@@ -103,6 +103,16 @@ void FileIo::setMode(FileOperation fileOp)
 }
 
 
+void FileIo::showOpen(const std::string& directory)
+{
+	this->scanDirectory(directory);
+	mMode = FileOperation::Load;
+	title(constants::WindowFileIoTitleLoad);
+	mFileOperation.text(constants::WindowFileIoLoad);
+	this->show();
```
DanRStevens, line 104: Huh, this didn't stand out to me before, though the two `this->` parts should be unnecessary. Sometimes they're needed in a template class, though this isn't a template class.

appOPHD/UI/FileIo.cpp
```diff
@@ -113,6 +113,16 @@ void FileIo::showOpen(const std::string& directory)
 }
 
 
+void FileIo::showSave(const std::string& directory)
+{
+	this->scanDirectory(directory);
+	mMode = FileOperation::Save;
+	title(constants::WindowFileIoTitleSave);
+	mFileOperation.text(constants::WindowFileIoSave);
+	this->show();
```
DanRStevens, line 114: Same comment about using `this->`.

DanRStevens (approved):
Looks good.

One additional possibility, it may be possible to mark `scanDirectory` as `protected` or `private` now.

PR #1585, opened 2025-02-15
Refactor/file io signal split

oscar139:


appOPHD/UI/FileIo.cpp
```diff
@@ -180,12 +180,21 @@ void FileIo::onClose()
...
-	mSignal(mFileName.text(), mMode);
+	if(mMode == FileOperation::Save)
+	{
+		mSaveSignal(mFileName.text());
+	}
+
+	if(mMode == FileOperation::Load)
+	{
+		mSignal(mFileName.text(), mMode);
+	}
```
DanRStevens: Minor point: This technically breaks some old uses of `mSignal`, since existing client code will still expect save events to be delivered to the old signal. It's sometimes nice to provide a transition point where the new feature becomes functional, without breaking the old feature. That can be done by delivering the event to both `Signal` objects. Client code would only ever really be designed to listen to one of them. By keeping both functional, there's an easy transition period where client code can be updated at their own pace, without being forced to update. Though considering the context of the whole PR, I expect the old signal to be removed by the end, so we're talking about a really short transition period here. It's almost not worth mentioning.

appOPHD/UI/FileIo.h
```diff
@@ -20,15 +20,17 @@ class FileIo : public Window
 		Save
 	};
 
-	using FileOperationSignal = NAS2D::Signal<const std::string&, FileOperation>;
+	using FileSaveSignal = NAS2D::Signal<const std::string&>;
+	using FileOpenSignal = NAS2D::Signal<const std::string&>;
```
DanRStevens, line 24: I'm wondering if we should adjust the terminology to be "Save" and "Load". Those would be obvious opposites. With "Open", and obvious opposite is "Close", but that doesn't make sense here. Using "Open" also feels a bit out of place since you need to "open" a file to "save" to it. Though I guess it is kind of standard terminology, though perhaps more in the context of opening a file for editing.

In other places, the terms "save" and "load" were used, so there's a bit of an inconsistency.

PR #1646, opened 2025-02-26
Bug fix/text input disabled

oscar139:
In rearranging the calls to initialize `~MapViewState` no longer leaves the status of `SDL_TextInput` off. This had the side effect of causing the `NAS2D::Utility<StructureManager>::get().dropAllStructures();` in `MapViewState::load` where  `StructureManager` was still populated with pointers from objects from the now deleted `MapViewState`.  

Edit:
New strategy. I do not see any advantage of disabling `SDL_TextInput` by `~MapViewState`.

### DIFF
--- a/appOPHD/States/MapViewState.cpp
+++ b/appOPHD/States/MapViewState.cpp
@@ -240,8 +240,6 @@ MapViewState::~MapViewState()
 	eventHandler.mouseWheel().disconnect({this, &MapViewState::onMouseWheel});
 	eventHandler.windowResized().disconnect({this, &MapViewState::onWindowResized});
 
-	eventHandler.textInputMode(false);
-
 	NAS2D::Utility<std::map<class MineFacility*, Route>>::get().clear();
 }
 


appOPHD/States/GameState.cpp
```diff
@@ -165,7 +165,6 @@ void GameState::onLoadGame(const std::string& saveGameName)
 			throw std::runtime_error("Save game file does not exist: " + saveGamePath);
 		}
 		auto newMapViewState = std::make_unique<MapViewState>(*mMainReportsState.get(), saveGamePath);
-		newMapViewState->initialize();
 		mNewMapViewState = std::move(newMapViewState);
```
DanRStevens, line 168: This moves saved game loading out of the `try` block. The actual loading is deferred until `initialize` is called, so we kind of want that done early here. The `initializeMapViewState` isn't called until the `update` method is run, which is well outside of the `try` block.

Actually the `MapViewState::initialize` call was originally in `GameState::initializeMapViewState`, but it was moved out to support early loading of saved games.

DanRStevens (commented):
The second commit looks like the actual fix. The first commit I think may have some unintended consequences.

DanRStevens (approved):
I can accept this. Maybe slightly odd that we enable text input mode in two places, which can be called multiple times, and never disable it anywhere. But then given how we're using it, there's not really an obvious better solution. Particularly since text input mode is an on/off switch, without any kind of reference counting, and our calls to turn it on (and potentially off) are by classes that may have multiple instances, and those instances won't know about each other.

I'm wondering if it would make sense to enable text input mode globally somewhere early during app loading, and never touching it again.

Issue #894, opened 2021-05-23
Use event bubbling for UI event dispatch


DanRStevens:
We should rework event dispatch so it propagates through a tree of controls. Effectively we should use an event bubbling system, much like is used in Windows. Messages are delivered to a single top-level control, such as the main Window. The Window would be a container control. If the control is a container, then it should propagate event messages to contained controls.

Two main sources of events are keyboard and mouse messages. For keyboard messages, the concept of control "focus" is relevant (related: #884). Keyboard messages to a container are routed to the contained control that has the focus. Mouse messages are instead routed based on screen position of the mouse message, and checking if it overlaps with any of the contained controls. It's possible multiple controls will overlap. In that case, we may wish to route to the top most control. Alternatively, we could potentially look at routing to all controls that overlap with the mouse position.

Some messages, such as resize events, might be propagated to all contained controls.

If a container is disabled or invisible, it can block propagation of all messages to contained controls.

I believe the event bubbling method would help ensure a correct by default behavior in many cases. In particular, developers would no longer need to proactively defend against controls responding to messages when they are not visible or enabled. There seem to have been a number of UI bugs related to controls responding to events when they are not visible, or are obscured by other controls drawn over top. For instance, the following bugs seem to have had something to do with message routing:
- #779
- #702
- #863
- #403

ldicker83:
This would resolve a lot of complexity issues in event handling of controls (and would make modal dialog boxes way easier to implement, e.g., have an invisible control that covers the entire screen with the modal dialog in the middle, or just set the dialog to consume all events).

Anyway, I tried a system like this in the past that failed horribly which has discouraged me from attempting it again. Instead I went with the sig/slot approach.

I would think this is a better way to handle it but I don't have a lot of good insight into it. I do support it, especially now that I've had some time to think about it and see the problems with the current implementation.

I can at least say that the current implementation suffers from the same problem that Qt and wxWidgets libraries both also suffer from so I at least feel somewhat vindicated knowing that it's not just poor design but simply prone to errors if not carefully handled. :D

DanRStevens:
Currently the NAS2D `EventHandler` class is responsible for receiving and delivering user input events. This is largely handled in `EventHandler::pump()`. This is done through processing of [`SDL_Event`](https://wiki.libsdl.org/SDL2/SDL_Event) messages.

The `EventHandler::pump()` method is called from `StateManager::update()`, also in NAS2D.

[...]

PR #1650, opened 2025-02-27
Removed all calls to EventHandler::textInputMode

oscar139:


DanRStevens (approved):

ldicker83:
Why did we do this? Just curious because I don't see an explanation of the decision making behind it.

DanRStevens:
Looking back without the context fresh in my mind, this does appear a bit cryptic. We could probably have done a better job explaining this one.

In PR #1646 there is a bit of explanation, and in Issue #1651 there were links to relevant SDL2 documentation. This PR is cleanup after the fix in PR #1646.

The problem was text input was being inadvertently disabled. If you build the old commit, and load a saved game a couple of times, the second time it won't be possible to type any text input into the filename text field.

To test, checkout the problem code from before merging the fix in PR #1646:
```
git checkout 031ecf83a6135b1f7c459ce1dea29d70a0efc4f4^
```

One of the main reasons for the problem is the SDL functions were global in nature. Meanwhile the calls to them were local to a specific text field. We weren't really uses the functions in a way they were intended to be used. You might enable text input in one part of the code, and then disable it in an unrelated part of the code. This is basically how text input got disable. The destructor for `MapViewState` turned off text input.

There was no clear reason to disable text input. The default is for it to be on. By removing all calls to these functions, things just work, and are kept simple.

ldicker83:
Understood.

We may want to limit it to very specific use cases where text editing input would make sense (e.g., a text input field). Tends to work better than capturing and translating each character as it comes in and applying modifiers as needed.

Anyway, sounds good!

DanRStevens:
It might make more sense to use those functions in the future once we have a better defined concept of keyboard focus. That would happen after updates to event routing (Issue #894). Since the SDL function can pop up an input window when called, it makes sense to tie it to a control having active keyboard focus. When a text input field gets focus, it can call to enable text input, and when such a control loses focus it can disable text input.

Making calls when gaining and losing focus would work much more naturally than during object creation and destruction, which the previous code did. Sometimes objects are created in the background and not made visible until some other event occurs. In that case, we really don't want to be popping up an input window until the control is actually visible and gains keyboard focus. Worse, using constructors and destructors means creating two objects and then destroying one would disable text input while one of the text input controls still exists, but can no longer receive input because it's disabled. Object construction/destruction is a rather different concept from gaining and losing keyboard focus.

Right now, the global event listening design, with multiple event subscribers, means we don't have a well defined concept of a single control with active keyboard focus.

PR #1658, opened 2025-03-05
Refactor/swap file io signal for delegate

oscar139:


appOPHD/States/MainMenuState.cpp
```diff
@@ -26,7 +26,8 @@ MainMenuState::MainMenuState() :
 	}},
 	lblVersion{constants::Version},
 	mReturnState{this},
-	mFade{{this, &MainMenuState::onFadeComplete}}
+	mFade{{this, &MainMenuState::onFadeComplete}},
+	mFileIoDialog{{this, &MainMenuState::onLoadGame}}
```
DanRStevens, line 30: When connecting using the `Delegate`, we should remove the `connect` call using the `Signal`.

Edit: Which is in the following commit.

appOPHD/States/GameState.cpp
```diff
@@ -20,15 +20,17 @@ NAS2D::Point<int> MOUSE_COORDS; /**< Mouse Coordinates. Used by other states/wra
...
 {
 	initializeGameState();
 }
 
 
 GameState::GameState(const Planet::Attributes& planetAttributes, Difficulty selectedDifficulty) :
 	mMainReportsState{std::make_unique<MainReportsUiState>()},
-	mMapViewState{std::make_unique<MapViewState>(*this, planetAttributes, selectedDifficulty)}
+	mMapViewState{std::make_unique<MapViewState>(*this, planetAttributes, selectedDifficulty)},
+	mFileIoDialog{{this, &GameState::onLoadGame}}
```
DanRStevens: Same for this class. When connecting with `Delegate` we should remove the `connect` call for `Signal`.

Edit: Which is in the following commit.

appOPHD/UI/FileIo.cpp
```diff
@@ -19,8 +19,14 @@ using namespace NAS2D;
...
 
 FileIo::FileIo(FileLoadDelegate fileLoadDelegate) :
+	FileIo{fileLoadDelegate, FileSaveDelegate{}}
+{}
+
+
+FileIo::FileIo(FileLoadDelegate fileLoadDelegate, FileSaveDelegate fileSaveDelegate) :
 	Window{"File I/O"},
 	mFileLoadDelegate{fileLoadDelegate},
+	mFileSaveDelegate{fileSaveDelegate},
```
DanRStevens: Probably makes sense to add a call to the new delegate in this commit, so the new feature is usable. It would need to be a checked call, since a `Delegate` is default initialized to `nullptr` values. Additionally, we don't always allow saving, such as when `FileIo` is opened from the main menu. Though it would never be activated with a "Save" action in that case.

appOPHD/States/MapViewState.h
```diff
@@ -118,6 +118,8 @@ class MapViewState : public Wrapper
 	QuitSignal::Source& quit() { return mQuitSignal; }
 	MapChangedSignal::Source& mapChanged() { return mMapChangedSignal; }
 
+	FileIo::FileSaveDelegate fileSaveDelegate() { return {this, &MapViewState::onSaveGame}; }
```
DanRStevens, line 121: This feels a bit weird. Maybe `onSaveGame` should be `public`. Effectively this is making the `onSaveGame` method `public`, so might as well make the original method `public` rather than wrapping access in a new `public` method.

With `public` access to `MapViewState::onSaveGame`, the `Delegate` could be constructed in the caller. Or alternatively, the caller could have it's own handler member, which calls `MapViewState::onSaveGame`. Or perhaps the caller's handler could make a direct call to `save`. Maybe `save` should be made `public`, and perhaps we don't really need `MapViewState::onSaveGame`, since that class will no longer be directly responsible for this event, since it's no longer in control of the `FileIo` dialog.

appOPHD/UI/FileIo.cpp
```diff
@@ -197,7 +197,7 @@ void FileIo::onFileIo()
 {
 	if(mMode == FileOperation::Save)
 	{
-		mSaveSignal(mFileName.text());
+		mFileSaveDelegate(mFileName.text());
```
DanRStevens, line 200: This should probably be checked before calling. The save handler is optional, so a call when it's unset will cause a seg fault.

DanRStevens (commented):
Noting one small safety issue. It won't be an issue in practice with the current code. Though if someone were to write new uses for `FileIo`, and did a bad job of it, there would be a possibility of a seg fault.

DanRStevens (approved):

Issue #1573, opened 2025-02-11
Improve build times


DanRStevens:
Ideas taken from:
- Issue #1418
- Issue #975
  - https://github.com/OutpostUniverse/OPHD/issues/975#issuecomment-949078696

----

Commands from the above linked issues were used with GCC, Clang, and MSBuild to get build time reports.

Related idea, since the project is being split into sub projects, it may be possible to cache build outputs per sub project, which may help improve incremental build performance.

----

GCC: 
```sh
make CXXFLAGS_EXTRA="-ftime-report"
```

GCC showed the vast majority of the time was spent parsing, with template instantiation a distant 2nd.

----

```sh
make CXX="clang++" CXXFLAGS_EXTRA="-ftime-report"
```

Clang results are a bit harder to read and draw conclusions from. Seems most of the time was spent on various passes and analysis, and also a good chunk on code generation.

It's also possible to include the `-ftime-trace` flag to get `.json` output alongside the object files:
```sh
make CXX="clang++" CXXFLAGS_EXTRA="-ftime-report -ftime-trace"
```

The output can be viewed using Chrome: `chrome://tracing/`
Drag and drop a trace file into Chrome, such as: `.build/Debug_Linux_appOPHD/Intermediate/main.json`

These flame charts are very useful for determining which header files are slow to `#include` and parse. Sometimes the results are surprising.

![Image](https://github.com/user-attachments/assets/bdaba077-64de-44a8-aeb3-ade3c24275c7)

----

There was an outdated note about getting build timings from MSVC using:
```cmd
msbuild /fl /flp:Verbosity=diagnostic Your.sln
```

Documentation:
https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2022

Of interest is the documentation for the `-consoleLoggerParameters` flag, which is related to the `-fileLoggerParameters` (`-flp`) flag.

It seems the `Verbosity=diagnostic` flag no longer displays timing information, though it does give a lot of other detailed information. Some timing information can be found using `PerformanceSummary`:
```
msbuild . /maxCpuCount /warnAsError /property:RunCodeAnalysis=true /consoleLoggerParameters:PerformanceSummary
```
>       234 ms  RunMergeNativeCodeAnalysis                 7 calls
>       235 ms  GetCopyToOutputDirectoryItems              7 calls
>       313 ms  ClCompile                                  7 calls
[...]

PR #1666, opened 2025-03-08
Use `std::to_string` instead of `NAS2D::stringFrom`

DanRStevens:
These are simple cases of converting an `int` to a `std::string` and the `<string>` header has already been implicitly included. There is no need for an additional include or for the additional generality of `stringFrom`.

Related:
- Issue #1573


ldicker83:
Agreed. Generally speaking a lot of the string utility functions we have in NAS2D can probably be eliminated with the C++17/20 features in std.

PR #1668, opened 2025-03-08
Refactor/extract colony ship to class

oscar139:


appOPHD/States/MapViewStateTurn.cpp
```diff
@@ -414,37 +416,6 @@ void MapViewState::updateResources()
...
 
 
-/**
- * Check for colony ship deorbiting; if any colonists are remaining, kill
- * them and reduce morale by an appropriate amount.
- */
-void MapViewState::checkColonyShip()
-{
-	if (mTurnCount == constants::ColonyShipOrbitTime)
-	{
```
DanRStevens, line 424: Or perhaps rather than an event, we could have continued to poll the colony ship status here, and handle the UI related portions of the crash event.

The updates to core game mechanics are of course better handled by `ColonyShip`.

DanRStevens (approved):

appOPHD/States/ColonyShip.cpp
```diff
@@ -0,0 +1,24 @@
...
+{
+	if (mColonyShipData.turnsOfManeuveringFuel > 0) { --mColonyShipData.turnsOfManeuveringFuel; }
+
+	if (mColonyShipData.turnsOfManeuveringFuel == 0)
+	{
+		mCrashData = mColonyShipData;
+		mColonyShipData.cargoLanders = 0;
+		mColonyShipData.colonistLanders = 0;
+	}
+}
```
DanRStevens, line 24: Just noting this section seems to handle this event differently than we normally do in other parts of the code. Other sections would tend to use `Signal` or `Delegate` to report an event. Here the `std::optional` field `mCrashData` is being set, which is later polled.

A potential alternative might be to return crash data from this function.

No change needed here. Just pointing out the style is different. We should probably try this out so we have some experience with this style.

DanRStevens (approved):
Update looks good.

Issue #1648, opened 2025-02-27
Single stage init for `MapViewState` game loading


DanRStevens:
Currently `MapViewState` has a constructor overload taking a saved game name, which it stores locally, and then completes the load later during `initialize`. We should instead use single stage initialization, where the full load is done by the time the constructor completes, and if there was any error loading, it should `throw` from the constructor, preventing object instantiation. This prevents creating zombie objects which exist, and can have methods called, but don't really work because they're in some kind of failed state.

Related:
- Issue #1647
- Issue #1336
- Issue #1337

oscar139:
Additionally this is also related to:

- Issue #650

As `MapViewState` is reduced in responsibilities I suspect the changes for single stage init will be easier.

PR #1675, opened 2025-03-19
Refactor load save game

oscar139:
Moved colony ship related loading code to colony ship files. Game components of `GameState` now load and save to and from a common `XmlDocument` instead of making repeated file IO operations.

### DIFF
--- a/appOPHD/States/ColonyShip.cpp
+++ b/appOPHD/States/ColonyShip.cpp
@@ -2,6 +2,47 @@
 
 #include "appOPHD/Constants/Strings.h"
 
+#include <NAS2D/ParserHelper.h>
+
+namespace
+{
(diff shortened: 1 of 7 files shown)


DanRStevens (approved):
I have some reservations about adding `XmlDocument` as a member field. We can probably make some easy improvements regarding forward declarations. With that said, I don't think there's really anything blocking here, and this seems to move the code forward. I'll leave it to you if you want to address any feedback before or after merging.

oscar139:
https://github.com/OutpostUniverse/OPHD/pull/1675#discussion_r2003773516 this relates to Issue #1648. A temporary `XmlSaveDocument` can be used after MapViewState has a single stage init.

PR #2198, opened 2026-06-01
Prepare tag v0810

ldicker83:
Preparation for v0.8.10 release.

Changes version strings and changelog.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,62 @@
 # Change Log
 This is the changelog for OutpostHD.
 
+## [0.8.10] - 2026-06-01
+
+Interim release between major feature updates. A lot of improvements to the code that address some of the instability in previous releases.
+
+### Added
+- Technology catalog
+- Research Report panel in Reports Panel (WIP)
+- Added new announcement types for colony ship crashes with cargo and colonists (Alexander Hunt)
+- Add `<Ctrl> + <Shift> + <F12>` key combo for debug border drawing
+
+### Changed
+- Up/Down keybindings now move the camera in a more intuitive way for the user
+- Tubes can now be placed like roads -- they will automatically choose their direct based on what they're connected to
+- Moved structure definitions from hard-coded values in source to external definition file
+- Structures that are in a destroyed state can no longer be repaired
+- Mine Facility now transmits connection to Command Center
+- Larger font used throughout all views for improved visibility
+- Added difficulty-dependent "grace period" for crime and morale (Andrew Gorcester)
+- Added difficulty-dependent morale hit modifier for colony ship deorbit with colonists still aboard (Alexander Hunt)
+- Improved morale handling throughout code base (Alexander Hunt)
+- Improved morale change reasons reporting (Alexander Hunt)
+- Improved memory handling/management (should help with some odd crash behavior)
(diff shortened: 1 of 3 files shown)


ldicker83:
Compile failures look like a possible caching issue? @DanRStevens could you investigate? I don't know where to begin looking -_-

CHANGELOG.md
```diff
@@ -1,6 +1,62 @@
...
+- Adjustments to crime handling based on difficulty
+- Various UI visual improvements including larger fonts, control placement and dynamic sizing
+- ProductCatalog will throw an exception instead of silently failing
+- Less punishing ore haul falloff
+- Fix offset error in drawLine
+- Fix "Take Me There" button visibility on message change
+
+
+### Fixed
+- Robots that are set to self will no longer complete their tasks before terminating
```
DanRStevens: I think that should read "self destruct"

CHANGELOG.md
```diff
@@ -1,6 +1,62 @@
...
+- Added difficulty-dependent morale hit modifier for colony ship deorbit with colonists still aboard (Alexander Hunt)
+- Improved morale handling throughout code base (Alexander Hunt)
+- Improved morale change reasons reporting (Alexander Hunt)
+- Improved memory handling/management (should help with some odd crash behavior)
+- Adjustments to crime handling based on difficulty
+- Various UI visual improvements including larger fonts, control placement and dynamic sizing
+- ProductCatalog will throw an exception instead of silently failing
+- Less punishing ore haul falloff
+- Fix offset error in drawLine
+- Fix "Take Me There" button visibility on message change
```
DanRStevens: Wondering if these should be moved to the "Fixed" section.

appOPHD/Constants/Strings.h
```diff
@@ -15,7 +15,7 @@ namespace constants
 	// =====================================
 	// = MISCELLANEOUS
 	// =====================================
-	const std::string Version = "v0.8.7";
+	const std::string Version = "v0.8.10";
```
DanRStevens, line 18: Lots of patches 😉

DanRStevens (approved):
Confirmed on the caching issue. We don't include the version of the underlying Windows image as part of the cache key. Maybe we should, now that we know updates can be a problem.

Anyway, caches were cleared, and your jobs were re-run, and they succeed now.

----

Edit:
Adding link to the Caches page, so you'll know where to find it in the future:
https://github.com/OutpostUniverse/OPHD/actions/caches

----

Edit 2:
As for why the underlying Windows image matters, it probably related to the `WholeProgramOptimization` setting used when compiling NAS2D:
https://learn.microsoft.com/en-us/cpp/build/reference/gl-whole-program-optimization?view=msvc-170
> The format of files produced with /GL in the current version often isn't readable by later versions of Visual Studio and the MSVC toolset. Unless you're willing to ship copies of the .lib file for all versions of Visual Studio you expect your users to use, now and in the future, don't ship a .lib file made up of .obj files produced by /GL . For more information, see [Restrictions on binary compatibility](https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170#restrictions).

DanRStevens (approved):

PR #2244, opened 2026-07-15
Invert morale index to get correct modifiers based on morale

ldicker83:
Fixes population model index calculation. As it was, higher morale would provide worse modifiers.

### DIFF
--- a/libOPHD/Population/PopulationModel.cpp
+++ b/libOPHD/Population/PopulationModel.cpp
@@ -30,7 +30,7 @@ namespace
 	 */
 	std::size_t moraleIndex(int morale)
 	{
-		return static_cast<std::size_t>(std::clamp(morale, 1, 999) / 200);
+		return static_cast<std::size_t>(4 - (std::clamp(morale, 0, 999) / 200));
 	}
 }
 


libOPHD/Population/PopulationModel.cpp
```diff
@@ -30,7 +30,7 @@ namespace
 	 */
 	std::size_t moraleIndex(int morale)
 	{
-		return static_cast<std::size_t>(std::clamp(morale, 1, 999) / 200);
+		return static_cast<std::size_t>(4 - (std::clamp(morale, 0, 999) / 200));
```
DanRStevens, line 33: Should probably use `moraleModifierTable.size()` rather than the hardcoded constant `4`. Though yes, it was backwards.

It might also make sense to calculate the `200` based on the number of entries in the table.

DanRStevens (approved):
I'm a bit amused by this. That would definitely explain why trying to work on morale didn't seem to improve outcomes. Didn't realize it made things worse.

libOPHD/Population/PopulationModel.cpp
ldicker83: Not a bad idea about using size, didn't think about that. Will make that part of a refactor change set after I do some more work here with balancing/tuning the numbers a bit.

PR #2248, opened 2026-07-28
Adjust population model

ldicker83:
Adjusts the population model such that it is easier to develop population.

Wanted this commit to have more to it but I think a lot of the other changes I want to make are more refactoring type changes so going to commit this and come up with a different branch for refactoring.

### DIFF
--- a/libOPHD/Population/PopulationModel.cpp
+++ b/libOPHD/Population/PopulationModel.cpp
@@ -16,11 +16,11 @@ namespace
 	const int adultToRetireeBase = 2000;
 
 	const std::array moraleModifierTable{
-		MoraleModifier{50, 50, 110, 80},  // Excellent
-		MoraleModifier{25, 25, 90, 75},   // Good
-		MoraleModifier{0, 0, 60, 40},     // Fair
-		MoraleModifier{-25, -25, 40, 20}, // Poor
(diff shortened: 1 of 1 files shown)


libOPHD/Population/PopulationModel.cpp
```diff
@@ -11,9 +11,9 @@
 
 namespace
 {
-	constexpr auto studentToScientistRate = 35;
-	constexpr auto studentToAdultBase = 190;
-	constexpr auto adultToRetireeBase = 2000;
+	constexpr auto StudentToScientistRate = 35;
+	constexpr auto StudentToAdultBase = 190;
+	constexpr auto AdultToRetireeBase = 2000;
```
DanRStevens, line 16: I know it's common to use UpperCamelCase for global constants, though elsewhere the code is using lowerCamelCase.

I think it makes sense to use lowerCamelCase consistently for all constants, not just local constants. Plus it saves having to rename variables when they are marked const. It also avoids ambiguity with types which typically use UpperCamelCase.

ldicker83: For local/global constants I want to switch to UpperCamelCase, that's going to be the general plan going forward.

PR #2249, opened 2026-07-30
Refactor population model

ldicker83:
Some basic/minor refactoring.

Changes `PopulationModel::update()` to use a parameter object instead of a long list of parameters. This will aid in future updates where I intend to inject transient modifiers into the update loop that are tracked outside of the population model (e.g., first landing fertility bonus and mortality resistance).

### DIFF
--- a/libOPHD/Population/Morale.h
+++ b/libOPHD/Population/Morale.h
@@ -12,7 +12,7 @@ struct MoraleModifier
 {
 	int researchBonus{0};
 	int productionBonus{0};
-	int fertilityCost{0};
+	int fertilityResistance{0};
 	int mortalityResistance{0};
 };
 
--- a/libOPHD/Population/Morale.h
+++ b/libOPHD/Population/Morale.h
@@ -12,8 +12,8 @@ struct MoraleModifier
 {
 	int researchBonus{0};
 	int productionBonus{0};
-	int fertilityRate{0};
-	int mortalityRate{0};
+	int fertilityCost{0};
+	int mortalityResistance{0};
 };
 
 
--- a/libOPHD/Population/PopulationModel.cpp
+++ b/libOPHD/Population/PopulationModel.cpp
@@ -95,7 +95,7 @@ void PopulationModel::spawnPopulation(int morale, int residences, int nurseries,
 
 	const int totalAdults = mPopulation.worker + mPopulation.scientist;
 
-	const int divisorChild = MoraleModifierTable[moraleIndex(morale)].fertilityCost;
+	const int divisorChild = MoraleModifierTable[moraleIndex(morale)].fertilityResistance;
 	const int divisorStudent = ((std::max(mPopulation.adults(), StudentToAdultBase) / 40) * 3 + 13) * 4;
 	const int divisorAdult = ((std::max(mPopulation.adults(), StudentToAdultBase) / 40) * 3 + 38) * 4;
 	const int divisorRetiree = (mPopulationGrowth.retiree < MinRetireeGrowthThreshold) ?
--- a/libOPHD/Population/PopulationModel.cpp
+++ b/libOPHD/Population/PopulationModel.cpp
@@ -17,12 +17,12 @@ namespace
 	constexpr auto AdultToRetireeBase = 2000;
 	constexpr auto MinRetireeGrowthThreshold = 1200; // Prevents retirees for approximately 150 turns
 
-	const std::array moraleModifierTable{
-		MoraleModifier{50, 50, 30, 110},  // Excellent
-		MoraleModifier{25, 25, 40, 90},   // Good
-		MoraleModifier{0, 0, 50, 70},     // Fair
-		MoraleModifier{-25, -25, 70, 50}, // Poor
-		MoraleModifier{-50, -50, 90, 30}  // Terrible
+	const std::array moraleModifierTable {
+		MoraleModifier{0, 0, 30, 110},	// Excellent
+		MoraleModifier{0, 0, 40, 90},	// Good
+		MoraleModifier{0, 0, 50, 70},	// Fair
+		MoraleModifier{0, 0, 70, 50},	// Poor
+		MoraleModifier{0, 0, 90, 30}	// Terrible
 	};
 
 
--- a/libOPHD/Population/PopulationModel.cpp
+++ b/libOPHD/Population/PopulationModel.cpp
@@ -47,7 +47,7 @@ namespace
 
 	void retireAdults(int toRetire, PopulationTable& population)
 	{
-		for (toRetire; toRetire > 0;)
+		while (toRetire > 0)
 		{
 			/** Workers retire earlier than scientists. */
 			auto& retireRole = (randomNumber.generate(0, 100) <= 45) ? population.scientist : population.worker;
--- a/libOPHD/Population/PopulationModel.cpp
+++ b/libOPHD/Population/PopulationModel.cpp
@@ -66,7 +66,7 @@ void PopulationModel::spawnPopulation(int morale, int residences, int nurseries,
 
 	int totalAdults = mPopulation.worker + mPopulation.scientist;
 
-	int divisorChild = moraleModifierTable[moraleIndex(morale)].fertilityRate;
+	int divisorChild = moraleModifierTable[moraleIndex(morale)].fertilityCost;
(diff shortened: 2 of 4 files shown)


libOPHD/Population/PopulationModel.cpp
```diff
@@ -54,46 +87,31 @@ void PopulationModel::removePopulation(const PopulationTable& population)
...
-		/** Workers retire earlier than scientists. */
-		auto& retireRole = randomNumber.generate(0, 100) <= 45 ?
-			mPopulation.scientist : mPopulation.worker;
-		if (retireRole > 0)
-		{
-			--retireRole;
-			--toRetire;
-		}
-	}
+	assertNoOverritirement(newRoles.retiree, mPopulation.employable());
```
DanRStevens: Typo in "Overritirement"

libOPHD/Population/Morale.h
```diff
@@ -12,8 +12,8 @@ struct MoraleModifier
 {
 	int researchBonus{0};
 	int productionBonus{0};
-	int fertilityRate{0};
-	int mortalityRate{0};
+	int fertilityCost{0};
+	int mortalityResistance{0};
```
DanRStevens, line 16: Probably wise to rename these fields, since they are kind of the inverse of "rate". Though renaming each to something different does perhaps raise questions as to why they are now different. Maybe we could settle on a common term for the inverse of a rate. I kind of feel like "resistance" is reasonable. I can't really think of anything better. Checking for words with an inverse meaning to "rate" gives things like "inverse rate", "period", "interval", and "resistance".

libOPHD/Population/PopulationModel.cpp
```diff
@@ -43,6 +43,22 @@ namespace
...
+		for (toRetire; toRetire > 0;)
+		{
+			/** Workers retire earlier than scientists. */
+			auto& retireRole = (randomNumber.generate(0, 100) <= 45) ? population.scientist : population.worker;
+
+			if (retireRole > 0)
+			{
+				--retireRole;
+				--toRetire;
+			}
```
DanRStevens, line 59: Hmm, couldn't this potentially select a role with `0` people, and then produce a negative result?

Probably out of scope for the current change, since this would be pre-existing. Though might be something to look at later.

libOPHD/Population/PopulationModel.cpp
```diff
@@ -59,6 +59,12 @@ namespace
 			}
 		}
 	}
+
+
+	int childGrowth(int residences, int nurseries, PopulationTable& population)
+	{
+		return (residences > 0 || nurseries > 0) ? population.scientist / 4 + population.worker / 2 : 0;
```
DanRStevens, line 66: Out of scope for this change, though I've sometimes looked at this line and thought it made nurseries seem kind of pointless. There will pretty much always be residences. Maybe nurseries should provide some kind of additional bonus.

libOPHD/Population/PopulationModel.h
```diff
@@ -14,7 +25,7 @@ class PopulationModel
 	void addPopulation(const PopulationTable& population);
 	void removePopulation(const PopulationTable& population);
 
-	int update(int morale, int food, int residences, int universities, int nurseries, int hospitals);
+	int update(const UpdateParameters& params);
```
DanRStevens, line 28: Probably wise when there are this many parameters. Especially since they are all the same type, so there's no type checking to warn about a bad parameter order. Mind you, checking parameter order is basically just pushed off to initializing the `UpdateParameters` struct. Though potentially it could have fields initialized by name.

libOPHD/Population/PopulationModel.cpp
```diff
@@ -130,11 +148,11 @@ void PopulationModel::killRoles(const PopulationTable& divisor)
...
 {
-	const auto mortalityRate = moraleModifierTable[moraleIndex(morale)].mortalityRate;
+	const auto mortalityRate = MoraleModifierTable[moraleIndex(morale)].mortalityResistance;
 
-	int divisorChild = mortalityRate + (nurseries * 10);
-	int divisorStudent = mortalityRate + (hospitals * 65);
-	int divisorAdult = mortalityRate + 250 + (hospitals * 60);
+	const int divisorChild = mortalityRate + (nurseries * 10);
+	const int divisorStudent = mortalityRate + (hospitals * 65);
+	const int divisorAdult = mortalityRate + 250 + (hospitals * 60);
```
DanRStevens, line 155: Continuing the discussion from the last PR regarding lowerCamelCase vs UpperCamelCase for constants:
There is now a discrepancy here. The `MoraleModifierTable` was switched to UpperCamelCase, while the divisors remain lowerCamelCase. The only difference is scope.

There seems to be a trend in C++ to use `const` basically everywhere whenever possible. I agree with that trend. Though it can mean pretty much everything ends up being marked `const`. That's especially true when code is extracted to functions, and parameters end up being a `const` copy of some computed argument. It makes `const` very common, and nothing particularly special. It would seem odd to signify `const` using a naming convention. Especially since `const` gets added to old code so frequently, it would be a pain to go renaming things as the `const` modifier is added. Or renaming things when it's removed, because some code update requires a value to be variable.

Having a type modifier affect the name also feels a bit reminiscent of Hungarian notation. That's sort of fallen out of style. Given that refactoring sometimes changes the type of a variable, it becomes a burden to keep naming conventions up to date with types.

Plus, UpperCamelCase is already used for types. It seems a bit weird and arbitrary to overload it with both types and constant values. Constant values are much more like variables than types, so it seems more logical for constants and variables to share a naming convention rather than constants and types.

Mostly though, I find it reduces maintenance burdens when constants and variables share a common convention. It's fairly common for `const` to be added or removed, or for the scope of a variable to change. What is const or global now might very well be refactored to be not a const or global later.

DanRStevens (commented):
There's a typo I would fix before merging. Also might be worth giving some consideration to the comment on naming conventions. Otherwise this looks good.

libOPHD/Population/Morale.h
ldicker83: I'll go with resistance.

libOPHD/Population/PopulationModel.h
ldicker83: I'd considered using named initializers. It's a little messy looking in-code but it's at least an option now. :)

libOPHD/Population/PopulationModel.cpp
ldicker83: For the first part, the point _is_ scope. Module level / global symbols I've been trending toward UpperCamelCase with locally scoped symbols as lowerCamelCase. This makes it visually very easy to distinguish where something is scoped.

For the second part, I've noticed that with const too, though that's not really why I ended up doing it. I've found, at least with VC's compiler, that explicitly marking local variables as const hints to the optimizer that the variable can be optimized away. This may still be the case with recent versions even if they're not declared const, but in older versions I did see the difference. I don't know if there's an underlying set of rules governing that and I haven't tested it with other compilers. This is mostly from experience.

It also lets someone reading the code know that these values don't change anywhere in the local scope. Helps to relieve mental load. Granted these could be wrapped up into the areas where they're called but that makes things really sloppy.

DanRStevens: In your previous comment it seems to suggest you wanted to use UpperCamelCase for all constants. Or at least I didn't interpret what you wrote to be about scope.
https://github.com/OutpostUniverse/OPHD/pull/2248#discussion_r3677036972
> For local/global constants I want to switch to UpperCamelCase, that's going to be the general plan going forward.

Making a distinction based on scope maybe makes a little more sense, though I find the traditional use of UpperCamelCase for constants to be at odds with the use of UpperCamelCase for types.

What about namespace scoped constants? There's been a trend in the code to wrap globals into unnamed namespaces, which at least limits their access to the same translation unit. Should those constants be UpperCamelCase or lowerCamelCase? I think unnamed namespace scoped constants have been pretty consistently lowerCamelCase so far.

As for local/global distinction, if functions are kept reasonably short, which they should be, it should be readily apparent if something is locally scoped or not. I personally strive to keep functions within one screen of text. At one screen of text you can see the whole function without scrolling, which means less context you need to keep in your head. I think the OPHD and NAS2D source code do a pretty good job of meeting that standard. The few large functions I can think of that still require scrolling are reasonably simple functions with a lot of `case` labels, such as for structure types, or window events.

Short functions combined with IDE features to jump to a definition seem to largely alleviate the need to encode scope information into a name.

